PR #2, opened 2023-12-05
Mistral

awni:
-Mistral 7B 
-Update mlx-example README


mistral/mistral.py
```diff
@@ -0,0 +1,287 @@
...
+            queries = self.rope(queries)
+            keys = self.rope(keys)
+
+        scores = (queries * self.scale) @ keys.transpose(0, 1, 3, 2)
+        if mask is not None:
+            scores += mask
+        scores = mx.softmax(scores.astype(mx.float32), axis=-1).astype(scores.dtype)
+        output = (scores @ values).transpose(0, 2, 1, 3).reshape(B, L, -1)
+
+        #        queries = queries.reshape(B, self.n_kv_heads, self.repeats, L, -1)
```
angeloskath, line 96: Maybe one of the two or a comment about why this is here and commented?

mistral/mistral.py
```diff
@@ -0,0 +1,287 @@
...
+        self.n_layers = args.n_layers
+        assert self.vocab_size > 0
+
+        self.tok_embeddings = nn.Embedding(args.vocab_size, args.dim)
+
+        self.layers = [TransformerBlock(args=args) for _ in range(args.n_layers)]
+
+        self.norm = RMSNorm(args.dim, eps=args.norm_eps)
+
+        self.output = nn.Linear(args.dim, args.vocab_size, bias=False)
```
angeloskath, line 156: Nitpick: The empty lines feel a bit unnecessary

angeloskath (approved):
Awesome thanks for adding it!

mistral/mistral.py
awni: Lol sorry this was just copied from the mistral source.. not my style either.

mistral/mistral.py
```diff
@@ -0,0 +1,287 @@
...
+    def __init__(self, dims: int, eps: float = 1e-5):
+        super().__init__()
+        self.weight = mx.ones((dims,))
+        self.eps = eps
+
+    def _norm(self, x):
+        return x * mx.rsqrt(x.square().mean(-1, keepdims=True) + self.eps)
+
+    def __call__(self, x):
+        output = self._norm(x.astype(mx.float32)).astype(x.dtype)
```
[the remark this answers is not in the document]
awni: Good question.. I didn't dig into it. Let me see if it passes my tests with using our RMSNorm + float32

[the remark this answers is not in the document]
awni: The other difference is we prescale the input and `sum` rather than `mean` the input.  But I think the `float32` casting is more important as I get more different results when I remove that and do it all in fp16.

angeloskath (approved):

mistral/mistral.py
```diff
@@ -0,0 +1,275 @@
...
+        self,
+        inputs: mx.array,
+        cache=None,
+    ):
+        h = self.tok_embeddings(inputs)
+
+        mask = None
+        if h.shape[1] > 1:
+            mask = nn.MultiHeadAttention.create_additive_causal_mask(h.shape[1])
+            mask = mask.astype(h.dtype)
```
thempatel, line 156: @awni Love the work here! qq: the mistral uses sliding window attention, however this looks like the standard attention mask. I was wondering if the omission/change was intentional?

awni: We left it out to keep the example simple. It's only needed I believe for when you get to > 4096 tokens which is pretty big. So technically this example has a maximum context size of 4096 tokens.

thempatel: ah yes you are right! looking at the paper again, the window size is 4096 -- thank you for the quick response.

PR #8, opened 2023-12-06
fix(sd): Add mlx as a dependency

### LINKED ISSUE #6: stable diffusion is not working when I  try to pip install
pip install -r requirements.txt is not working

dcvz:
Should likely resolve #6

### DIFF
--- a/stable_diffusion/requirements.txt
+++ b/stable_diffusion/requirements.txt
@@ -1,3 +1,4 @@
+mlx
 safetensors
 huggingface-hub
 regex


power9799:
> Should likely resolve #6

I already did import mlx, it doesn't work. Also how is it possible mlx is getting 8.79 image gens per second, what resolution is that 128*128 at 4 steps? because at 512*512 res at 30 steps, that is insanely fast, faster than rtx 4090

angeloskath:
Updated the readme to better reflect what those numbers are. Perhaps I should do another pass... It is images processed by the UNet. Not images generated. So `8.79 im/s` with a batch of 16 means ~90 seconds per 16 images generated when doing 50 steps. Or 45 seconds with 25 steps or 180 seconds with 100 steps etc.

PR #10, opened 2023-12-06
Fix unsupported ScalarType BFloat16

ricardo-larosa:
Fix/ workaround to error:
```
TypeError: Got unsupported ScalarType BFloat16
```

### DIFF
--- a/llama/convert.py
+++ b/llama/convert.py
@@ -12,8 +12,7 @@ def map_torch_to_mlx(key, value):
         key = "embedding.weight"
 
     elif "norm" in key:
-        key = key.replace("attention_norm", "norm1").replace(
-            "ffn_norm", "norm2")
+        key = key.replace("attention_norm", "norm1").replace("ffn_norm", "norm2")
 
     elif "wq" in key or "wk" in key or "wv" in key or "wo" in key:
         key = key.replace("wq", "query_proj")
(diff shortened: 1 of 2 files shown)


ricardo-larosa:
Yes, you will need at least 32GB of memory for the 7B model.

llama/convert.py
```diff
@@ -32,11 +33,12 @@ def map_torch_to_mlx(key, value):
     elif "rope" in key:
         return None, None
 
-    return key, value.numpy()
+    return key, value.numpy() if value.dtype != torch.bfloat16 else value.to(torch.float32).numpy()
```
tanliboy, line 36: Should we use float16`value.to(torch.float16).numpy()` to avoid increasing the size?

tanliboy: With float16
```
  13476925163 Dec  6  consolidated.00.pth
  13476918398 Dec  6  mlx_llama_weights.npz
```

awni: I agree it's a good idea to use `float16` to save the weights. It will be half the size which is nice.  I did the same in the Mistral example, it works fine to store the weights in fp16.

ricardo-larosa: Changed to `torch.float16`
```
13476918398 Dec  7 07:49 mlx_llama-2-7b.npz
```

awni (approved):
Thank you!!

ricardo-larosa:
For reference there is an issue created in PyTorch https://github.com/pytorch/pytorch/issues/90574

tanliboy:
The bfloat16 uses 7 bits for precision and 8 bits for range; the float16 uses 10 bits for precision and 5 bits for range, meaning that it is actually more precise than the bfloat16. Fortunately, the llama2 weights fall within the float16 range, so it is a safe conversion for inference—in fact, huggingface uses float16 for Llama model inference IIRC. On the other hand, if you want to train the model, you should avoid using the float16 since it can result in nan errors.

PR #14, opened 2023-12-06
convert.py requires numpy

gavi:
`convert.py` requires numpy which is not present in `requirements.txt`


angeloskath:
Oh sorry for merging a later PR I started reading them top to bottom. Thanks for the contribution though!

Solved in #24.

PR #19, opened 2023-12-06
Fix: transformer_lm bug in eval_fn

### LINKED ISSUE #3: error in transformer_lm example
I tried to run the transformer LM example in `mlx-examples/transformer_lm/main.py` as follows:
```
python main.py --gpu
```
which produced the output:
```
Training a transformer with 153.883 M parameters
Iter 10: Train loss 9.027, It/sec 2.114
Iter 20: Train loss 8.388, It/sec 2.272
...
Iter 990: Train loss 6.922, It/sec 2.280
Iter 1000: Train loss 6.916, It/sec 2.277
  File "/Users/vogel/data/mlx-examples/transformer_lm/main.py", line 192, in <module>
    main(args)
  File "/Users/vogel/data/mlx-examples/transformer_lm/main.py", line 115, in main
    val_loss = eval_fn(params, valid)
                       ^^^^^^
NameError: name 'params' is not defined. Did you mean: 'nparams'?
```

From a quick skim of the code, I suspect `model` should be passed to `eval_fn()` instead of `params`, e.g.:
```
val_loss = eval_fn(model, valid)
```
and `eval_fn()` should be something like:
```
def eval_fn(model, dataset):
        inputs, targets = map(mx.array, to_samples(context_size, dataset))
        loss = 0
        for s in range(0, targets.shape[0], batch_size):
            bx, by = inputs[s : s + batch_size], targets[s : s + batch_size]
            bx, by = map(mx.array, (bx, by))
            losses = model.loss(bx, by, reduce=False)
            loss += mx.sum(losses).item()
        return loss / len(targets)
```

LeonEricsson:
Fixes #3

### DIFF
--- a/transformer_lm/main.py
+++ b/transformer_lm/main.py
@@ -81,13 +81,13 @@ def main(args):
     optimizer = optim.SGD(learning_rate=args.learning_rate)
     loss_and_grad_fn = nn.value_and_grad(model, model.loss)
 
-    def eval_fn(params, dataset):
+    def eval_fn(model, dataset):
         inputs, targets = map(mx.array, to_samples(context_size, dataset))
         loss = 0
         for s in range(0, targets.shape[0], batch_size):
             bx, by = inputs[s : s + batch_size], targets[s : s + batch_size]
             bx, by = map(mx.array, (bx, by))
-            losses = self.loss(bx, by, reduce=False)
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
Thanks for fixing that. It does seem we didn't let it reach eval for a while now :-)

PR #23, opened 2023-12-06
feat: benchmark all models (with `--all` flag)

adhishthite:
While running a benchmark, if the user adds the `--all` flag,

```
python benchmark.py  --all
```

Then, all the models - `[tiny, base, small, medium, large]` will be benchmarked.

Also, updated `.gitignore` to ignore `.vscode/` and `.idea/`

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,5 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+.idea/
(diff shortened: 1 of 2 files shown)


awni (approved):
LGTM, thanks!

awni (approved):

PR #43, opened 2023-12-08
BERT implementation

jbarrow:
Added and documented the BERT implementation. It currently uses the HF tokenizer, because that handles a lot of nice things (like padding, masking, and token_type_ids).

It's tested against the HuggingFace implementation at: batch sizes >= 1, different models (`bert-base-uncased`, `bert-base-cased`, `bert-large-uncased`, `bert-large-cased`), etc.


bert/README.md
```diff
@@ -0,0 +1,52 @@
+# mlxbert
+
+An implementation of BERT [(Devlin, et al., 2019)](https://aclanthology.org/N19-1423/) within mlx.
```
awni, line 3: Nit: we like to use capital `MLX` in text.

bert/model.py
```diff
@@ -0,0 +1,246 @@
...
+
+    print("MLX BERT:")
+    print(mlx_output)
+
+    print("\n\nMLX Pooled:")
+    print(mlx_pooled[0, :20])
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Convert BERT weights to MLX.")
```
awni, line 231: Could you change the description to be about running the model?

bert/model.py
```diff
@@ -0,0 +1,246 @@
...
+    parser.add_argument(
+        "--bert-model",
+        type=str,
+        default="bert-base-uncased",
+        help="The huggingface name of the BERT model to save.",
```
awni, line 236: I like the `choices` kwarg for this sort of thing. You could add it and list the available models and it will handle making sure the user supplies one in the list.

bert/model.py
```diff
@@ -0,0 +1,246 @@
...
+    )
+    parser.add_argument(
+        "--mlx-model",
+        type=str,
+        default="weights/bert-base-uncased.npz",
+        help="The output path for the MLX BERT weights.",
```
awni, line 242: Nit: change `help` to `The path to the MLX BERT weights.`, "output" makes it sounds like its going to save something there.

bert/model.py
```diff
@@ -0,0 +1,246 @@
...
+        "This is an example of BERT working on MLX.",
+        "A second string",
+        "This is another string.",
+    ]
+
+    model = Bert(model_configs[bert_model])
+
+    weights = mx.load(mlx_model)
+    weights = tree_unflatten(list(weights.items()))
+    weights = tree_map(lambda p: mx.array(p), weights)
```
awni, line 210: This line is superflous, `mx.load` loads `mx.array`s

bert/model.py
```diff
@@ -0,0 +1,246 @@
...
+
+class BertEmbeddings(nn.Module):
+    def __init__(self, config: ModelArgs):
+        self.word_embeddings = nn.Embedding(config.vocab_size, config.intermediate_size)
+        self.token_type_embeddings = nn.Embedding(2, config.intermediate_size)
+        self.position_embeddings = nn.Embedding(
+            config.max_position_embeddings, config.intermediate_size
+        )
+        self.norm = nn.LayerNorm(config.intermediate_size, eps=config.layer_norm_eps)
+
```
awni, line 166: Nit: there's some inconsistency between when you use `nn.Layer` or just plain `Layer` (you imported both `nn` and the individual layers at the top. I don't mind one or the other, but could you make them consistent?

bert/model.py
```diff
@@ -0,0 +1,246 @@
...
+    model = Bert(model_configs[bert_model])
+
+    weights = mx.load(mlx_model)
+    weights = tree_unflatten(list(weights.items()))
+    weights = tree_map(lambda p: mx.array(p), weights)
+
+    model.update(weights)
+
+    tokenizer = AutoTokenizer.from_pretrained(bert_model)
```
awni, line 214: Could you wrap this into a single `load_model` function? It would be really nice for a user of this example to be able to do something like:

```
model, tokenizer = load_model(...)
output = model(tokenizer.encode(text))
```

awni: Also maybe you could add a shrot section in the README on how to use the model like that and what the outputs are (the shape / meaning)?

awni (commented):
This is really nicely done!! I love it. I left some mostly minor comments. 

Please address and then we can merge it into the examples. 

Also could you add a `requirements.txt` for dependenies?

bert/model.py
```diff
@@ -0,0 +1,246 @@
...
+        intermediate_size=1024, num_attention_heads=16, num_hidden_layers=24
+    ),
+}
+
+
+class MultiHeadAttention(Module):
+    """
+    Minor update to the MultiHeadAttention module to ensure that the
+    projections use bias.
+    """
```
awni, line 44: Is the `bias` really the only change from the `nn` version? If so we should just add an option to that version.. would greatly reduce duplication!

bert/convert.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def convert(bert_model: str, mlx_model: str) -> None:
+    model = BertModel.from_pretrained(bert_model)
+    # save the tensors
+    tensors = {
+        replace_key(key): tensor.numpy() for key, tensor in model.state_dict().items()
+    }
+    numpy.savez(mlx_model, **tensors)
+    # save the tokenizer
```
awni, line 29: stary comment?

bert/model.py
jbarrow: I added an additional helper function to the class to convert the HuggingFace tokenizer's {0, 1} mask to an additive causal mask, but that doesn't need to exist **on** the `MultiHeadAttention` class. So, yes, the existing MLX multi-head attention implementation could simply take a `use_bias` param and I could remove it from this implementation. I can test it out and submit a PR to both placed if it works.

bert/model.py
jbarrow: Swapped everything over the `nn`, so it doesn't depend on package internals.

bert/model.py
awni: Perfect! Though I would use the name `bias` rather than `use_bias` for consistency with our `Linear` layer etc.

jbarrow:
Outside of anything that requires a change to MLX core, I think I've made the changes.

I'll put off removing the `MultiHeadAttention` class until I've made that change, it's been pulled into MLX, and there's been a released version. 😄

jbarrow:
Submitted and tested a PR for the bias change in `mlx`. I can submit a new PR here once it's merged/released.

bert/README.md
```diff
@@ -0,0 +1,76 @@
...
+Which will show the following outputs:
+```
+MLX BERT:
+[[[-0.17057164  0.08602728 -0.12471077 ... -0.09469379 -0.00275938
+    0.28314582]
+  [ 0.15222196 -0.48997563 -0.26665813 ... -0.19935863 -0.17162783
+   -0.51360303]
+  [ 0.9460105   0.1358298  -0.2945672  ...  0.00868467 -0.90271163
+   -0.2785422 ]]]
+```
```
awni, line 58: I get different values for MLX Bert and HF Bert. Could you double check these are correct / up to date on your machine? I don't think there should be any non-determinism in the forward..

```
 HF BERT:
[[[-0.5250895  -0.19938672 -0.28210217 ... -0.6112563   0.19114657
    0.8227602 ]
  [-0.8783872  -0.37107885 -0.52238107 ... -0.5067151   1.0847608
    0.31066865]
  [-0.70010006 -0.54245055 -0.26593646 ... -0.26886988  0.3833898
    0.6557663 ]
```

bert/model.py
```diff
@@ -0,0 +1,248 @@
...
+            dims=config.intermediate_size,
+            num_heads=config.num_attention_heads,
+        )
+        self.pooler = nn.Linear(config.intermediate_size, config.vocab_size)
+
+    def __call__(
+        self,
+        input_ids: mx.array,
+        token_type_ids: mx.array,
+        attention_mask: mx.array | None = None,
```
awni, line 189: This is Python >= 3.10, so let's stick with Optional for now so it runs on >=3.8. (Also I didn't know about this new type hint style, neat!)

awni (commented):
Ok, just two more comments, then we can merge this and follow up after the change to core multiheadedattention.

bert/README.md
jbarrow: Ah, correct, those values were from when the example input was "test". 🤦 

Updating the README.

awni (approved):
Brilliant, thank you for putting this together!

PR #45, opened 2023-12-08
📝 Clarify python command for llama example

bbelescot:
Thanks for this project, a minor documentation change i propose after installing the project and giving a shot to some examples. The llama python instruction was unclear. I propose to put a 'path', just like for the tokenizer on the same command line.


bbelescot:
@awni yes definitely, i'll add it

bbelescot:
done :)

PR #48, opened 2023-12-08
Update CODE_OF_CONDUCT.md

nepalivai:
Typo fixes


awni:
As per other thread since this is an Apple general code of conduct, we should leave it as is.

PR #51, opened 2023-12-09
Update BERT to take advantage of bias param in MultiHeadAttention

jbarrow:
Reduced duplication in the implementation. Tested on my machine (with a local mlx build) and gives the same results as before and as the HF implementation.

### DIFF
--- a/bert/model.py
+++ b/bert/model.py
@@ -7,7 +7,6 @@ import mlx.core as mx
 import mlx.nn as nn
 import argparse
 import numpy
-import math
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


bert/model.py
```diff
@@ -186,11 +117,26 @@ def __call__(
...
+
         y = self.encoder(x, attention_mask)
         return y, mx.tanh(self.pooler(y[:, 0]))
+    
+
+    def convert_mask_to_additive_causal_mask(
+        self, mask: mx.array, dtype: mx.Dtype = mx.float32
+    ) -> mx.array:
+        mask = mask == 0
+        mask = mask.astype(dtype) * -1e9
```
awni, line 137: BTW if the mask is 0/s and 1/s you could do `mx.log(mask).astype(dtype)` instead.

awni (approved):
Love to see all those red lines. I'll wait till we do the next release then merge this.

jbarrow:
A few more red lines thanks to your mask suggestion. 😄

Before you merge, but after there's a new MLX release I'll set a minimum `mlx` version in `requirements.txt`.

jbarrow:
Updated `requirements.txt`!

PR #52, opened 2023-12-09
Mistral: Pass argument --tokens_per_eval for token generation

ricardo-larosa:
Pass the batch size of token generation via argument `--tokens_per_eval`

```
python mistral.py --model_path model --prompt "hello" --temp 0 --max_tokens 256 --tokens_per_eval 1
```

### DIFF
--- a/mistral/mistral.py
+++ b/mistral/mistral.py
@@ -247,6 +247,12 @@ if __name__ == "__main__":
         type=float,
         default=1.0,
     )
+    parser.add_argument(
+        "--tokens_per_eval",
+        help="The batch size of tokens to generate.",
+        type=int,
+        default=10,
+    )
     parser.add_argument("--seed", type=int, default=0, help="The PRNG seed")
 
     args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


mistral/mistral.py
```diff
@@ -247,6 +247,12 @@ def sample(logits):
         type=float,
         default=1.0,
     )
+    parser.add_argument(
+        "--batch_size",
```
awni, line 251: I'm fine to make this an argument, but I think the name `batch_size` may mislead some people as it usually refers to batching over different examples not within one example.

Could you change it to something like `tokens_per_eval` or something a little more specific?

awni (changes requested):
Thanks! Minor request then we can merge it! 🚀

mistral/mistral.py
ricardo-larosa: Sounds good, changed to `tokens_per_eval`

awni (approved):
thanks!

Issue #39, opened 2023-12-08
llama lora fune tune using previous adapters weights


thegodone:
Can we reuse a previous adapters.npz in a second fine tune process iteration ?

let say I fine tune for 600 iteration during the first fine tune process, I want to restart the fine tune a second time but I want to benefits of the first run adapters.npz weights. Can we add an option to load them in order to go close to the previous fine tune version of model ?

awni:
Done in #16

You can load the previously saved adapter weights using the argument `--resume_adapter_file <path_to_adapters.npz>`

Issue #16, opened 2023-12-06
Mistral Lora example


zakkor:
Would be nice to get a Mistral 7B Lora example, similar to the Llama Lora one. Thanks!

awni:
We will get this done ASAP.

awni:
Mistral is now supporting in the LoRA fine tuning example: https://github.com/ml-explore/mlx-examples/pull/53

https://github.com/ml-explore/mlx-examples/tree/main/lora

PR #53, opened 2023-12-09
Generalize lora finetuning to Mistral

awni:
- Changed the lora fine-tuning example to work with either llama or mistral models. Since they are the same arch basically they use the same model file. #16 
- Added an option to resume from an already trained adapter #39


bert/model.py
```diff
@@ -214,19 +215,29 @@ def run(bert_model: str, mlx_model: str):
...
-    mlx_output, mlx_pooled = model(**tokens)
-    mlx_output = numpy.array(mlx_output)
-    mlx_pooled = numpy.array(mlx_pooled)
-
-    print("MLX BERT:")
-    print(mlx_output)
-
-    print("\n\nMLX Pooled:")
-    print(mlx_pooled[0, :20])
+    vs = model_configs[bert_model].vocab_size
```
angeloskath, line 222: Were the BERT changes supposed to be in this PR.

angeloskath (approved):
Everything looks great to me. Not sure if the BERT changes were supposed to be in this PR though (besides the style fixes)...

bert/model.py
awni: Ugh good catch, that was not intentional.. just doing some timing for myself :).

PR #69, opened 2023-12-11
Add Graph Convolutional Network example

TristanBilot:


gcn/datasets.py
```diff
@@ -0,0 +1,114 @@
...
+
+        os.remove(file_path)
+
+
+def train_val_test_mask(labels, num_classes):
+    """Splits the loaded dataset into train/validation/test sets."""
+
+    train_set = mx.array(list(range(140)))
+    validation_set = mx.array(list(range(200, 500)))
+    test_set = mx.array(list(range(500, 1500)))
```
awni, line 46: Nit: use `mx.arange`

gcn/datasets.py
```diff
@@ -0,0 +1,114 @@
...
+
+        # Extract the .tgz file
+        with tarfile.open(file_path, "r:gz") as tar:
+            tar.extractall(path=extract_to)
+        print(f"Cora dataset extracted to {extract_to}")
+
+        os.remove(file_path)
+
+
+def train_val_test_mask(labels, num_classes):
```
awni, line 41: These function parameters are unused?

gcn/datasets.py
```diff
@@ -0,0 +1,114 @@
...
+
+    return train_set, validation_set, test_set
+
+
+def enumerate_labels(labels):
+    """Converts the labels from the original
+    string form to the integer [0:MaxLabels-1]
+    """
+    unique = list(set(labels))
+    labels = np.array([unique.index(label) for label in labels])
```
awni, line 56: Cleaner / more scalable to do something like:

```
label_map = dict{v : e for e, v in enumerate(set(labels))}
labels = np.array([label_map[label] for label in labels])
```

gcn/datasets.py
```diff
@@ -0,0 +1,114 @@
...
+    return labels
+
+
+def normalize_adjacency(adj):
+    """Normalizes the adjacency matrix according to the
+    paper by Kipf et al.
+    https://arxiv.org/pdf/1609.02907.pdf
```
awni, line 63: Nit: use the abs version here as well https://arxiv.org/abs/1609.02907

gcn/main.py
```diff
@@ -0,0 +1,121 @@
...
+
+def loss_fn(y_hat, y, weight_decay=0.0, parameters=None):
+    l = mx.mean(nn.losses.cross_entropy(y_hat, y))
+
+    if weight_decay != 0.0:
+        assert parameters != None, "Model parameters missing for L2 reg."
+
+        l2_reg = mx.zeros(
+            1,
+        )
```
awni, line 21: This formatting looks a little off, make sure you run `black` (see e.g. the [CONTRIBUTIN.md](https://github.com/ml-explore/mlx-examples/blob/main/CONTRIBUTING.md) file on the pre-commit hooks for more info)

gcn/main.py
```diff
@@ -0,0 +1,121 @@
...
+        for leaf in tree_flatten(parameters):
+            l2_reg += mx.sum(leaf[1] ** 2)
```
awni, line 23: This can be done more simply with:
```
l2_reg = sum(mx.sum(p**2) for p in tree_flatten(parameters)).sqrt()
```

gcn/main.py
```diff
@@ -0,0 +1,121 @@
...
+        l2_reg = mx.sqrt(l2_reg)
+        return l + weight_decay * l2_reg.item()
```
awni, line 25: You don't want to call `.item()` here, that will trigger the graph to eval which will slow things down.

gcn/datasets.py
```diff
@@ -0,0 +1,114 @@
...
+https://github.com/tkipf/gcn
+https://github.com/senadkurtisi/pytorch-GCN/tree/main
+"""
+
+
+def download_cora():
+    """Downloads the cora dataset into a local cora folder."""
+
+    url = "https://linqs-data.soe.ucsc.edu/public/lbc/cora.tgz"
+    extract_to = "."
```
awni, line 20: Could you add `cora/` to a `.gitignore` in this example since the default is to download it in the same dir.

awni (changes requested):
This is amazing, thanks for adding it!! I left a few comments, please address then we can merge it into the examples!

gcn/datasets.py
awni: Also just curious, how did you choose those numbers? I notice the dataset is 2708 examples so it looks like you are dropping some samples?

TristanBilot:
thanks for your returns! I just pushed the fixes.

awni (approved):
Awesome, looks great!

PR #75, opened 2023-12-12
Mixtral

awni:
Runs on an M2 Ultra at about 8ish tokens per second

-Current hypothesis is that improving memory use will speed things up a lot (quantization, etc)
-Can also try a `vmap` to avoid the evals in each layer


angeloskath (approved):
Beautiful

PR #77, opened 2023-12-12
Added CIFAR-10 + ResNet example

SarthakYadav:
Added an implementation for training ResNets for CIFAR-10, as specified in the original [paper](https://arxiv.org/abs/1512.03385), as currently there aren't any CNN examples.

Since `mlx` currently doesn't have a native `BatchNorm` implementation, using `LayerNorm` instead, will add options for switching to `BatchNorm` in the future. 

Supports the following ResNet configurations: ResNet-20, ResNet-32, ResNet-44, ResNet-56, ResNet-110, ResNet-1202.

Also showcases using `mlx-data` for data loading/augmentation.

### DIFF
--- a/cifar/README.md
+++ b/cifar/README.md
@@ -1,9 +1,13 @@
 # CIFAR and ResNets
 
-An example of training a ResNet on CIFAR-10 with MLX. Several ResNet configurations in accordance with the original [paper](https://arxiv.org/abs/1512.03385) are available. Also illustrates how to use `mlx-data` to download and load the dataset.
-
+An example of training a ResNet on CIFAR-10 with MLX. Several ResNet
+configurations in accordance with the original
+[paper](https://arxiv.org/abs/1512.03385) are available. The example also
+illustrates how to use [MLX Data](https://github.com/ml-explore/mlx-data) to
+load the dataset.
 
 ## Pre-requisites
+
 Install the dependencies:
 
 ```
@@ -11,6 +15,7 @@ pip install -r requirements.txt
 ```
 
 ## Running the example
+
 Run the example with:
 
 ```
@@ -29,23 +34,18 @@ For all available options, run:
 python main.py --help
 ```
 
-
-## Throughput
-
-On the tested device (M1 Macbook Pro, 16GB RAM), I get the following throughput with a `batch_size=256`:
-```
-Epoch: 0 | avg. tr_loss 2.074 | avg. tr_acc 0.216 | Train Throughput: 415.39 images/sec
-```
-
-When training on just the CPU (with the `--cpu` argument), the throughput is significantly lower (almost 30x!):
-```
-Epoch: 0 | avg. tr_loss 2.074 | avg. tr_acc 0.216 | Train Throughput: 13.5 images/sec
-```
-
 ## Results
-After training for 100 epochs, the following results were observed:
+
+After training with the default `resnet20` architecture for 100 epochs, you
+should see the following results:
+
 ```
-Epoch: 99 | avg. tr_loss 0.320 | avg. tr_acc 0.888 | Train Throughput: 416.77 images/sec
-Epoch: 99 | test_acc 0.807
+Epoch: 99 | avg. Train loss 0.320 | avg. Train acc 0.888 | Throughput: 416.77 images/sec
+Epoch: 99 | Test acc 0.807
 ```
-At the time of writing, `mlx` doesn't have in-built `schedulers`, nor a `BatchNorm` layer. We'll revisit this example for exact reproduction once these features are added.
+
+Note this was run on an M1 Macbook Pro with 16GB RAM.
+
+At the time of writing, `mlx` doesn't have built-in learning rate schedules,
+nor a `BatchNorm` layer. We intend to update this example once these features
+are added.
--- a/cifar/README.md
+++ b/cifar/README.md
@@ -1,11 +1,10 @@
 # CIFAR and ResNets
 
-* This example shows how to run ResNets on CIFAR10 dataset, in accordance with the original [paper](https://arxiv.org/abs/1512.03385).
-* Also illustrates how to use `mlx-data` to download and load the dataset.
+An example of training a ResNet on CIFAR-10 with MLX. Several ResNet configurations in accordance with the original [paper](https://arxiv.org/abs/1512.03385) are available. Also illustrates how to use `mlx-data` to download and load the dataset.
 
 
 ## Pre-requisites
-* Install the dependencies:
+Install the dependencies:
 
 ```
 pip install -r requirements.txt
@@ -21,7 +20,7 @@ python main.py
 By default the example runs on the GPU. To run on the CPU, use: 
 
 ```
-python main.py --cpu_only
+python main.py --cpu
 ```
 
 For all available options, run:
@@ -29,3 +28,24 @@ For all available options, run:
 ```
 python main.py --help
 ```
+
+
+## Throughput
+
+On the tested device (M1 Macbook Pro, 16GB RAM), I get the following throughput with a `batch_size=256`:
+```
+Epoch: 0 | avg. tr_loss 2.074 | avg. tr_acc 0.216 | Train Throughput: 415.39 images/sec
+```
+
+When training on just the CPU (with the `--cpu` argument), the throughput is significantly lower (almost 30x!):
+```
+Epoch: 0 | avg. tr_loss 2.074 | avg. tr_acc 0.216 | Train Throughput: 13.5 images/sec
+```
+
+## Results
+After training for 100 epochs, the following results were observed:
+```
+Epoch: 99 | avg. tr_loss 0.320 | avg. tr_acc 0.888 | Train Throughput: 416.77 images/sec
+Epoch: 99 | test_acc 0.807
+```
+At the time of writing, `mlx` doesn't have in-built `schedulers`, nor a `BatchNorm` layer. We'll revisit this example for exact reproduction once these features are added.
--- a/cifar/README.md
+++ b/cifar/README.md
@@ -47,5 +47,5 @@ Epoch: 99 | Test acc 0.807
 Note this was run on an M1 Macbook Pro with 16GB RAM.
 
 At the time of writing, `mlx` doesn't have built-in learning rate schedules,
-nor a `BatchNorm` layer. We intend to update this example once these features
+or a `BatchNorm` layer. We intend to update this example once these features
 are added.
--- a/cifar/README.md
+++ b/cifar/README.md
@@ -1,0 +1,31 @@
+# CIFAR and ResNets
+
+* This example shows how to run ResNets on CIFAR10 dataset, in accordance with the original [paper](https://arxiv.org/abs/1512.03385).
+* Also illustrates how to use `mlx-data` to download and load the dataset.
+
+
+## Pre-requisites
+* Install the dependencies:
+
+```
+pip install -r requirements.txt
+```
+
+## Running the example
+Run the example with:
+
+```
+python main.py
+```
+
+By default the example runs on the GPU. To run on the CPU, use: 
+
+```
+python main.py --cpu_only
+```
+
+For all available options, run:
+
+```
+python main.py --help
+```
--- a/cifar/dataset.py
+++ b/cifar/dataset.py
@@ -36,4 +36,4 @@ def get_cifar10(batch_size, root=None):
     num_tr_steps_per_epoch = num_tr_samples // batch_size
     num_test_steps_per_epoch = num_test_samples // batch_size
 
-    return tr_iter, test_iter, num_tr_steps_per_epoch, num_test_steps_per_epoch
+    return tr_iter, test_iter
--- a/cifar/dataset.py
+++ b/cifar/dataset.py
@@ -1,0 +1,39 @@
+import mlx.core as mx
+from mlx.data.datasets import load_cifar10
+import math
+
+
+def get_cifar10(batch_size, root=None):
+
+    tr = load_cifar10(root=root)
+    num_tr_samples = tr.size()
+
+    mean = mx.array([0.485, 0.456, 0.406]).reshape((1, 1, 3))
+    std = mx.array([0.229, 0.224, 0.225]).reshape((1, 1, 3))
+
+    tr_iter = (
+        tr.shuffle()
+        .to_stream()
+        .image_random_h_flip("image", prob=0.5)
+        .pad("image", 0, 4, 4, 0.0)
+        .pad("image", 1, 4, 4, 0.0)
+        .image_random_crop("image", 32, 32)
+        .key_transform("image", lambda x: (x.astype("float32") / 255.0))
+        .key_transform("image", lambda x: (x - mean) / std)
+        .batch(batch_size)
+    )
+
+    test = load_cifar10(root=root, train=False)
+    num_test_samples = test.size()
+
+    test_iter = (
+        test.to_stream()
+        .key_transform("image", lambda x: (x.astype("float32") / 255.0))
+        .key_transform("image", lambda x: (x - mean) / std)
+        .batch(batch_size)
+    )
+
+    num_tr_steps_per_epoch = num_tr_samples // batch_size
+    num_test_steps_per_epoch = num_test_samples // batch_size
+
+    return tr_iter, test_iter, num_tr_steps_per_epoch, num_test_steps_per_epoch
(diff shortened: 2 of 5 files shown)


cifar/main.py
```diff
@@ -0,0 +1,108 @@
...
+
+    print("num_params: {:0.04f} M".format(model.num_params() / 1e6))
+    mx.eval(model.parameters())
+
+    optimizer = optim.Adam(learning_rate=args.lr)
+
+    for epoch in range(args.epochs):
+        # get data every epoch
+        # or set .repeat() on the data stream appropriately
+        train_data, test_data, tr_batches, _ = get_cifar10(args.batch_size)
```
awni, line 93: It looks like 1tr_batches` is unused, could you remove it

cifar/dataset.py
```diff
@@ -0,0 +1,39 @@
...
+        test.to_stream()
+        .key_transform("image", lambda x: (x.astype("float32") / 255.0))
+        .key_transform("image", lambda x: (x - mean) / std)
+        .batch(batch_size)
+    )
+
+    num_tr_steps_per_epoch = num_tr_samples // batch_size
+    num_test_steps_per_epoch = num_test_samples // batch_size
+
+    return tr_iter, test_iter, num_tr_steps_per_epoch, num_test_steps_per_epoch
```
awni, line 39: This last two outputs are unused, I would just remove them for now to keep things simple.

cifar/main.py
```diff
@@ -0,0 +1,108 @@
...
+        losses.append(loss_value)
+        accs.append(acc_value)
+
+        if batch_counter % 10 == 0:
+            print(
+                f"Epoch {epoch:02d}[{batch_counter:03d}]: tr_loss {loss_value:.3f}, tr_acc {acc_value:.3f}"
+            )
+
+    mean_tr_loss = np.mean(np.array(losses))
+    mean_tr_acc = np.mean(np.array(accs))
```
awni, line 62: Use `mx` here

cifar/main.py
```diff
@@ -0,0 +1,108 @@
+import argparse
+import resnet
+import numpy as np
```
awni, line 3: You could probably remove numpy from this example to simplify things. Looks like it is barely used and easily replaceable with mlx

cifar/main.py
```diff
@@ -0,0 +1,108 @@
...
+        acc = eval_fn(model, x, y)
+        acc_value = acc.item()
+        accs.append(acc_value)
+    mean_acc = np.mean(np.array(accs))
+
+    return mean_acc
+
+
+def main(args):
+    np.random.seed(args.seed)
```
awni, line 80: Remove `np.seed`, don't use numpy RNG

cifar/main.py
```diff
@@ -0,0 +1,108 @@
...
+    mx.random.seed(args.seed)
+
+    model = resnet.__dict__[args.arch]()
+
+    print("num_params: {:0.04f} M".format(model.num_params() / 1e6))
+    mx.eval(model.parameters())
+
+    optimizer = optim.Adam(learning_rate=args.lr)
```
awni, line 88: I thought it was more standard to use SGD + momentum for Resnets? Does Adam work better? If not, could you go with the original work here?

cifar/requirements.txt
```diff
@@ -0,0 +1,3 @@
+mlx
+mlx-data
+numpy
```
awni, line 3: Remove numpy.

cifar/resnet.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def __call__(self, x):
+        return mx.pad(
+            x[:, ::2, ::2, :],
+            pad_width=[(0, 0), (0, 0), (0, 0), (self.dims // 4, self.dims // 4)],
+        )
+
+
+class Block(nn.Module):
+    expansion = 1
```
awni, line 41: This is always fixed at 1, can we remove to simplify things.

cifar/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+```
+python main.py --cpu_only
+```
+
+For all available options, run:
+
+```
+python main.py --help
+```
```
awni, line 31: Could you add a section on results. And a little table with some train / valid values over the course of training + the final test set accuracy. Also an approximate images / sec would be helpful so people know it's running at an expected speed.

cifar/resnet.py
```diff
@@ -0,0 +1,137 @@
...
+        self.layer3 = self._make_layer(block, 64, num_blocks[2], stride=2)
+
+        self.linear = nn.Linear(64, num_classes)
+
+    def _make_layer(self, block, dims, num_blocks, stride):
+        strides = [stride] + [1] * (num_blocks - 1)
+        layers = []
+        for stride in strides:
+            layers.append(block(self.in_dims, dims, stride))
+            self.in_dims = dims * block.expansion
```
awni, line 99: Let's remove `self.in_dims` and just use `dims`, since it's always fixed at `dims`.

cifar/resnet.py
```diff
@@ -0,0 +1,137 @@
...
+class ResNet(nn.Module):
+    """
+    Creates a ResNet model for CIFAR-10, as specified in the original paper.
+    """
+
+    def __init__(self, block, num_blocks, num_classes=10):
+        super().__init__()
+        self.conv1 = nn.Conv2d(3, 16, kernel_size=3, stride=1, padding=1, bias=False)
+        self.bn1 = nn.LayerNorm(16)
+        self.in_dims = 16
```
awni, line 86: Remove.

cifar/resnet.py
```diff
@@ -0,0 +1,137 @@
...
+            in_dims, dims, kernel_size=3, stride=stride, padding=1, bias=False
+        )
+        self.bn1 = nn.LayerNorm(dims)
+
+        self.conv2 = nn.Conv2d(
+            dims, dims, kernel_size=3, stride=1, padding=1, bias=False
+        )
+        self.bn2 = nn.LayerNorm(dims)
+
+        if stride != 1 or in_dims != dims:
```
awni, line 60: This should really just be a check for `stride != 1` since the shortcut hardcodes that assumption.

awni (changes requested):
Really nice! I especially love that you are using `mlx.data`, it's great to have an example with that.

Could you address the comments and then we can merge? Thank!!

cifar/README.md
```diff
@@ -0,0 +1,31 @@
...
+Run the example with:
+
+```
+python main.py
+```
+
+By default the example runs on the GPU. To run on the CPU, use: 
+
+```
+python main.py --cpu_only
```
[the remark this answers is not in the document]
SarthakYadav: done

cifar/main.py
```diff
@@ -0,0 +1,108 @@
...
+    "--arch",
+    type=str,
+    default="resnet20",
+    help="model architecture [resnet20, resnet32, resnet44, resnet56, resnet110, resnet1202]",
+)
+parser.add_argument("--batch_size", type=int, default=128, help="batch size")
+parser.add_argument("--epochs", type=int, default=100, help="number of epochs")
+parser.add_argument("--lr", type=float, default=1e-3, help="learning rate")
+parser.add_argument("--seed", type=int, default=0, help="random seed")
+parser.add_argument("--cpu_only", action="store_true", help="use cpu only")
```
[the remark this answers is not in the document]
SarthakYadav: Done

cifar/main.py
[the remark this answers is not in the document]
SarthakYadav: Done. I couldn't find a reference to .reset() in the `mlx-data` documentation. Maybe that's worth highlighting along with a full example in the official docs?

cifar/main.py
SarthakYadav: Yeah, it indeed SGD is used. But then if we're going for full one-to-one reproduction we effectively need a learning rate schedule as well. Since there are no schedulers in `mlx` yet I didn't want to do that to keep things simple, and I assumed if I am not doing that Adam is a better choice.

cifar/resnet.py
SarthakYadav: Well I decided to keep this because extension to bigger ResNets for ImageNet based on this is trivial for Bottleneck blocks etc (and it's also how PyTorch implements ResNets). But sure, can be removed.

cifar/resnet.py
SarthakYadav: We need to have one variable dims that keeps track of changing dimensions between blocks. Or did you mean something else?

cifar/resnet.py
SarthakYadav: sure. fixed.

cifar/README.md
SarthakYadav: I have a base M1 MacBook, so I cannot run full training (usually ~330 epochs with SGD). But sure, I'll add some sample logs and final accuracy value for the amount of epochs I do run.

cifar/main.py
awni: Sounds good 👍

cifar/resnet.py
awni: I would just have it be an input to `_make_layer` rather than a member of the class which is a little confusing..

cifar/resnet.py
awni: Have `in_dims` be an input to `_make_layer` rather than a member of the class.

SarthakYadav:
> @SarthakYadav I think you need to push your updates as I don't see any of the changes yet.

Just did :). Let me know if this works!

cifar/resnet.py
```diff
@@ -0,0 +1,135 @@
+"""
+Implementation of ResNets for CIFAR-10 as per the original paper [https://arxiv.org/abs/1512.03385].
+Configurations include ResNet-20, ResNet-32, ResNet-44, ResNet-56, ResNet-110, ResNet-1202.
+
+There's no BatchNorm is mlx==0.0.4, using LayerNorm instead.
+
+Authors:
+    Sarthak Yadav, 2023
```
awni, line 8: Hey Sarthak, I really want to make sure you get credit for this, but I think we should avoid putting individual authors in code files as that will be really difficult to maintain (and also hard to make fair to future developers who edit and improve this example).

For now could you please remove this?

I will discuss with the team about how we can best assign credit to our many contributors and then follow up and make sure to get your name added where it makes sense!

SarthakYadav: > For now could you please remove this?

Of course, that's not a problem. Out of habit more than anything, this was the only file I wrote a header for 😅. There usually isn't an "established" way to do this between open-source frameworks (SpeechBrain does it this way) and that's fine!

awni (approved):
Thanks a ton for putting this example together, this is really great. 

Any interest in making it work on ImageNet? 

PS I added a few updates, before merging.

SarthakYadav:
> PS I added a few updates, before merging.

Yep, saw them. Thanks!

> Any interest in making it work on ImageNet?

That's gonna be a challenge on my tiny Macbook, but I sure! I can put a working implementation together, maybe you can run it on better a machine!

hbaghar:
Thanks for this example @SarthakYadav. I am trying to benchmark MLX vs torch on my M1 by comparing this script against the one [here](https://pytorch-tutorial.readthedocs.io/en/latest/tutorial/chapter03_intermediate/3_2_2_cnn_resnet_cifar10/) by changing the model to `model = ResNet(ResidualBlock, [3, 3, 3]).to(device)`. I found the MLX version you have shared to be much slower compared to the torch example I have shared.

Any idea why this might be happening? And how we can address this? In case it is useful, also sharing the `line_profiler` stats I got:

```
Total time: 821.054 s
File: mlx_main.py
Function: train_epoch at line 32

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    32                                           @profile
    33                                           def train_epoch(model, train_iter, optimizer, epoch):
    34         5          2.0      0.4      0.0      def train_step(model, inp, tgt):
    35                                                   output = model(inp)
    36                                                   loss = mx.mean(nn.losses.cross_entropy(output, tgt))
    37                                                   acc = mx.mean(mx.argmax(output, axis=1) == tgt)
    38                                                   return loss, acc
    39                                           
    40         5       2045.0    409.0      0.0      train_step_fn = nn.value_and_grad(model, train_step)
    41                                           
    42         5          0.0      0.0      0.0      losses = []
    43         5          0.0      0.0      0.0      accs = []
    44         5          0.0      0.0      0.0      samples_per_sec = []
    45                                           
    46      3915    8895686.0   2272.2      1.1      for batch_counter, batch in enumerate(pbar:= tqdm(train_iter, unit="batch")):
    47      3910     144089.0     36.9      0.0          x = mx.array(batch["image"])
    48      3910      19232.0      4.9      0.0          y = mx.array(batch["label"])
    49      3910       3318.0      0.8      0.0          tic = time.perf_counter()
    50      3910    7510315.0   1920.8      0.9          (loss, acc), grads = train_step_fn(model, x, y)
    51      3910    4995520.0   1277.6      0.6          optimizer.update(model, grads)
    52      3910  797797923.0 204040.4     97.2          mx.eval(model.parameters(), optimizer.state)
    53      3910       9897.0      2.5      0.0          toc = time.perf_counter()
    54      3910       5851.0      1.5      0.0          throughput = 1/(toc - tic)
    55      3910     874069.0    223.5      0.1          loss = loss.item()
    56      3910     768844.0    196.6      0.1          acc = acc.item()
    57      3910       5954.0      1.5      0.0          losses.append(loss)
    58      3910       1911.0      0.5      0.0          accs.append(acc)
    59      3910       1413.0      0.4      0.0          samples_per_sec.append(throughput)
    60      3910       5601.0      1.4      0.0          if batch_counter % 100 == 0:
    61        80      10685.0    133.6      0.0              tqdm.write(
    62        80         92.0      1.1      0.0                  " | ".join(
    63        40         20.0      0.5      0.0                      (
    64        40        707.0     17.7      0.0                          f"Epoch {epoch:02d} [{batch_counter:03d}]",
    65        40        108.0      2.7      0.0                          f"Train loss {loss:.3f}",
    66        40         26.0      0.7      0.0                          f"Train acc {acc:.3f}",
    67        40         24.0      0.6      0.0                          f"Throughput: {throughput:.2f} batch/second",
    68                                                               )
    69                                                           )
    70                                                       )
    71                                           
    72         5        284.0     56.8      0.0      mean_tr_loss = mx.mean(mx.array(losses))
    73         5         70.0     14.0      0.0      mean_tr_acc = mx.mean(mx.array(accs))
    74         5        103.0     20.6      0.0      samples_per_sec = mx.mean(mx.array(samples_per_sec))
    75         5          1.0      0.2      0.0      return mean_tr_loss, mean_tr_acc, samples_per_sec

```

awni:
The mlx version should be way faster than torch on the cpu. I'm having a hard time parsing the output you shared. Could you report the throughput recorded by the `main.py` script (e.g. how many images per second you are getting)?

hbaghar:
Ah, I see. I figured that the MLX code should be running on gpu so I ran the torch code using the `mps` accelerator. For a batch size of 64 I get close to ~4.8-5 batches/sec which is around 310-320 images/sec. Using torch on cpu I get ~2.7 batches/sec.

Can you elaborate a little on why you mentioned this should be faster only on cpu? And is there a way to make MLX use hardware acceleration similar to `mps`?

awni:
Oh cool - what about torch with the `mps` backend? What's the throughput for that one?

For convolutions we may be a little slower depending on the size but it should not be that different. We frankly haven't spent a lot of time optimizing convolutions so we are leaving some perf on the table (especially for certain sizes).

SarthakYadav:
@hbaghar I haven't benchmarked the PyTorch version on my MacBook, but the implementation you posted looks correct. The throughput may be simply what it is, but it's also possible that there's something wrong with my environment. To be sure, I'll try cleaning things up and redoing the tests in a new Python environment. I don't think in terms of my implementation there isn't much that could go wrong since the network architecture is quite simple, and I confirmed the model sizes with a [reference](https://github.com/akamaster/pytorch_resnet_cifar10). 

Nonetheless, I'll recheck things on my end.

thegodone:
One comment, generally Layernorm is faster than Bacthnorm to compute. 
Can you add a pretrain section to reload models like Resnet50 pretrain on imageNet 1000 tasks from keras or torch ? https://pytorch.org/vision/main/models/resnet.html or https://www.tensorflow.org/api_docs/python/tf/keras/applications/resnet50/ResNet50

awni:
@thegodone I like the idea of supporting pre-trained Resnets. Do you want to file an issue for that so we can prioritize / track it?

PR #79, opened 2023-12-12
Enable FP16 for Whisper

awni:
Enable FP16 for Whisper and make it the default (following the original implementation).


angeloskath (approved):
Lovely and possibly big impact in memory constrained devices.

Issue #56, opened 2023-12-10
llama.py failed to generate on large mode.


Haixing-Hu:
For example, the program reports the following error on llama-2-13b-chat:

```
python3 llama.py llama-2-13b-chat.npz tokenizer.model 'Hello, who are you?'
[INFO] Loading model from disk.
Press enter to start generation
------
Traceback (most recent call last):
  File "/Users/starfish/working/llm/mlx-examples/llama/llama.py", line 306, in <module>
    generate(args)
  File "/Users/starfish/working/llm/mlx-examples/llama/llama.py", line 180, in generate
    for token in model.generate(x, args.temp):
  File "/Users/starfish/working/llm/mlx-examples/llama/llama.py", line 125, in generate
    x, c = l(x, mask=mask)
           ^^^^^^^^^^^^^^^
  File "/Users/starfish/working/llm/mlx-examples/llama/llama.py", line 77, in __call__
    y = self.norm1(x)
        ^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/mlx/nn/layers/normalization.py", line 85, in __call__
    return self.weight * x * n
           ~~~~~~~~~~~~^~~
ValueError: Shapes (5120) and (1,7,2560) cannot be broadcast.
```

 And the program reports the following error on llama-2-70b:

```shell
python3 llama.py llama-2-70b.npz tokenizer.model 'Hello, who are you?'
[INFO] Loading model from disk.
Press enter to start generation
------
Traceback (most recent call last):
  File "/Users/starfish/working/llm/mlx-examples/llama/llama.py", line 306, in <module>
    generate(args)
  File "/Users/starfish/working/llm/mlx-examples/llama/llama.py", line 180, in generate
    for token in model.generate(x, args.temp):
  File "/Users/starfish/working/llm/mlx-examples/llama/llama.py", line 125, in generate
    x, c = l(x, mask=mask)
           ^^^^^^^^^^^^^^^
  File "/Users/starfish/working/llm/mlx-examples/llama/llama.py", line 77, in __call__
    y = self.norm1(x)
        ^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/mlx/nn/layers/normalization.py", line 85, in __call__
    return self.weight * x * n
           ~~~~~~~~~~~~^~~
ValueError: Shapes (8192) and (1,7,1024) cannot be broadcast.
```

awni:
[...]

PR #82, opened 2023-12-12
llama v2 with sharded weights

awni:
Enable support for llama v2 and larger llama models with sharded weights to address #56 

I am moving this example to be more like the `mistral` / `mixtral` example just for the sake of consistency. Still keeping the nicely commented generation loop since that's pretty useful IMO.


angeloskath (approved):
🚀

PR #84, opened 2023-12-12
Update convert.py

iammerrick:
Docs are right, however, the code has a typo.


awni (approved):
Oops, thanks for fixing!

PR #87, opened 2023-12-13
Whisper - fixed errors in test and benchmark.

Stv-X:
Current version of [whisper/test.py](https://github.com/ml-explore/mlx-examples/blob/main/whisper/test.py) may cause `test_fp16(self)` to fail, fixed.
```
FAIL: test_fp16 (__main__.TestWhisper.test_fp16)
----------------------------------------------------------------------
  File "/mlx-examples/whisper/test.py", line 64, in test_fp16
    self.assertEqual(logits.dtype, mx.float16)
AssertionError: float32 != float16
```
Current version of [whisper/benchmark.py](https://github.com/ml-explore/mlx-examples/blob/main/whisper/benchmark.py) may cause type error in `decode(model, mels)` and `everything()`, fixed.
```
  File "/mlx-examples/whisper/whisper/decoding.py", line 550, in _get_audio_features
    raise TypeError(
TypeError: audio_features has an incorrect dtype: float32
```

### DIFF
--- a/whisper/test.py
+++ b/whisper/test.py
@@ -5,13 +5,11 @@ import unittest
 import mlx.core as mx
 import numpy as np
 import os
-import subprocess
 import torch
 
 import whisper
 import whisper.audio as audio
 import whisper.load_models as load_models
-import whisper.torch_whisper as torch_whisper
 import whisper.decoding as decoding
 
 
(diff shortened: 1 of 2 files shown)


whisper/test.py
```diff
@@ -62,7 +60,7 @@ def test_fp16(self):
         dims = mlx_model.dims
         mels = mx.array(np.random.randn(1, 3_000, dims.n_mels), mx.float16)
         tokens = mx.array(np.random.randint(0, dims.n_vocab, (1, 20)), mx.int32)
-        logits = mlx_model(mels, tokens)
+        logits = mlx_model(mels, tokens).astype(mx.float16)
```
awni, line 63: Casting to fp16 here defeats the point of the test which is to make sure the enitre inference is in fp16... the test failed for you when you run it without this?

awni: Ah, I think the problem is you are using an old mlx which was incorrectly upcasting, we will put out a new release today and see if that fixes the issue.

awni:
Closing this to prefer the fix in #90

PR #88, opened 2023-12-13
fix "request access" form url for Llama models

dastrobu:
this PR makes the docs on Llama model downloads consistent for lora and llama examples.


awni (approved):
Nice, thanks!

PR #90, opened 2023-12-13
Fix whisper fp16 inference

bofenghuang:
Hi @awni,

Thanks for working on enabling fp16 for the whisper model! However, the inference still fails with the following error:

```
  File "/Users/bofenghuang/Projects/mlx-examples/whisper/tmp.py", line 5, in <module>
    result = whisper.transcribe(
  File "/Users/bofenghuang/Projects/mlx-examples/whisper/whisper/transcribe.py", line 244, in transcribe
    result: DecodingResult = decode_with_fallback(mel_segment)
  File "/Users/bofenghuang/Projects/mlx-examples/whisper/whisper/transcribe.py", line 176, in decode_with_fallback
    decode_result = model.decode(segment, options)
  File "/Users/bofenghuang/Projects/mlx-examples/whisper/whisper/decoding.py", line 724, in decode
    result = DecodingTask(model, options).run(mel)
  File "/Users/bofenghuang/Projects/mlx-examples/whisper/whisper/decoding.py", line 612, in run
    audio_features: mx.array = self._get_audio_features(mel)  # encoder forward pass
  File "/Users/bofenghuang/Projects/mlx-examples/whisper/whisper/decoding.py", line 550, in _get_audio_features
    raise TypeError(
TypeError: audio_features has an incorrect dtype: float32
```

As well as the unit test `TestWhisper.test_fp16`

```
F
======================================================================
FAIL: test_fp16 (test.TestWhisper)
----------------------------------------------------------------------
  File "/Users/bofenghuang/Projects/mlx-examples/whisper/test.py", line 66, in test_fp16
    self.assertEqual(logits.dtype, mx.float16)
AssertionError: float32 != float16

----------------------------------------------------------------------
Ran 1 test in 0.974s

FAILED (failures=1)
```

I found this is caused by `x / math.sqrt(2)` in `nn.gelu` function which casts dtype from float16 to float32.

My workaround is re-cast dtype to float16 after each gelu function. This works for me on both `transcribe` function and the unit test.

Feel free to ignore this PR if you prefer to fix at another level ;)

### DIFF
--- a/whisper/whisper/whisper.py
+++ b/whisper/whisper/whisper.py
@@ -117,7 +117,7 @@ class ResidualAttentionBlock(nn.Module):
         if self.cross_attn:
             y, cross_kv = self.cross_attn(self.cross_attn_ln(x), xa, kv_cache=cross_kv)
             x += y
-        x = x + self.mlp2(nn.gelu(self.mlp1(self.mlp_ln(x))))
+        x = x + self.mlp2(nn.gelu(self.mlp1(self.mlp_ln(x))).astype(x.dtype))
         return x, (kv, cross_kv)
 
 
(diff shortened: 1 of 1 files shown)


awni:
Yes, I think the issue is the version of MLX has a bug in the divide which is inadvertently casting to fp32. Which is why your fix is what it is :). I'm happy to land this at least until we do release of MLX, then we can put a minimum version in the requirements for this example!

bofenghuang:
@awni Thanks for the details. Excited to see how this project unfolds!

PR #94, opened 2023-12-13
Update README.md to fix git-lfs command

jbax3:
Trying to help folks out who miss the need to specifically call git-lfs


awni (approved):
👍

Issue #92, opened 2023-12-13
Mixtral conversion to nzp not working on M3 max 128gb


technopremium:
Hi, 

I have an M3 max 128gb and I wanted to try mixral, but after downloading the weights and combining them, I tried to run the converter and it start but then the process gets killed. 

Do we need more than 128gb for this to work? 

Thanks in advance.

awni:
No, it shouldn't need more than 128 GB to do the conversion... the conversion only uses max about 100GB, I just measured it.

I recommend asitop to view RAM usage as the conversion is running: https://github.com/tlkh/asitop

awni:
100 GB is still a lot, so can work on reducing the RAM for conversion as I address #81

foysavas:
For anyone else struggling with this and also on a M3 Max 128GB, you need to get your resting memory usage down below 11GB before running the weight conversion script.

lafintiger:
I am on a Mac studio m2 ultra with 192gb of ram.
Here is what I am getting when I run convert.py

Any suggestions would be appreciated.

(mlx) lafintiger@VincentacStudio mixtral % python convert.py --model_path mixtral-8x7b-32kseqlen
Traceback (most recent call last):
  File "/Users/lafintiger/aidev/mlx-examples/mixtral/convert.py", line 19, in <module>
    state = torch.load(str(model_path / "consolidated.00.pth"))
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lafintiger/anaconda3/envs/mlx/lib/python3.11/site-packages/torch/serialization.py", line 1028, in load
    return _legacy_load(opened_file, map_location, pickle_module, **pickle_load_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lafintiger/anaconda3/envs/mlx/lib/python3.11/site-packages/torch/serialization.py", line 1246, in _legacy_load
    magic_number = pickle_module.load(f, **pickle_load_args)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_pickle.UnpicklingError: invalid load key, 'v'.
(mlx) lafintiger@VincentacStudio mixtral %

lafintiger:
[...]

Issue #81, opened 2023-12-12
Support the Mistral AI official huggingface weights for Mixtral-8x7B-v0.1


lostmygithubaccount:
per the current example:

> Download the models from HuggingFace:
> git clone https://huggingface.co/someone13574/mixtral-8x7b-32kseqlen

I'd really rather not have to re-download many GBs of weight files, and less-so from `someone13574` (no offense) when there weights are posted by Mistral AI itself: https://huggingface.co/mistralai/Mixtral-8x7B-Instruct-v0.1

would this work already? could the example be updated to use these weights?

awni:
Yea it's a good point. Will look into updating the example to use the official weights instead of the ones from MixtralKit

thegodone:
One major issue how to run it with 64 GB memory ? This is a huge model, we need to think of reducing the precision. Is it also possible to think of delta in the weights between the chat (instruction version Mixtral-8x7B-Instruct-v0.1) and the navite generator (mixtral-8x7b-32kseqlen). But I don't see an easy way so far to run on M1 64 GB.  running the convert give me a terminal killed message

```
(tf) tgg@gvalmu00008 mixtral % python convert.py --model_path mixtral-8x7b-32kseqlen/
zsh: killed     python convert.py --model_path mixtral-8x7b-32kseqlen/
```

caseybasichis:
I'm able to get the someone13574 running.

I attempted the cat/convert process on the official instruct model, but no dice.

Is there a way to modify the convert script to get the instruct version going?

awni:
This one right? https://huggingface.co/mistralai/Mixtral-8x7B-Instruct-v0.1

I'll test it after getting the official Mistral one working

caseybasichis:
That's the one.

awni:
#107

PR #107, opened 2023-12-14
Use official HF for mixtral

awni:
- Use the official Mixtral Hugging Face repo
- Weight conversion is now streaming so it uses quite a bit less RAM in total (~30GB instead of 100+)

Issue #92 and #81


angeloskath (approved):
Looks great

lostmygithubaccount:
running a bit slow, but running for me!

PR #108, opened 2023-12-14
Phi-2: Stop generating at eos token

devonthomas35:


devonthomas35:
Example before change:
```
python phi2.py --prompt "Instruct:How was gpt-3 trained?\nOutput:" --max_tokens 2048
[INFO] Generating with Phi-2...
Instruct:How was gpt-3 trained?\nOutput:gpt-3 was trained using a large amount of text data from the internet, which was used to train a deep learning model. The model was trained using a technique called reinforcement learning, which involves giving the model feedback on its performance and adjusting its parameters to improve its accuracy. The model was trained for several years using a combination of human-written text and machine-generated text.
<|endoftext|>Instruction: Write a short summary of the main idea and key points of the following paragraph.
Input: The human brain is composed of billions of neurons, which communicate with each other through electrical and chemical signals. These signals form complex networks that enable various cognitive functions, such as memory, learning, attention, and emotion. The brain is also divided into different regions that specialize in different tasks, such as vision, language, movement, and reasoning. The brain is constantly changing and adapting to new experiences and stimuli, a process known as neuroplasticity.
Output: The paragraph explains the basic structure and function of the human brain, and how it can change and learn from the environment.
<|endoftext|>User: Write a short summary of the main idea and key points of the following paragraph. The human brain is composed of billions of neurons, which communicate with each other through electrical and chemical signals. These signals form complex networks that enable various cognitive functions, such as memory, learning, attention, and emotion. The brain is also divided into different regions that specialize in different tasks, such as vision, language, movement, and reasoning. The brain is constantly changing and adapting to new experiences and challenges, a process known as neuroplasticity.
```
After: 
```
python phi2.py --prompt "Instruct:How was gpt-3 trained?\nOutput:" --max_tokens 2048
Tokenizer has EOS token: <|endoftext|>
[INFO] Generating with Phi-2...
Instruct:How was gpt-3 trained?\nOutput:gpt-3 was trained using a large amount of text data from the internet, which was used to train a deep learning model. The model was trained using a technique called reinforcement learning, which involves giving the model feedback on its performance and adjusting its parameters to improve its accuracy. The model was trained for several years using a combination of human-written text and machine-generated text.
```

phi2/phi2.py
```diff
@@ -202,7 +202,10 @@ def load_model():
...
     tokens = []
     for token, _ in zip(generate(prompt, model), range(args.max_tokens)):
-        tokens.append(token)
+        if token == tokenizer.eos_token_id:
+            break
+        else:
+            tokens.append(token)
 
         if (len(tokens) % 10) == 0:
             mx.eval(tokens)
```
awni, line 211: You should do the check inside the conditional otherwise you are evaling the model for every token.

awni (changes requested):
Thanks for catching this! See the comment / let me know if it makes sense. I can make the change if it's not clear.

phi2/phi2.py
devonthomas35: Maybe I am misunderstanding. I am not changing how often the eval happens. That seems like an unrelated change.

awni: No you are implicitly changing it :). When you do `token == tokenizer...` we `mlx` will implicitly evaluate the result of `token` under the hood in order to use it for the control flow in the program. If you did not have that line then `token` is not evaluated until line 211 when we explicitly call `mx.eval`.

devonthomas35: ah correct due to lazy eval

phi2/phi2.py
```diff
@@ -206,9 +206,16 @@ def load_model():
...
+            eos_index = next((i for i, t in enumerate(tokens) if t.item() == tokenizer.eos_token_id), None)
```
awni, line 209: Neat! Also too bad `index` doesn't have a non-throwing version...

awni (approved):
Thanks for the fix!!

PR #110, opened 2023-12-15
feat: stop phi-2 output at <|endoftext|>

lostmygithubaccount:
I'm not 100% sure this is correct, but it seems to help

stop the output of phi-2 when "<|endoftext|>" appears


lostmygithubaccount:
just noticed https://github.com/ml-explore/mlx-examples/pull/108 beat me to it and looks cleaner, closing

Issue #100, opened 2023-12-14
Mixtral example is giving non-sensical output


samikama:
It seems that mixtral example is giving non-sensical output when used with non-default arguments. Is this expected?
Here is an example
```
python mixtral.py --model_path ../../mixtral/ -m 2000 --prompt "keep continuing the following story. In the beginning the Universe was created."
[INFO] Loading model from disk.
[INFO] Starting generation...
keep continuing the following story. In the beginning the Universe was created. This has made a lot of people very angry and been widely regarded as a bad move"""
import random


def test(actual, expected):
    """
      >>> test(10, 10)
      True
    """
    if actual == expected:
        return True
    else:
        return False


def secret(a=1, b=3, c=6, d=12):
    x, y, z, w = a, b, c, d
    a, b, c, d = my_secret(a, b, c, d, y)
    b, y = my_Formatting(a, b, c, d, y)
    x, z = my_XOR(a, b, c, d, x, y, z, w)
    z, w = my_SecretNa(x, z, w)
    y, z, w = my_Formatting(x, y, z, w)
    w, x, y, z = my_XOR(x, y, z, w, x, y, z, w)
    y, z, w = my_Formatting(x, y, z, w)
    w, x, y, z = my_XOR(x, y, z, w, x, y, z, w)
    y, z, w = my_Formatting(x, y, z, w)
    w, x, y, z = my_XOR(x, y, z, w, x, y, z, w)
    b, c = my_NewSecret(x, y, z, w, b, c)
    w, z, x, y = my_NewInfo(a, b, c, d, w, z, x, y)
    b, y, z, x = my_NewChoice(a, b, c, d, b, y, z, x)
    task = my_NextTask(b, y, z, x, task)

    return task


def my_secret(a, b, c, d, x):
    a, b, c, d = 3, 2, 5, 4
    x, y, z, w = 100, 300, 500, 800
    b, y, z, x = 200, 100, 120, 1600

    return x


def my_Formatting(a, b, c, d, x):
    a, b, c, d = 3, 2, 5, 4
    x, y, z, w = 100, 300, 500, 800
    b, y, z, x = 200, 100, 120, 1600

    return x


def my_XOR(a, b, c, d, x, y, z, w):
    a, b, c, d = 3, 2, 5, 4
    x, y, z, w = 100, 300, 500, 800
    b, y, z, x = 200, 100, 120, 1600

    return x


def my_secret2(a, b, c, d, x):
[...]

PR #112, opened 2023-12-15
[Bugfix] Fix RoPE base bug in Mixtral example

awni:
- Set RoPE base correctly
- Default temp to 0 as it seems to work better
- Following [reference](https://github.com/mistralai/mistral-src/blob/main/moe_one_file_ref.py) upcast gate scores prior to softmax

#100

### DIFF
--- a/mixtral/mixtral.py
+++ b/mixtral/mixtral.py
@@ -41,6 +41,26 @@ class RMSNorm(nn.Module):
         return self.weight * output
 
 
+class RoPE(nn.RoPE):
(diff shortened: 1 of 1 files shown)


andresy (approved):

angeloskath (approved):
Nice. Maybe the base should be configurable in the nn.RoPE class. Either way that should be a different PR.

awni:
> Maybe the base should be configurable in the nn.RoPE class.

Yea I will make a PR for that (and then we can update this on the next version).

Issue #70, opened 2023-12-11
Error while finetuning with LoRA: Unable to allocate 12 bytes.


shamikbose:
I tried to follow the example in the `lora` directory, but I get the following error

`libc++abi: terminating due to uncaught exception of type std::runtime_error: [malloc_or_wait] Unable to allocate 12 bytes.`

I tried to run with the default settings and only changed the `iters` value to 100

`lora % python lora.py --model mlx_model --train --iters 100`

I get the following output before I get the error above:

```
Loading pretrained model
Total parameters 7243.436M
Trainable parameters 1.704M
Loading datasets
Training
```

awni:
32 GB might be a stretch for the Lora finetuning as it uses 32 bit precision right now. One thing you can try is changing to a lower precision (change [here](https://github.com/ml-explore/mlx-examples/blob/main/lora/lora.py#L265) to `mx.float16` or `mx.bfloat16`). I can't promise it will work because we haven't tested that yet.

We will make it an argument once we have had a chance to test it.

You can also try using a smaller batch size `--batch_size 1` should save a lot of memory, for example. But it will be slower.

shamikbose:
@awni @ivanfioravanti I managed to get it working with `mx.float16` and batch_size=2. This is what training currently looks like
<img width="479" alt="image" src="https://github.com/ml-explore/mlx-examples/assets/50837285/c83104b1-ecbd-4552-9a82-b29a52301947">

awni:
Oof 4 tokens per second is slow! For context the m2 ultra is 100x faster (with batch size 4) and 32-bit precision..

It feels like you may be swapping given how slow it is. I think for the lower memory machines we need something like Q-lora / reduce memory requirements for this to not be too too slow.

shamikbose:
Yeah, GPU usage seems to be quite low (~50%). I saw in a different issue discussion that quantization was also on the cards, so that would be generally quite helpful

awni:
#115 can help reduce memory requirements by letting you use fewer LoRA layers.

DamascusGit:
and where on 265 do i exactly make the change to mx.float16? I can't seem to figure that one out.

> #115 can help reduce memory requirements by letting you use fewer LoRA layers.
> 
> The README.md is also updated with some tips to reduce memory use.

awni:
If you pull the latest it defaults to `float16`.

Issue #68, opened 2023-12-11
How can I edit the training dataset for Lora?


nimesttech:
I'm trying to finetune on Lora, but I can't find where the Training Dataset is located so I can create my own. Anyone managed to do it on their own dataset?

awni:
This is the [data loading](https://github.com/ml-explore/mlx-examples/blob/main/lora/wikisql.py) part of the example.

Right now, the easiest way to change it is the following:

1. Make a new python file `mydata.py`
2. Put a class in it like `MyData` which holds your data in whatever format. As long as it has `__getitem__` (which returns a string sample) and `__len__` (which returns the length of the dataset) then it should work. See the [wikisql as an example](https://github.com/ml-explore/mlx-examples/blob/main/lora/wikisql.py#L92-L96)
3. Put a `load` function in `mydata.py` which returns the training, validation, and test sets, [example here](https://github.com/ml-explore/mlx-examples/blob/main/lora/wikisql.py#L14C2-L18)
4. Change the references to `wikisql` in `lora.py` to `mydata` 

Sorry it's a bit cumbersome right now, but its conceptually pretty simple.

We can write a json loader or a text file loader for custom datasets if that would be helpful..

shamikbose:
> We can write a json loader or a text file loader for custom datasets if that would be helpful..

That would be very helpful! I would like to help if possible

awni:
Happy to take a PR for this if you would like to implement it!

shamikbose:
Playing around with it now. Looks like the changes might be minimal if we wanted to use HF datasets. Can we use streaming datasets?

nimesttech:
One thing that would be REALLY healpful for a user point of view, would be to be able to determinate the path for a jsonl, csv or json file for the training dataset, similarly to how we do it on pytorch or on llm platforms such as openai for example

Having a json format for example {"prompt":<TEXT>, "completion":<TEXT>} 

I think it is easier that way then already trying to implement a complicated dataset that extracts SQL queries

justinh-rahb:
I just trained a 7B model on my workstation over my lunchbreak without having to rent the capability by the hour, which is very cool. A more generalized loader function for using alternative datasets would be extremely appreciated for further testing. I'm very impressed with what I've seen so far, and now I need to push it to it's logical conclusion :) This framework is going to be a game changer for ML development.

shamikbose:
[...]

PR #115, opened 2023-12-15
Customize dataset with lora

awni:
Adds a few changes:

- Change the data loader to read `.jsonl` files and make these for the Wikisql dataset
- Loop over training set if it is small to get to the requested number of iterations
- Configurable number of LoRA layers for memory control
- Change to keep base weights in fp16 to reduce memory pressure.
- Updates to README.md

#68 
#70


angeloskath (approved):
Very nice!

PR #116, opened 2023-12-15
phi-2: fix --temp/--seed arguments.

idoru:
`seed` was not being passed to generate(). 

This quick fix will let folks experiment with `--temp/--seed`!

### DIFF
--- a/phi2/phi2.py
+++ b/phi2/phi2.py
@@ -201,7 +201,7 @@ if __name__ == "__main__":
     print(args.prompt, end="", flush=True)
 
     tokens = []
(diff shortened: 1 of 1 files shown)


awni (approved):
Thanks, that was on oversight. You wrote something about fixing `--seed` is that supposed to be in this PR?

idoru:
Sorry if it was a bit confusing. I just meant that `--seed` did not have any effect until this fix because when the temperature is defaulted to 0 the `logits` are sampled with `argmax()` and not `random()`

PR #121, opened 2023-12-16
Rope theta to support Coda Llama

awni:
Use the same derived RoPE as Mixtral until we release the updates in `mlx.nn`.


angeloskath (approved):
Looks great!

PR #134, opened 2023-12-18
Add Qwen example

JuniMay:
This is a PR to add the "Qwen (通义千问)" example.

The architecture of Qwen is similar to Llama, except for the bias in the attention layers. Currently, this inference script has only been tested on the 1.8B and 7B version of Qwen.

The generation code for this script is adapted from the Phi-2 example.


awni:
This is awesome! We are discussing the best way to include this change without having too many very similar LLM examples which might lead to confusion / clutter. Will get back to you on it soon!

awni (approved):
Wonderful thanks for adding this! I moved it to a subdir `llms/` where we will plan to put similar llama-style llms for now. Also made a few minor stylistic changes. Tested everything and it runs great for me! 🚀

Issue #139, opened 2023-12-18
[llama]  --few-shot broken


dastrobu:
Running the few show example 

```
python llama.py llama-2-7b tokenizer.model --max-tokens 1000 --few-shot --prompt sample_prompt.txt
```

is now broken: 

```
Traceback (most recent call last):
  File "/.../mlx-examples/llama/llama.py", line 388, in <module>
    few_shot_generate(args)
  File "/...//mlx-examples/llama/llama.py", line 324, in few_shot_generate
    generate(prompt.replace("{}", question))
  File "/...//mlx-examples/llama/llama.py", line 297, in generate
    if len(tokens) >= args.num_tokens:
                      ^^^^^^^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'num_tokens'. Did you mean: 'max_tokens'?
```

awni:
Yup, that's a bug. Will fix!

dastrobu:
@awni or merge: https://github.com/ml-explore/mlx-examples/pull/140

awni:
Thank you!!

dastrobu:
@awni I also noted that the you need to pass the file name as `--prompt sample_prompt.txt`, which is a bit confusing. Shouldn't we refactor to pass the file as `--few-shot sample_prompt.txt`? I can create a different issue/PR if you like since it is not really related to this one.

awni:
Cool, that change sounds good to me! If it is unspecified it uses the `prompt`, otherwise it reads the few shot prompt from the provided file?

dastrobu:
Well currently, it always wants to read a file and fails on 
```
python llama.py $MODEL_PATH $TOKENIZER_MODEL_PATH --few-shot
```
with `FileNotFoundError: [Errno 2] No such file or directory: 'In the beginning the Universe was created.'` which is just confusing. 

I would probably always expect a file for the few-shot learning and ignore `--prompt` in this case as passing a long multiline few shot example as inline string to `--prompt` is a bit strange.

dastrobu:
See #141 addressing the [comment](https://github.com/ml-explore/mlx-examples/issues/139#issuecomment-1861255611) above.

PR #141, opened 2023-12-18
pass few shot file name to --few-shot 

dastrobu:
pass few shot file name to --few-shot to avoid interpreting --prompt once a string and once as filename.

relates to #139


angeloskath (approved):
Thanks I like this change a lot.

PR #144, opened 2023-12-19
Fix Lora with llama 2 from meta

awni:
As per title, just a couple minor fixes.

Fixes: https://github.com/ml-explore/mlx/issues/213


angeloskath (approved):
👍

PR #145, opened 2023-12-19
Add llms subdir + update README

awni:
- Move the similar LLM models to a subdirectory.
- Update README to include new models and reflect changes.
- Updated to use same pre-commit config as mlx


angeloskath (approved):
Looks great!

PR #147, opened 2023-12-19
T5: Change default dtype to bfloat16

jbochi:
I didn't manage to make it work with float16, but from my tests it works ok with bfloat16. There are some small discrepancies in the generated text, but I think it will usually be worth the speed-up.

Also:
- Adds myself to acknowledgments :)
- Adds param to serialize to float32 (I even considered making it the default. Serializing to float16 can result in some differences, even when running in float32).
- Sets default `n_positions` to 512, which is not set in some models, such as the t5 v1.1 series.

### DIFF
--- a/ACKNOWLEDGMENTS.md
+++ b/ACKNOWLEDGMENTS.md
@@ -6,3 +6,5 @@ with a short description of your contribution(s) below. For example:
 - Jane Smith: Added the `foo` example.
 
 MLX Examples was developed with contributions from the following individuals:
+
+- Juarez Bochi: Added support for T5 models.
--- a/t5/convert.py
+++ b/t5/convert.py
@@ -71,7 +71,7 @@ if __name__ == "__main__":
         help="The model data type.",
         type=str,
         choices=["float16", "float32"],
-        default="float16",
+        default="float32",
     )
     args = parser.parse_args()
     convert(args.model, args.dtype)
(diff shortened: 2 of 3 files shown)


t5/convert.py
```diff
@@ -64,5 +66,12 @@ def convert(model_name):
         help="Name of the T5 model.",
         default="t5-small",
     )
+    parser.add_argument(
+        "--dtype",
+        help="The model data type.",
+        type=str,
+        choices=["float16", "float32"],
+        default="float16",
```
awni, line 74: It's unfortunate that we can't save `bfloat16` that's something that will work once we have safetensor support :). If you think saving to `fp16` is a bad idea, feel free to change the default, I'm not wedded to `fp16` as the default.

ACKNOWLEDGMENTS.md
```diff
@@ -6,3 +6,5 @@ with a short description of your contribution(s) below. For example:
 - Jane Smith: Added the `foo` example.
 
 MLX Examples was developed with contributions from the following individuals:
+
+- Juarez Bochi: Added support for T5 models.
```
awni, line 10: 💯

t5/convert.py
jbochi: Oh right, I meant to change it, but forgot. Thanks!

Serializing to `bfloat16` with safetensors will be great! I think most of these models were trained on bfloat16, and the serialization as `float16` is quite destructive.

awni (approved):
LGTM, thanks!!

PR #149, opened 2023-12-20
add speculative decoding example for llama

andersonbcdefg:
Example that loads huggingface llama weights, and "engine" class that holds the draft and regular model to do speculative decoding. Basic test script so you can compare their speed on a prompt where SD should obviously help (repeating a monologue from Hamlet)


awni:
Looks super cool! Will try and get to this soon :)

awni:
I started looking at this. I added the ability to sample with a temperature of 0.0. The first thing I noticed is there is not complete overlap when using the same model and draft_model and a temp of 0.0. I think they should match completely in that case, let's fix that first.

I will take look at perf next..

andersonbcdefg:
Cool! Yeah not sure what's going on with the lack of perfect overlap. I checked that they _generally_ overlap when draft model and verifier model are the same, but don't know what would cause this edge case. I copied the sampling from the existing Llama example. I wouldn't be surprised if there are small bugs related to KV-cache, as that's the part I'm least familiar with / gave me the most pain. The logic of truncating it the right amount etc. required lots of mental shape rotation and debugging 😄

awni:
I think I have it working now, I think it was a minor off-by-one with the cache as you say. I will push it shortly, just want to confirm!

andersonbcdefg:
Nice!

awni:
Just a quick update. I pushed to your branch. It works now in the sense that I get the same output with and without speculative decoding. Using tiny llama and llama 7b it accepts 110 out of 125 tokens which seems pretty good to me

The main problem is it is slower with speculative decoding (about 9 seconds vs 11.5 seconds). I didn't really spend any time on perf yet so there could be some hidden bottleneck there. Will investigate that next.

andersonbcdefg:
I ran into this too, I initially was getting similar times for both and was going to make a note of that in my PR, but then when I started testing it again right before submitting the PR I was getting 15s for regular and 8s for speculative so I removed the note about performance. There's probably some ways to unlock faster perf but MLX is new territory for me so I'm not sure what to try. :) Prefill the draft model on CPU while prefilling big model on GPU? Idk haha.

awni:
Ok, so I made a bunch of changes. The two big ones being:
1. Switch to use T5 like the original paper. It has two nice properties: first a very small draft model (30M) and second the tasks tend to be a lot more constrained (encoder-decoder style) so speculative decoding seems to work better.
2. Using the original accept reject criteria based on the model and draft model probabilities.

I think we can merge this as an example as it's pretty cool. A big thing to figure out would be if / how we can use it for more open ended text generation.  I'll leave that as a future endeavor!

awni (approved):
🚀

PR #152, opened 2023-12-20
Add `--model_path` to phi-2 example script

pcuenca:
This is for symmetry with the rest of the example scripts, and to avoid workarounds like this one because of the hardcoded path: https://huggingface.co/mlx-community/phi-2/blob/674d36e7f00751d12543893cd62e2f6c733016d6/README.md?code=true#L33


awni (approved):
Nice, thanks!

PR #154, opened 2023-12-20
Add max_seq_length truncation support in iterate_batches function

wyanzhao:
Description:
This pull request is to add the functionality of setting the maximum sequence length for iterate_batches function during model training and evaluation.

Key Changes:
- Added `--max-seq-length` as a command-line argument in `build_parser`, enabling users to define the maximum sequence length for batch processing.
- Implemented truncation logic in `iterate_batches` to ensure that sequences longer than `max_seq_length` are appropriately truncated.
- Updated the `train` and `evaluate` functions to accept and utilize the `max_seq_length` parameter passed through the command line.
- Adjusted the invocation of the `evaluate` function within the main training loop and testing sections to include the new `max_seq_length` parameter.

The support of sequence length truncation is expected to prevent potential out-of-memory errors during intensive training workloads, especially when working with datasets containing exceptionally long sequences.

### DIFF
--- a/phi2/phi2.py
+++ b/phi2/phi2.py
@@ -1,4 +1,5 @@
 import argparse
+from pathlib import Path
 from typing import Optional
 from dataclasses import dataclass
 from mlx.utils import tree_unflatten
@@ -154,9 +155,10 @@ def generate(prompt: mx.array, model: Phi2, temp: Optional[float] = 0.0):
         yield y
 
 
-def load_model():
+def load_model(model_path: str):
     model = Phi2(ModelArgs())
-    weights = mx.load("weights.npz")
+    model_path = Path(model_path)
+    weights = mx.load(str(model_path / "weights.npz"))
     model.update(tree_unflatten(list(weights.items())))
     tokenizer = AutoTokenizer.from_pretrained("microsoft/phi-2", trust_remote_code=True)
     return model, tokenizer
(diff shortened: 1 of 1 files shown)


lora/lora.py
```diff
@@ -205,15 +211,18 @@ def iterate_batches(dset, tokenizer, batch_size, train=False):
...
                 for j in range(batch_size)
             ]
             lengths = [len(x) for x in batch]
 
+            # Ensure max length does not exceed max_seq_length
+            max_len = min(max(lengths), max_seq_length)
+            
             # Pad to the max length
-            batch_arr = np.zeros((batch_size, max(lengths)), np.int32)
+            batch_arr = np.zeros((batch_size, max_len), np.int32)
```
awni, line 225: I don't think we should do this. If someone has just a few long documents, then this could silently drop almost all of their data.

My suggestion instead is to pre split data as a preprocessing step prior to making the json files. The added benefit of doing it that way is split points might be applications specific.

One thing we could do here is add a warning, like if `max(lengths) > 2048` print a warning suggesting the user pre split their data to save memory.

awni (changes requested):
Thanks for thinking of this, it's likely a common issue!

That said, I would not simply truncate the examples the way you implemented it. If someone has just a few long documents, then this could silently drop almost all of their data.

My suggestion instead is to pre split data as a preprocessing step (as mentioned in the README) prior to making the json files. The added benefit of doing it that way is natural  split points might be applications specific.

One thing we could do here is add a warning, like if `max(lengths) > 2048` print a warning suggesting the user pre-split their data to save memory. Do you want to change it to that?

lora/lora.py
wyanzhao: Sure, I would add this to the code.

PR #157, opened 2023-12-20
Use config.json, add model_type

pcuenca:
This would go with PRs like this one to the model repos: https://huggingface.co/mlx-community/mistral-7B-v0.1/discussions/7

The advantages of standardizing on `config.json` instead of `params.json`, and adding a `model_type` field, would be:

- Better compatibility with repos in the Hugging Face Hub, easier retrieval of stats.
- Easy (hopefully) support of existing safetensors repos when https://github.com/ml-explore/mlx/pull/215 is ready. IIUC, Mistral safetensors files should work out of the box.
- Easier election of inference code down the line, based on `model_type`.

### DIFF
--- a/mistral/convert.py
+++ b/mistral/convert.py
@@ -1,6 +1,7 @@
 # Copyright © 2023 Apple Inc.
 
 import argparse
+import json
 import numpy as np
 from pathlib import Path
 import torch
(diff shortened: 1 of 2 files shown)


awni:
I think we need to also change `convert.py` to copy `params.json` to `config.json` if they go through the non-HF step so that path doesn't break.

pcuenca:
Ah yes, of course!

pcuenca:
@awni done :)

awni (approved):
🚀  Thanks!!

PR #158, opened 2023-12-20
Add config.json to Mixtral.

Vaibhavs10:
Similar to https://github.com/ml-explore/mlx-examples/pull/157

PR #159, opened 2023-12-20
Use config.json in llama

pcuenca:
Similar to #157, #158


llama/convert.py
```diff
@@ -134,5 +135,5 @@ def tiny_llama(model_path):
     weights, params = globals()[args.model_name](model_path)
     np.savez(str(model_path / "weights.npz"), **weights)
     if params is not None:
-        with open(model_path / "params.json", "w") as fid:
+        with open(model_path / "config.json", "w") as fid:
```
awni, line 138: @pcuenca sorry this example is a bit of a mess (trying to keep support for everything).

The Meta distributed llama models come with `params.json`. So if you load them in the `llama` convert (around L47) and just return them instead of returning `None`, that should work nicely

llama/convert.py
```diff
@@ -134,5 +135,5 @@ def tiny_llama(model_path):
     weights, params = globals()[args.model_name](model_path)
     np.savez(str(model_path / "weights.npz"), **weights)
     if params is not None:
```
awni, line 137: We can remove the check for `None` here also since we always want the `config.json` to be written now.

awni (commented):
Thanks!! Mind making a couple changes to support the Meta distributions?

llama/convert.py
pcuenca: My bad, I read the code too fast and now I see I only updated the `tiny_llama` case.

awni:
Sorry, I just merged #145 which moves some stuff around so you'll need to rebase. Shouldn't be any conflict.

llama/convert.py
```diff
@@ -124,7 +124,7 @@ def tiny_llama(model_path):
         help=(
             "Name of the model to convert. Use 'llama' for models in the "
             "Llama family distributed by Meta including Llama 1, Llama 2, "
-            "Coda Llama, and Llama chat."
```
awni, line 125: 👍

awni (approved):
🚀  thank you!

PR #161, opened 2023-12-20
Add support for byt5 models

jbochi:
The t5 example doesn't support the ByT5 series ([paper](https://arxiv.org/abs/2105.13626)) because it uses more layers in the decoder than in the encoder. Also, for some reason, we have to use `AutoTokenizer` instead of the `T5Tokenizer`.

The model is not fine-tuned, so the output is garbage, but I checked it matches HF's.

```sh
$ export MODEL="google/byT5-small" && python convert.py --dtype float32 --model $MODEL && python t5.py --model $MODEL --prompt "Life is like a box of chocolates."
Saving weights to google-byT5-small.npz
[INFO] Generating with T5...
Input:  Life is like a box of chocolates.
ÿ chocolates chocolates chocolates chocolates chocolates chocolates chocolates chocolates chocolates
Time: 0.33 seconds, tokens/s: 298.93
```

I also checked with [this](https://huggingface.co/kaiyuy/leandojo-lean4-tacgen-byt5-small) fine-tuned model:

```sh
$ export MODEL="kaiyuy/leandojo-lean4-tacgen-byt5-small" && python convert.py --dtype float32 --model $MODEL && python t5.py --model $MODEL --prompt "a b : ℕ
⊢ a + b = b + a"
Saving weights to kaiyuy-leandojo-lean4-tacgen-byt5-small.npz
[INFO] Generating with T5...
Input:  a b : ℕ
⊢ a + b = b + a
rw [add_comm, add_comm]
Time: 0.14 seconds, tokens/s: 173.76
```

### DIFF
--- a/t5/t5.py
+++ b/t5/t5.py
@@ -5,8 +5,8 @@ from typing import List, Optional, Tuple
 import mlx.core as mx
 import mlx.nn as nn
 import numpy as np
-from mlx.utils import tree_map, tree_unflatten
-from transformers import T5Config, T5Tokenizer
+from mlx.utils import tree_map, tree_flatten, tree_unflatten
+from transformers import T5Config, AutoTokenizer
 
 
 def _relative_position_bucket(
@@ -251,8 +251,9 @@ class TransformerDecoderLayer(nn.Module):
 class TransformerDecoder(nn.Module):
     def __init__(self, config: T5Config):
         super().__init__()
+        n_layers = getattr(config, "num_decoder_layers", config.num_layers)
         self.layers = [
-            TransformerDecoderLayer(config) for i in range(config.num_layers)
+            TransformerDecoderLayer(config) for i in range(n_layers)
         ]
         self.ln = RMSNorm(config.d_model, eps=config.layer_norm_epsilon)
         self.relative_attention_bias = RelativePositionBias(config, bidirectional=False)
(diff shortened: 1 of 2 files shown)


t5/t5.py
```diff
@@ -5,8 +5,8 @@
 import mlx.core as mx
 import mlx.nn as nn
 import numpy as np
-from mlx.utils import tree_map, tree_unflatten
-from transformers import T5Config, T5Tokenizer
+from mlx.utils import tree_map, tree_flatten, tree_unflatten
```
awni, line 8: Why include `tree_flatten`?

awni (approved):
Thanks for adding this! Could you remove the unused import and then would love to merge it.

t5/t5.py
jbochi: Ooops. My bad. I added it to confirm all weights were initialized and that the shapes matched. Perhaps that's something we could implement in `mlx.utils`?

awni: Yea I've been thinking about something like that. I've had annoying bugs with that in the past. Actually I was wondering if it made sense to have `Module.update` have a mode where it warns or throws if:

1. Not all the parameters were updated
2. If the shapes of updated parameters are unexpected.

jbochi: Nice, that would be great! I'm using this to diagnose any issues:

```python
    current_weights = tree_flatten(model.parameters())
    weights_to_load = list(weights.items())
    current_weights_dict = dict(current_weights)
    current_weights_keys = set(current_weights_dict.keys())
    weights_to_load_dict = dict(weights_to_load)
    weights_to_load_keys = set(weights_to_load_dict.keys())
    print("Missing weights: ", sorted(current_weights_keys - weights_to_load_keys))
    print()
    print("Weights ignored: ", sorted(weights_to_load_keys - current_weights_keys))
    for key in current_weights_keys & weights_to_load_keys:
        if weights_to_load_dict[key].shape != current_weights_dict[key].shape:
            print("Shape mismatch for key: ", key)
            print("Expected shape: ", current_weights_dict[key].shape)
            print("Loading shape: ", weights_to_load_dict[key].shape)
```

PR #162, opened 2023-12-20
Quantize example

awni:


llms/llama/llama.py
```diff
@@ -326,28 +331,36 @@ def generate(question):
...
+    if "hidden_dim" not in config:
+        config["hidden_dim"] = weights["layers.0.feed_forward.w1.weight"].shape[0]
+    if config.get("vocab_size", -1) < 0:
+        config["vocab_size"] = weights["output.weight"].shape[-1]
+    if "rope_theta" not in config:
+        config["rope_theta"] = 10000
+    unused = ["multiple_of", "ffn_dim_multiplier"]
+    for k in unused:
+        config.pop(k, None)
+    return config
```
awni, line 350: I don't like the need for this function. I think the way to fix it is to do this mapping once in the conversion script and save the updated `config.json`. Unfortunately that would require updating all the Hugging Face models which is a bit tedious :\

awni: FWIW that's the approach taken by the other examples, so they avoid this baggage.

llms/mixtral/convert.py
```diff
@@ -1,59 +1,152 @@
...
+
+    # Extract the subset of quantized weights:
+    all_weights = dict(tree_flatten(model.parameters()))
+    quantized_weights = {}
+    for k, v in all_weights.items():
+        if k not in weights:
+            continue
+        quantized_weights[k] = v
+        prefix = k.split(".")[:-1]
+        for qw in ["scales", "biases"]:
```
angeloskath, line 80: Nice.

llms/mixtral/mixtral.py
```diff
@@ -244,20 +244,33 @@ def decode(self, t: List[int]) -> str:
...
-    weights = tree_map(lambda p: p.astype(dtype), weights)
     model = Mixtral(model_args)
+    #    model.update(weights)
+    #    quantization = {
+    #        "group_size": 64,
+    #        "bits": 4,
+    #    }
+    if quantization is not None:
+        # TODO: Quantize gate matrices when < 32 tiles supported
+        print("QUANTIZING")
```
angeloskath, line 268: No `[INFO] Quantizing`?

llms/mixtral/convert.py
```diff
@@ -1,59 +1,152 @@
...
-        state = torch.load(tf)
-        new_state = {}
-        for k, v in state.items():
-            new_state.update(convert(k, v, args))
-        np.savez(str(model_path / f"weights.{e}.npz"), **new_state)
+        weights = convert(tf, config)
+        if args.quantize:
+            print("[INFO] Quantizing")
+            weights, config = quantize(weights, config, args)
+        np.savez(str(mlx_path / f"weights.{e}.npz"), **weights)
```
angeloskath, line 147: So currently using `mx.savez` without pre-evaluating would result in memory blowing up (bug 239 in mlx) but is there another reason to not use it? Did we encounter that before?

Should we change it to `mx.savez` after that is fixed?

angeloskath (approved):
This looks really good!

I left basically one comment on the use of `np.savez` ... We do have an oversight in `mx.savez` currently but I was wondering if that was intentional for some other reason.

For example, perhaps the convert scripts were not using mlx before.

llms/mixtral/convert.py
awni: Actually the reason is simpler: `np.savez` handles `np` and `mx` arrays whereas `mx.savez` can only save `mx` arrays. So I left it at `np` rather than try to convert them to `mx` arrays in cases where no quantization is used.

PR #164, opened 2023-12-21
update path to load weights

deven367:
Since the reorg in #145, the paths to the load the model weights need to be updated. This PR updates the path to load the weights for `phi2`.

image

cc @awni


awni (approved):

awni:
Thanks for the fix!!

PR #172, opened 2023-12-22
add deepseek coder example

mzbac:
The DeepSeek coder is currently the most popular state-of-the-art open source coding mode. It would be great to have the example to run it via mlx. I tried using the Llama example, but I encountered a few issues:
- It uses LlamaLinearScalingRotaryEmbedding, which is not supported by the current Llama example.
- It can't provide a SentencePiece model at the moment. Details can be found here: https://huggingface.co/deepseek-ai/deepseek-coder-33b-instruct/discussions/6

So, I thought it would be better to create a separate example for it instead of modifying the existing Llama example. However, I am happy to merge it with the Llama example if you think there is a good way to do so.

the scripts are adapted from Qwen and llama examples :)

### DIFF
--- a/llms/deepseek-coder/README.md
+++ b/llms/deepseek-coder/README.md
@@ -1,0 +1,30 @@
+# Deepseek Coder
+
+Deepseek Coder is an advanced series of code language models based on LLama architecture, trained from scratch on a massive corpus of 2T tokens, with a unique composition of 87% code and 13% natural language in both English and Chinese.
+
+### Setup
+
+Install the dependencies:
+
+```
+pip install -r requirements.txt
+```
+
+Next, download and convert the model. 
+```sh
+python convert.py --model-path <path_to_huggingface_model> --mlx-path <path_to_save_converted_model>
+```
+
+By default, the conversion script will save 
+the converted `weights.npz`, `tokenizer`, and `config.json` there in the mlx-path you speficied .
+
+
+### Run
+
+Once you've converted the weights to MLX format, you can interact with the
+Deepseek coder model:
+
+```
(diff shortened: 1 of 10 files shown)


awni:
Awesome!! 

I will give some thought about the best way to incorporate it into the existing llms exampes. It may not be for this PR (which we could do as a standalone example in LLMs), but I'm wondering if we can have a more general LLM example to which we can start to add new models. 

- We'd probably want to use the HF tokenizers since they should work with almost all models. 
- Each model would need a custom function to convert it to the MLX format but other than that the rest of the code could be shared.

mzbac:
A general example for LLM models would be great, it will be similar to what transformers did. I also think that we currently lack a consistent naming convention for config/model args across different LLM examples. I'm wondering if there's a way to unify the json config and model args somehow?

rosaboyle (approved):
Tested. Looks great.

mzbac:
@awni Please let me know if you want me to work on merging this with the llama example. Currently, I am not sure how to handle the configuration for rope scaling settings and wondering if you prefer the current configuration convention or unifying the configuration with all the llama examples.

awni:
@mzbac sorry for the delay. I think for now let's keep it as a standalone model. At some point soon, I will start a general implementation which we can then fold or incorporate some of the standalone models into. This will be good to have as a reference.

awni:
This is great, thanks for adding it! And I think it's pretty much good to go. I pushed a few changes for the sake of consistency with other llm examples (trying to keep everything as close as possible for a future integration).

awni (approved):
🚀

Issue #155, opened 2023-12-20
Contribute Hugging Face models to the MLX Community


awni:
We encourage you to join the [MLX Community](https://huggingface.co/mlx-community) and upload new MLX converted models and versions of existing models.

Vaibhavs10:
Love it! Here's some more information about uploading to the 🤗 hub. 

Once you've converted the checkpoints via the `convert.py` files in the respective model directory, you can upload large files over to the hub as follows:

First, make sure to install `huggingface_hub` via `pip install --upgrade huggingface_hub`.

Second, run `huggingface-cli login` and paste a `write` token over from `https://huggingface.co/settings/tokens`.

Next, run the below code:

```python

from huggingface_hub import HfApi
from huggingface_hub import logging

logging.set_verbosity_info()

api = HfApi()

api.upload_folder(folder_path="<LOCAL FOLDER PATH>", 
                  repo_id="mlx-community/<MODEL_NAME>", 
                  repo_type="model", 
                  multi_commits=True, 
                  multi_commits_verbose=True)
```

maxtheman:
Is there any more information about what's needed to author a `convert.py` for a given model? I'm seeing a lot of similarities between them in terms of loading the weights and then grouping into a .npz file but I'm not clear what's driving all of the differences.

Bert needs lots of keys replaced: https://github.com/ml-explore/mlx-examples/blob/main/bert/convert.py
Phi needs some: https://github.com/ml-explore/mlx-examples/blob/main/llms/phi2/convert.py
Mistral doesn't need any, very straightforward: https://github.com/ml-explore/mlx-examples/blob/main/llms/mistral/convert.py
and Mixtral has a bunch of stuff going on, but I think mostly for handling the MoE architecture.

In my case — I was just looking at converting Mamba to MLX. I think I have a good reference implementation for running the model but I'm having trouble understanding how I would convert the weights appropriately to test it. Looking at the repo on HF it's not obvious to me what I should be keying in on to start making decisions. https://huggingface.co/state-spaces/mamba-370m/tree/main

Any suggestions are appreciated.

awni:
It's a bit more of an art at this point, we've been pushing for some standardization / consistency which is why things look the same. But there are still some subtle differences.

The short answer to your question is:

The output of `convert.py` should be two things:

1. `weights.npz` (or multiple weights files for large models)
[...]

PR #174, opened 2023-12-22
Llama/unshard on load

dastrobu:
Similar to #92 I noticed that converting the llama-2 70b models takes quite a bit of RAM (succeeded at around 140GB on an 128GB machine with swaping). 

However, the resulting huge weights files are still very hard to handle (e.g. upload them to HF is impossible and would required extra steps).

So I think changing the conversion algorithm a bit: keep the shards on model conversion and then unshard the weights on loading. This would be more RAM efficient and file size friendly. 

This PR 
 * is backwards compatible (as far as I could test it), so already converted weights will still work. 
 * should also enable #155. 
 * follows the implementation of #107 as much as it can.

I tested locally with tiny llama, llama-2-13b-chat and llama-2-70b-chat. The largest 70b model now takes around 16GB on average while converting, with peaks around 32GB. On inference it still requires around 128 GB, which makes sense, given that weights are around 128 GB on disk. With a bit of swapping one can run it on a 128GB machine, though not really productive on an Apple M3 Max 128GB:

```
[INFO] Prompt processing: 206.761 s
[INFO] Full generation: 36003.395 s
```

### DIFF
--- a/llms/llama/convert.py
+++ b/llms/llama/convert.py
@@ -15,7 +15,6 @@ import torch
 from llama import Llama, ModelArgs, sanitize_config
 from mlx.utils import tree_flatten, tree_map, tree_unflatten
 
-
 def llama(model_path):
     SHARD_FIRST = ["wv", "wq", "wk", "w1", "w3", "output"]
     SHARD_SECOND = ["tok_embeddings", "wo", "w2"]
@@ -140,6 +139,22 @@ def quantize(weights, config, args):
     return quantized_weights, quantized_config
 
 
+def make_shards(weights: dict, max_file_size_gibibyte: int = 15):
+    max_file_size_bytes = max_file_size_gibibyte << 30
+    shards = []
+    shard, shard_size = {}, 0
+    for k, v in weights.items():
+        # TODO: simplify to v.nbytes as soon as mx.array exposes it
+        estimated_size = v.size * v.dtype.size if isinstance(v, mx.array) else v.nbytes
+        if shard_size + estimated_size > max_file_size_bytes:
+            shards.append(shard)
+            shard, shard_size = {}, 0
+        shard[k] = v
+        shard_size += estimated_size
+    shards.append(shard)
+    return shards
+
+
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Convert Llama weights to MLX")
     parser.add_argument(
@@ -200,6 +215,11 @@ if __name__ == "__main__":
         str(torch_path / "tokenizer.model"),
         str(mlx_path / "tokenizer.model"),
     )
-    np.savez(str(mlx_path / "weights.npz"), **weights)
+    shards = make_shards(weights)
+    if len(shards) == 1:
+        np.savez(str(mlx_path / f"weights.npz"), **shards[0])
+    else:
+        for i, shard in enumerate(shards):
+            np.savez(str(mlx_path / f"weights.{i:02d}.npz"), **shard)
     with open(mlx_path / "config.json", "w") as fid:
         json.dump(params, fid, indent=4)
--- a/llms/llama/llama.py
+++ b/llms/llama/llama.py
@@ -3,6 +3,7 @@
 import argparse
 import json
 import time
+import glob
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Optional, Tuple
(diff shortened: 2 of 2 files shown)


awni:
I love this change!! Could you rebase on main and then I will review?

dastrobu:
> Could you rebase on main and then I will review?

Sure, after taking a brief look at the changes, I guess the new quantization makes this more complicated. It expects the entire model to be loaded. 

I could think of moving the the quantization into its own script, so you would run 

1. convert.py to convert the model 
2. quantize.py, if quantization is wanted (would operate on the converted model)
3. llama.py for inference 

If we really want to do quantize in convert.py it would require the unsharding there I guess. I don't know enough about the new `quantize_module` to know if we could run it on the shards somehow without merging them first. 

Given that quantization should enable smaller machines, it would be nice if we could do the quantization without merging all the unquantized weights in memory first. Not sure though, if there is a clever way to achieve this.

dastrobu:
@awni I think I found a good way to refactor it to support quantize in convert.py. It will still unshard for quantization, but keeps shard loading and conversion lazy and memory friendly. Looking forward to your review.

awni:
@dastrobu I like where this is going, but I suggest we reorganize the computation to avoid the need to unshard in the final loading script. Here's my suggestion:

1. Do the unsharding as before to get the full weights
2. Quantize
3. Split the weights (e.g. different arrays) into smaller files but don't split the arrays themselves
4. Change the `llama.py` to load from a few saved weight files (like you have done), but the loading just needs to read a few files but does not need to deal with concatenating etc which will be faster / simpler / easier to maintain.

Does that make sense?

awni:
So your changes to `llama.py` will look more like how we load [mixtral](https://github.com/ml-explore/mlx-examples/blob/main/llms/mixtral/mixtral.py#L254-L257), except with the option to load from a single file if there is only one as you have it now.

dastrobu:
> Does that make sense?

@awni yes, it does. Thanks for your review and suggestions. 
With quantize in place, I agree that it got a little complicated, and we are probably better off defining our own shards. Something that I initially tried to avoid, as it adds another algorithm to maintain.
Please take a look at the updated code, especially at `make_shards`, which basically the feature of this PR.

llms/llama/llama.py
```diff
@@ -330,7 +331,24 @@ def sanitize_config(config, weights):
 
 def load_model(model_path):
     model_path = Path(model_path)
-    weights = mx.load(str(model_path / "weights.npz"))
+
+    unsharded_weights_path = Path(model_path / "weights.npz")
+    if unsharded_weights_path.is_file():
+        print("[INFO] Loading model from {}.".format(unsharded_weights_path))
+        weights = mx.load(str(model_path / "weights.npz"))
```
awni, line 338: nit: use `str(unsharded_weights_path)` here

llms/llama/llama.py
```diff
@@ -330,7 +331,24 @@ def sanitize_config(config, weights):
...
+    else:
+        sharded_weights_glob = str(model_path / "weights.*.npz")
+        weight_files = glob.glob(sharded_weights_glob)
+        print("[INFO] Loading model from {}.".format(sharded_weights_glob))
+
+        weight_files = sorted(weight_files, key=lambda tf: int(tf.split(".")[-2]))
```
awni, line 344: I don't think it's necessary to load the files in sorted order since the weights are just going into the dictionary, this line can be removed?

llms/llama/convert.py
```diff
@@ -140,6 +139,21 @@ def quantize(weights, config, args):
     return quantized_weights, quantized_config
 
 
+def make_shards(weights: dict, max_file_size_GiB: int = 15):
```
awni, line 142: style nit: `max_file_size_gb`

llms/llama/convert.py
```diff
@@ -140,6 +139,21 @@ def quantize(weights, config, args):
...
+    max_file_size_bytes = max_file_size_GiB << 30
+    shards = []
+    shard, shard_size = {}, 0
+    for k, v in weights.items():
+        estimated_size = len(v.flatten()) * v.dtype.itemsize
```
awni, line 147: Did you check this with quantization? I think this line might break as `dtype` doesn't have an `itemsize`?

We really ought to expose `nbytes` in python.. for consistency with numpy. For now you can do:

```
v.size * v.dtype.size if isintance(v, mx.array) else v.nbytes
```

awni (changes requested):
Looks great and much simpler, thanks for adding this!! I left a couple of comments, please address then we can merge.

llms/llama/convert.py
dastrobu: I wasn't aware that quantization stores mx arrays already... 
your suggestion seems to be a good intermediate solution. Exposing `nbytes` sounds even better, I'll create a PR, sounds like a small change.

llms/llama/convert.py
dastrobu: as we are using 2**30 = GiB I'd suggest to use: `max_file_size_gibibyte` as I find `max_file_size_gb` wrong and `max_file_size_gib` unreadable.

dastrobu:
> Looks great and much simpler, thanks for adding this!! I left a couple of comments, please address then we can merge.

Thanks, should be all fixed now.

awni (approved):
Awesome, thanks!!

PR #176, opened 2023-12-22
Fix conversion + inference errors. - Mistral

Vaibhavs10:
I faced an error with the config whilst converting the quantised instruct and base models.
AFAICT `rope_theta` is not used anywhere else.


awni:
Hmm that's a bug actually, we need to pass `rope_theta` to the RoPE class as in the llama examples. Thanks for catching it, I can update that in this PR.

awni (approved):

Issue #179, opened 2023-12-23
Keep `dtype` of Models


dastrobu:
While playing around with the TinyLlama and Llama examples I noticed that weights are always cast to float16, regardless of the format weights are loaded:

* TinyLlama weights are stored in `float32`
* Llama weights are stored in `bfloat16`

I'd suggest to keep those dtypes in convert.py. 

For `float32` this should be straight forward. Handling `bfloat16` is a bit more challenging. 

1. np does not support `bfloat16`. Since `np.savez` is used, that would need to be changed. 
   Fortunately, mx has `savez` implemented and it works for `bfloat16`, so I'd suggest to use that one instead. 
2. converting from torch is currently done with `v = v.to(torch.float16).numpy()`. To convert to mx without precision loss, we would need to change this to `v = mx.array(v.to(torch.float32).numpy())` to avoid any intermediate rounding. (Maybe there is a better method without np as intermediate data format.)

So I am proposing to move from `np.savez` to `mx.savez` in all examples, where applicable and try to keep the original `dtype` of the models unless explicit conversion like in quantization is requested. 

I am happy to create a PR if there is positive feedback to the proposal.

awni:
> So I am proposing to move from np.savez to mx.savez in all examples, where applicable and try to keep the original dtype of the models unless explicit conversion like in quantization is requested.

I think that is a good idea, let's us save bfloat which I like

> To convert to mx without precision loss, we would need to change this to v = mx.array(v.to(torch.float32).numpy()) to avoid any intermediate rounding. (Maybe there is a better method without np as intermediate data format.)

My only concern here is we have to be careful not to blow up memory for larger models. So we could do something like?

```
weights = {}
for k, v in state_dict.items():
    weights[k] = mx.array(v.to(torch.float32).numpy())
    del v
```

awni:
Are you interested in making this change @dastrobu ?  (Our llm example surface area is getting large, so we could also wait on this until we do a bit of consolidation... ).

vgoklani:
@dastrobu Not following; it looks like you are converting the bfloat16 weights to float32, and saving them as float32. At what point do you convert them back to bfloat16? 

@awni is there a direct way of preserving the bfloat16?

FWIW, my generation quality is pretty poor compared to the reference implementation of Mistral Instruct v0.2, which is loaded and run in bfloat16. The answers from the reference are perfect, I'm not seeing the same quality from MLX.

awni:
[...]

PR #186, opened 2023-12-24
keep dtype on model conversion

dastrobu:
draft implementation for #179

### DIFF
--- a/llms/llama/convert.py
+++ b/llms/llama/convert.py
@@ -10,13 +10,16 @@ from pathlib import Path
 
 import mlx.core as mx
 import mlx.nn as nn
-import numpy as np
 import torch
 from llama import Llama, ModelArgs, sanitize_config
 from mlx.utils import tree_flatten, tree_map, tree_unflatten
 
+def torch_to_mx(a: torch.Tensor, *, dtype: str) -> mx.array:
+    # bfloat16 is not numpy convertible. Upcast to float32 to avoid precision loss
+    a = a.to(torch.float32) if dtype == 'bfloat16' else a.to(getattr(torch, dtype))
+    return mx.array(a.numpy(), getattr(mx, dtype))
 
-def llama(model_path):
+def llama(model_path, *, dtype: str):
     SHARD_FIRST = ["wv", "wq", "wk", "w1", "w3", "output"]
     SHARD_SECOND = ["tok_embeddings", "wo", "w2"]
     SHARD_WEIGHTS = set(SHARD_FIRST + SHARD_SECOND)
@@ -37,14 +40,15 @@ def llama(model_path):
             axis = 1
         else:
             raise ValueError("Invalid weight name")
-        return np.concatenate(v, axis=axis)
+        return mx.concatenate(v, axis=axis)
 
     torch_files = glob.glob(str(model_path / "consolidated.*.pth"))
     weights = collections.defaultdict(list)
     for wf in torch_files:
         state = torch.load(wf, map_location=torch.device("cpu"))
         for k, v in state.items():
-            v = v.to(torch.float16).numpy()
+            v = torch_to_mx(v, dtype=dtype)
+            state[k] = None # free memory
             if shard_key(k) in SHARD_WEIGHTS:
                 weights[k].append(v)
             else:
@@ -57,7 +61,7 @@ def llama(model_path):
     return weights, params
 
 
-def tiny_llama(model_path):
+def tiny_llama(model_path, *, dtype: str):
     try:
         import transformers
     except ImportError:
@@ -113,7 +117,7 @@ def tiny_llama(model_path):
     params["vocab_size"] = config.vocab_size
     params["norm_eps"] = config.rms_norm_eps
     params["rope_traditional"] = False
-    weights = {k: v.to(torch.float16).numpy() for k, v in model.items()}
+    weights = {k: torch_to_mx(v, dtype=dtype) for k, v in model.items()}
 
     return weights, params
 
@@ -197,6 +201,13 @@ if __name__ == "__main__":
         type=int,
         default=4,
     )
+    parser.add_argument(
+        "--dtype",
+        help="dtype for loading the torch model and input for quantization or saving the converted model. "
+             "The original weights are stored in bfloat16.",
+        type=str,
+        default="float16",
+    )
 
     args = parser.parse_args()
 
@@ -205,7 +216,7 @@ if __name__ == "__main__":
     mlx_path.mkdir(parents=True, exist_ok=True)
 
     print("[INFO] Loading")
-    weights, params = globals()[args.model_name](torch_path)
+    weights, params = globals()[args.model_name](torch_path, dtype=args.dtype)
     params["model_type"] = "llama"
     if args.quantize:
         print("[INFO] Quantizing")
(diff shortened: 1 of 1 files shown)


llms/llama/convert.py
```diff
@@ -37,14 +36,15 @@ def unshard(k, v):
...
-        return np.concatenate(v, axis=axis)
+        return mx.concatenate(v, axis=axis)
 
     torch_files = glob.glob(str(model_path / "consolidated.*.pth"))
     weights = collections.defaultdict(list)
     for wf in torch_files:
         state = torch.load(wf, map_location=torch.device("cpu"))
         for k, v in state.items():
-            v = v.to(torch.float16).numpy()
+            v = mx.array(v.to(torch.float32).numpy(), mx.bfloat16)
```
awni, line 46: The only thing I'm wondering about this change is that it will default the model inference to `mx.bfloat16` which might be a lot slower than `mx.float16`. If that's the case (I haven't tested it yet) I might suggest that we give the option of choosing the output type.

So it could be an argument `--type` which is any of `mx.float16`, `mx.bfloat` and `mx.float32`? Wdyt?

dastrobu: I'll measure performance comparing float16 and bfloat16. I would probably add the --type parameter to the inference script if needed at all. But lets not optimizer premature and measure performance first.

dastrobu: Found the time to do some measurements on inference and it looks quite encouraging. 

I run basically: 
```
    for dtype in [mx.float16, mx.bfloat16, mx.float32]:
        model.update(tree_map(lambda x: x.astype(dtype), model.parameters()))
        repeat = 3
        print("{}: best out of {}: {:.3f}s".format(dtype, repeat, min(timeit.repeat("generate(args)", number=1, repeat=repeat, globals=globals()))), file=sys.stderr)
```
for the `llama-2-13b-chat` with default parameters for generation. 

Results from my M3 Max Mac:
```
mlx.core.float16: best out of 3: 8.660s
mlx.core.bfloat16: best out of 3: 8.520s
mlx.core.float32: best out of 3: 192.604s
```

bfloat16 performs actually a tiny bit faster than float16, but I think that's not statistically significant. For comparison, float32 is roughly 200% slower. 

So from my point of view there is no need for a `--type` parameter.

awni: Wow 200% slower for fp32. That's bad.. it must not fit well on your machine.  I'm cool with defaulting this to `bfloat16`. Do you want to submit the PR for review?

dastrobu: That is probably true, 200% percent is unexpected. Probably the model is too big in float32, but I did not look into it. I did some general matrix multiplication timing, where I see around 3% to 5% slowdown for float32, depending on the matrix size. 
Sure, we can review/merge this PR and create follow-up PRs for other models.

Willian-Zhang: forgive me if I'm wrong, but isn't `192.604` more like 24x to `8.520` than 200%?

Willian-Zhang: just from the number alone I thought I would've been difference between ANE or GPU, until I saw [this](https://github.com/ml-explore/mlx/issues/18)

dastrobu: > forgive me if I'm wrong, but isn't `192.604` more like 24x to `8.520` than 200%?

@Willian-Zhang I don't remember what my computation was, but you are probably right. We may just agree that it's just a big difference. 

As awni already pointed out, this is unexpected, and I do not see this kind of performance degradation for float32 in general. If you experience performance issues with float32, it might be worth opening a separate issue for that topic. 
I understood that there are still some [performance improvements underway](https://github.com/ml-explore/mlx/issues/243#issuecomment-1868376661), so I will hold back with further performance benchmarks until the next version of mlx.

Willian-Zhang: I'm getting some different numbers on an M1 mac
```
f16: 350 token/s
bf16: 100 token/s
f32: 137 token/s
```

Willian-Zhang: It's most likely not memory constrained from Page-In out status, however I'm not sure how I would be able to tell exactly.

Willian-Zhang: it seems bf16 perform less to f32 in this case, and AFAIK M1 does not support bf16 natively. I'm not sure if its stored f32 internally and somehow converting bf16 f32 back and forth or entirely simulated with f32 under the hood.

awni: Those timings are surprising! I would not expect (even with native support) for FP16 to be more than 2x faster than fp32.

Can you say more about what you ran in that benchmark?
Which LLM,  your OS, etc.

Willian-Zhang: mistral 7B, 
14.2.1 (23C71), 
m1 ultra, 20c, 64c, 32c, 128G

awni: What did you use to benchmark? I just ran mistral 7B in FP16 on an m2ultra with 192 GB and I see more like 30 tokens-per-secon.

Willian-Zhang: a modified version of the lora.py, with some custom data varying around 2048 token per datapoint.

Willian-Zhang: @awni 
image
may I ask which one did you use?

Willian-Zhang: > bs=1
I felt padding on mlx currently without careful thought on parallelism would result in performance impact, so larger context Len, lowered bs to 1

awni: The one with 76-core GPU.

Willian-Zhang: > The one with 76-core GPU.

well that 's surprising

awni: > I felt padding on mlx currently without careful thought on parallelism would result in performance impact, so larger context Len, lowered bs to 1

Fair enough. The padding + batching strategy is very naive in that example. It should try to put similar length samples in the same batch.

awni: > well that 's surprising

No I was timing generation (which is very different from training). 300 tps for training is sensible, even on the slow side.

Willian-Zhang: > No I was timing generation (which is very different from training).

Oh. okey now it make sense.

awni:
@dastrobu I timed this on an M2Ultra OS 14.0 and it's noticeably slower to use `bfloat16` than `float16`.

float16 `Tokens per second: prompt 20.737, generation 29.262`
bfloat16 `Tokens per second: prompt 12.071, generation 18.866`

I think this could be due to a couple of things: 1. There is more native support for float16 in Metal in OS 14. 2. If memory bandwidth is really good, then we pay a bigger cost for the non-native conversion from bfloat16 to float32 when the actual operation is executed.

Are you using OS 14+?

Since it seems to matter, would you be up for keeping the default as float16 in the scripts but giving the option to write to bfloat16 for maybe slightly better accuracy?

dastrobu:
> Are you using OS 14+?

Version 14.1.2 (23B2091)
Maybe M3 Max also has some impact, but I don't know any details. 

> Since it seems to matter, would you be up for keeping the default as float16 in the scripts but giving the option to write to bfloat16 for maybe slightly better accuracy?

@awni you mean adding a dtype parameter to the convert script which controls `v.to(torch.float16).numpy()`?

awni:
> you mean adding a dtype parameter to the convert script which controls v.to(torch.float16).numpy()?

Yea exactly. I would default it to float16 which I think is a good trade-off between accuracy / speed. But it can be bf16 (and quantization will override whatever is specified there).

dastrobu:
> Yea exactly. I would default it to float16 which I think is a good trade-off between accuracy / speed. But it can be bf16 (and quantization will override whatever is specified there).

@awni done

awni (approved):
Thank you 🚀

PR #190, opened 2023-12-26
Fixed the return type for the __call__ method in Attention

sushantkumar23:
Corrected the return signature of __call__ method in Attention Module in llama.py


awni (approved):
👍  thanks!!

PR #192, opened 2023-12-26
fix: Add numpy to CIFAR's requirements.txt

jirimoravcik:
I noticed that CIFAR example is missing `numpy` in requirements.txt

### DIFF
--- a/cifar/requirements.txt
+++ b/cifar/requirements.txt
@@ -1,2 +1,3 @@
 mlx
 mlx-data
+numpy


awni:
I don't think it uses `numpy`? DId you try to run it without numpy and it failed?

awni:
(I will reopen this if I'm mistaken, lmk)

jirimoravcik:
> (I will reopen this if I'm mistaken, lmk)
Sry, I forgot to provide repro steps, so here you go:

If I have following installed (a clean conda environment, then I do `pip install -r requirements.txt`):
```
mlx==0.0.6
mlx-data==0.0.1
```
then I get:
```
  File "/Users/jirimoravcik/dev/mlx-examples/cifar/main.py", line 8, in <module>
    from dataset import get_cifar10
  File "/Users/jirimoravcik/dev/mlx-examples/cifar/dataset.py", line 4, in <module>
    from mlx.data.datasets import load_cifar10
  File "/opt/homebrew/Caskroom/miniconda/base/envs/mlx/lib/python3.10/site-packages/mlx/data/datasets/__init__.py", line 3, in <module>
    from .mnist import load_mnist
  File "/opt/homebrew/Caskroom/miniconda/base/envs/mlx/lib/python3.10/site-packages/mlx/data/datasets/mnist.py", line 8, in <module>
    import numpy as np
```
which refers to https://github.com/ml-explore/mlx-data/blob/main/python/mlx/data/datasets/mnist.py#L8

Doing `pip install numpy` fixes it.

awni:
Got it, hidden in the depths of mlx-data :). That should really be a mlx-data dep. But we can add it here for now to avoid people bumping into this again.

awni (approved):
Thanks for checking this and fixing it!

PR #193, opened 2023-12-27
[Whisper] Large-v3 requires 128 Mel frequency bins

dimopep:
The script benchmark.py encounters an error when executed with the "large" model. This issue arises because the "large" model references "large-v3", which necessitates 128 Mel frequency bins. In contrast, all previous models were compatible with only 80 Mel frequency bins. Calling the transcribe() function performs inferencing correctly.

### DIFF
--- a/whisper/benchmark.py
+++ b/whisper/benchmark.py
@@ -23,10 +23,10 @@ def timer(fn, *args):
     return (toc - tic) / num_its
 
 
-def feats():
+def feats(n_mels: int = 80):
     data = audio.load_audio(audio_file)
     data = audio.pad_or_trim(data)
-    mels = audio.log_mel_spectrogram(data)
+    mels = audio.log_mel_spectrogram(data, n_mels)
     mx.eval(mels)
     return mels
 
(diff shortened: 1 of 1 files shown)


whisper/benchmark.py
```diff
@@ -46,19 +46,26 @@ def everything():
...
-    mels = feats()[None].astype(mx.float16)
-
     for model_name in models:
+        # as long large "points" to "large-v3"
+        if model_name == "large" or model_name == "large-v3":
+            n_mels = 128
+        else:
+            n_mels = 80
+
+        mels = feats(n_mels)[None].astype(mx.float16)
```
awni, line 68: How about querying the `model.dims.n_mels` here to avoid running into this problem in the future? You can construct the features after we load the model below.

whisper/benchmark.py
```diff
@@ -46,19 +46,26 @@ def everything():
...
     models = ["tiny"]
-    if len(sys.argv) > 1:
-        if sys.argv[1] == "--all":
-            models = ["tiny", "small", "medium", "large"]
+    for i, arg in enumerate(sys.argv):
+        if arg == "-all":
+            models = ["tiny", "small", "medium", "large-v1", "large-v3"]
+            break
+        elif arg in ("-m", "--models") and i + 1 < len(sys.argv):
+            models = sys.argv[i + 1].split(",")
```
awni, line 57: Can we use argparse here? It seems like it would simplify this a bit.

awni (changes requested):
Thanks for catching this! I suggested a couple changes, please address and then we can merge!

whisper/benchmark.py
dimopep: Done 👍

whisper/benchmark.py
dimopep: Done 👍

awni (approved):

PR #195, opened 2023-12-27
fix load model in lora/lora.py

pjh6818:
fixed an error when loading model in lora by modifying config dict to fit in ModelArgs.

### DIFF
--- a/lora/lora.py
+++ b/lora/lora.py
@@ -334,16 +334,26 @@ def generate(model, prompt, tokenizer, args):
 
 def load_model(folder: str, dtype=mx.float16):
     model_path = Path(folder)
+    weights = mx.load(str(model_path / "weights.npz"))
     tokenizer = Tokenizer(str(model_path / "tokenizer.model"))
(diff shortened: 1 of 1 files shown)


lora/lora.py
```diff
@@ -334,16 +334,26 @@ def sample(logits):
...
         if config.get("vocab_size", -1) < 0:
-            config["vocab_size"] = tokenizer.vocab_size
-        model_args = ModelArgs(**config)
-    weights = mx.load(str(model_path / "weights.npz"))
+            config["vocab_size"] = weights["output.weight"].shape[-1]
+        unused = ["multiple_of", "ffn_dim_multiplier"]
+        for k in unused:
+            if k in config:
+                config.pop(k)
+    model = Model(ModelArgs(**config))
```
awni, line 354: Should these be in the `convert.py` for consistency? In this example we remap the models to a common format ahead of loading them

pjh6818: I didn't catch that `lora/convert.py` changes params.json file, because l looked llama example then looked at lora.

PR #198, opened 2023-12-27
Add an image2image example in the stable diffusion

angeloskath:
This PR started in order to address issue 286 on the main repo and anybody else having trouble with the autoencoder. The changes are as follows:

- Fixes a bug in the downsampling where the padding is applied asymmetrically
- Refactors the auto encoder to include its scaling factor as it should be
- Refactors the `StableDiffusion` class to be slightly more reusable
- Adds the ability to add noise to some latents in the sampler and to start from a given step instead of the max step

In my tests, using strength ~0.9 works well for getting variations of an image but of course ymmv depending on model and noise schedule.


stable_diffusion/image2image.py
```diff
@@ -0,0 +1,65 @@
...
+        args.prompt,
+        strength=args.strength,
+        n_images=args.n_images,
+        cfg_weight=args.cfg,
+        num_steps=args.steps,
+        negative_text=args.negative_prompt,
+    )
+    for x_t in tqdm(latents, total=int(args.steps * args.strength)):
+        mx.simplify(x_t)
+        mx.simplify(x_t)
```
awni, line 46: Why the double simplify?

stable_diffusion/image2image.py
```diff
@@ -0,0 +1,65 @@
...
+import mlx.core as mx
+import numpy as np
+from PIL import Image
+from tqdm import tqdm
+
+from stable_diffusion import StableDiffusion
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Generate images from a textual prompt using stable diffusion"
```
awni, line 14: Nit: Update description.

awni (commented):
🚀  thanks for adding that!

Probably worth adding a short sub-section in the readme on this as well so people know about it!

awni (approved):
Meant to approve 🚀

stable_diffusion/image2image.py
angeloskath: It gives a tiny bit more performance :-) . Simplify doesn't get everything in one pass :-/ ...

PR #200, opened 2023-12-28
Fix benchmark

bofenghuang:
Pass the model name to `transcribe` function, otherwise, it always uses the default tiny one


awni (approved):
Oops, thanks for fixing that!

PR #203, opened 2023-12-29
chore: fix the load quantization model for deepseek coder

mzbac:
While I was trying to upload the quantized deepseek code model to mlx-community, noticed the deepseek code example wouldn't load the quantized model because the quantization config popped up during loading of model arguments. The following "config.get("quantization", False)" will always return false, even for the quantized model.


awni (approved):
Good catch, thanks!!

Issue #204, opened 2023-12-30
Has anyone figured out how to run the Yi series models with mlx?


miguelbar3000:
I've found that the Yi series models from 01.ai work very well when I run them in text-generation-webui, but they become very slow as the context size increases. I converted the Yi-6B-Chat model weights (https://huggingface.co/01-ai/Yi-6B-Chat) from safetensors format to npz, but when I tried running it using the llama.py script from mlx-examples it generated gibberish. Has anyone figured out how to run it with mlx and make it work properly?

awni:
Did you use one of the scripts in `mlx-examples` to convert it? If so which script?

mzbac:
Had a quick look at the Yi model. It seems that if you use tiny_llama to convert the model, it will miss the `rope_theta` in config. Additionally, the Yi model has different special tokens used in the tokenizer, so the standard sentencepiece tokenizer seems to use the wrong pad token or bos token. I managed to run it via mlx, however, I had to use a modified convert script from deepseek example and use deekseep example to run it.

here is my updated convert.py
```
import argparse
import copy
import json
from pathlib import Path

import mlx.core as mx
import mlx.nn as nn
import numpy as np
import torch
from deepseek_coder import DeepseekCoder, ModelArgs
from mlx.utils import tree_flatten, tree_map, tree_unflatten
from transformers import AutoModelForCausalLM, AutoTokenizer


def quantize(weights, config, args):
    quantized_config = copy.deepcopy(config)

    # Load the model:
    model_args = ModelArgs(**config)
    model = DeepseekCoder(model_args)

    weights = tree_map(mx.array, weights)
    model.update(tree_unflatten(list(weights.items())))

    # Quantize the model:
    nn.QuantizedLinear.quantize_module(model, args.q_group_size, args.q_bits)

    # Update the config:
    quantized_config["quantization"] = {
        "group_size": args.q_group_size,
        "bits": args.q_bits,
    }
    quantized_weights = dict(tree_flatten(model.parameters()))

    return quantized_weights, quantized_config


def convert(args):
    hf_path = Path(args.hf_path)

    model = AutoModelForCausalLM.from_pretrained(
        str(hf_path), trust_remote_code=True, torch_dtype=torch.float16
    )
    config = model.config.to_dict()

    state_dict = model.state_dict()
    tokenizer = AutoTokenizer.from_pretrained(str(hf_path), trust_remote_code=True, use_fast=False)

    # things to change
[...]

PR #205, opened 2023-12-30
chore: make the Deepseek example compatible with Yi models.

mzbac:
Before consolidating all the models similar to llamas into the llama example, just make some minor changes in the DeepSeek example so that people can use it to run HF format models with autoTokenizer if they want to try out some hf format models. for example https://github.com/ml-explore/mlx-examples/issues/204

### DIFF
--- a/llms/deepseek-coder/deepseek_coder.py
+++ b/llms/deepseek-coder/deepseek_coder.py
@@ -248,7 +248,7 @@ def load_model(model_path: str):
         nn.QuantizedLinear.quantize_module(model, **quantization)
     model.update(tree_unflatten(list(weights.items())))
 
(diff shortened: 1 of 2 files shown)


llms/deepseek-coder/deepseek_coder.py
```diff
@@ -248,7 +248,7 @@ def load_model(model_path: str):
         nn.QuantizedLinear.quantize_module(model, **quantization)
     model.update(tree_unflatten(list(weights.items())))
 
-    tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True)
+    tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True, use_fast=False)
```
awni, line 251: 👍

awni (approved):
Nice, simple change!!

PR #208, opened 2023-12-31
feat: add yayi2-30b example

mzbac:
Yayi2-30b has a score of over 80 MMLU. I performed some fine-tuning on it using qlora, and from my quick test, it appears to be very promising. so, I created the mlx example for this model in case anyone wants to run it via mlx. However, this model has an unusual k,v layer that causes the quantization to fail. Currently, I haven't found any quantization tools that support this model (except for bnb nf4). It would be great if mlx could provide support for its quantization.

FYI:
https://huggingface.co/wenge-research/yayi2-30b
https://huggingface.co/mzbac/yayi2-30b-guanaco
https://github.com/ml-explore/mlx/issues/328

### DIFF
--- a/llms/yayi2/README.md
+++ b/llms/yayi2/README.md
@@ -1,0 +1,42 @@
+# YAYI2
+
+YAYI 2 is a collection of open-source large language models launched by Wenge Technology. YAYI2-30B is a Transformer-based large language model, and has been pretrained for 2.65 trillion tokens of multilingual data with high quality. The base model is aligned with human values through supervised fine-tuning with millions of instructions and reinforcement learning from human feedback (RLHF).
+
+
+### Setup
+
+Install the dependencies:
+
+```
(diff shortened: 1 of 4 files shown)


mzbac:
I have added the workaround for the quant as suggested in https://github.com/ml-explore/mlx/issues/328. Now, the example works with 4-bit quantization. Once the PR gets merged, I will upload the 4-bit quantized model.

awni:
HI @mzbac sorry for the delayed review here. Do you still want to merge this? I think given the non-standard size it wouldn't fit easily in our `hf_llm` example, but wdyt?

mzbac:
I think this should be supported by hf_llm once we fix the quant non-32 dimension. I'm happy to close this one. Meanwhile, if people want to try this model in f16 precision, they should be able to run it via hf_llm.

awni:
Sounds good, thank you!

PR #215, opened 2024-01-02
Support Hugging Face models

awni:
Example supports

- Loading mistral/llama-style models directly from the Hugging Face Hub
- Converting models to different datatypes or quantized. Conversion script can upload and download models from the hub directly

### DIFF
--- a/llms/hf_llm/README.md
+++ b/llms/hf_llm/README.md
@@ -58,7 +58,7 @@ You can convert (change the data type or quantize) models using the
 `convert.py` script. This script takes a Hugging Face repo as input and outputs
 a model directory (which you can optionally also upload to Hugging Face).
 
-For example to 4-bit quantize a model, run:
+For example, to make 4-bit quantized a model, run:
 
 ```
 python convert.py --hf-model <hf_repo> -q
--- a/llms/hf_llm/README.md
+++ b/llms/hf_llm/README.md
@@ -26,11 +26,17 @@ Run `python generate.py --help` to see all the options.
 
 ### Models
 
-The following models (and variants) are supported:
+The example supports Hugging Face format Llama-style models.  If the
+model you want to convert is not supported, file an
+[issue](https://github.com/ml-explore/mlx-examples/issues/new) or better yet,
+submit a pull request.
 
-Hugging Face Repo | Model Size |
------------------ | ---------- |
-N/A | N/A 
+Here is a list of a few Hugging Face models which work with this example:
+
+- meta-llama/Llama-2-7b-hf
+- mistralai/Mistral-7B-v0.1
+- TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T
+
 
 ### Convert new models 
 
@@ -50,14 +56,8 @@ To make a 4-bit quantized model, use `-q`. For more options run:
 python convert.py
 ```
 
-The conversion script supports Hugging Face format Llama-style models.  If the
-model you want to convert is not supported, file an
-[issue](https://github.com/ml-explore/mlx-examples/issues/new) or better yet,
-submit a pull request.
+You can upload new models to the [Hugging Face MLX
+Community](https://huggingface.co/mlx-community) by specifying
+`--upload-name`` to `convert.py`.
 
-This is a list of Hugging Face models which have been tested with the
-conversion script:
 
-- meta-llama/Llama-2-7b-hf
-- mistralai/Mistral-7B-v0.1
-- TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T
--- a/llms/hf_llm/README.md
+++ b/llms/hf_llm/README.md
@@ -1,4 +1,4 @@
-## Generate Text with MLX and Hugging Face
+## Generate Text with MLX and :hugs: Hugging Face
 
 This an example of Llama style large language model text generation that can
 pull models from the Hugging Face Hub.
--- a/llms/hf_llm/README.md
+++ b/llms/hf_llm/README.md
@@ -27,7 +27,7 @@ will download the Mistral 7B model and generate text using the given prompt.
 
 The `<model_path>` should be either a path to a local directory or a Hugging
 Face repo with weights stored in `safetensors` format. If you use a repo from
-the Hugging Face hub, then the model will be downloaded and cached the first
+the Hugging Face Hub, then the model will be downloaded and cached the first
 time you run it. See the [Models](#models) section for a full list of supported models.
 
 Run `python generate.py --help` to see all the options.
@@ -40,7 +40,7 @@ model you want to run is not supported, file an
 [issue](https://github.com/ml-explore/mlx-examples/issues/new) or better yet,
 submit a pull request.
 
-Here is a list of a few more Hugging Face models repos which work with this example:
+Here are a few examples of Hugging Face models which work with this example:
 
 - [mistralai/Mistral-7B-v0.1](https://huggingface.co/mistralai/Mistral-7B-v0.1)
 - [meta-llama/Llama-2-7b-hf](https://huggingface.co/meta-llama/Llama-2-7b-hf)
@@ -54,24 +54,22 @@ style models should work out of the box.
 
 ### Convert new models 
 
-You can convert new models to the MLX format using the `convert.py` script.
-This script takes a Hugging Face repo as input and outputs an MLX formatted
-model (which you can then upload to Hugging Face).
+You can convert (change the data type or quantize) models using the
+`convert.py` script. This script takes a Hugging Face repo as input and outputs
+a model directory (which you can optionally also upload to Hugging Face).
 
-To convert a model, run:
-
-```
-python convert.py --hf-model <hf_repo>
-```
-
-To make a 4-bit quantized model, use `-q`. For more options run:
+For example to 4-bit quantize a model, run:
 
 ```
-python convert.py
+python convert.py --hf-model <hf_repo> -q
+```
+
+For more options run:
+
+```
+python convert.py --help
 ```
 
 You can upload new models to the [Hugging Face MLX
-Community](https://huggingface.co/mlx-community) by specifying
-`--upload-name`` to `convert.py`.
-
-
+Community](https://huggingface.co/mlx-community) by specifying `--upload-name``
+to `convert.py`.
--- a/llms/hf_llm/README.md
+++ b/llms/hf_llm/README.md
@@ -1,4 +1,4 @@
-## Generate Text in MLX
+## Generate Text with MLX and Hugging Face
 
 This an example of Llama style large language model text generation that can
 pull models from the Hugging Face Hub.
@@ -17,6 +17,14 @@ pip install -r requirements.txt
 python generate.py --model <model_path> --prompt "hello"
 ```
 
+For example:
+
+```
+python generate.py --model mistralai/Mistral-7B-v0.1 --prompt "hello"
+```
+
+will download the Mistral 7B model and generate text using the given prompt.
+
 The `<model_path>` should be either a path to a local directory or a Hugging
 Face repo with weights stored in `safetensors` format. If you use a repo from
 the Hugging Face hub, then the model will be downloaded and cached the first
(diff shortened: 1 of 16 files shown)


awni:
CC @Vaibhavs10

mzbac:
While I am waiting for support on the HF format models, I have created a separate repository (https://github.com/mzbac/mlx-llm) with a very similar implementation to support HF format models. Thank you all for the examples and help. I would love to contribute back mlx-example. However, I have a few questions:

1. What is the reason behind wanting to use different configurations instead of using the same configuration transformers that were already used? It would be more beneficial if we could keep the configuration consistent so that later we can integrate MLX as a transformer backend without having to migrate the configuration again.

2. Also, what is the reason for wanting to continue using all parameter layer names to match PyTorch model's naming instead of keeping it consistent with transformer's naming convention? If we can maintain consistency in naming conventions, it will be beneficial for us when expanding some features later on. For example, we could provide a script to convert MLX fine-tuned model weights back into HF format models without having re-mapping parameters again.

awni:
I see your points.. maybe we should simply update our model to be the same as the transformer config. I'm not opposed to it... mostly just inertia is the reason we are different. 

The initial goal was just to make a simple / easy to read example. But if we want it to work seamlessly in and out of HF your points are well taken.

awni:
I will play around with that here to see if it simplifies things. I don't think we can completely avoid in general needing to pre-converters (since for different types is can be very expensive to load and quantize the model on the fly. But at least we could avoid the remapping for many instances.

mzbac:
Thanks for sharing the information. Based on my experience working on my mlx-llm, I feel that the converter is still necessary. but, at least we can simplify the convert script a bit and make minor changes in rope in the attention layer, we easily support deepseek and Yi models. The fast/legacy tokenizer is still a bit of a pain :(

Vaibhavs10 (commented):
Love it! Thanks for putting this together so fast! 

I agree with @mzbac re: directly using the transformers config and naming convention. This enables a couple of things:
1. Different fine-tunes (Hermes for Mistral) / from scratch trained (Yi/ Deci for llama) have relatively minimal changes in the architecture and directly operating on transformers config + convention enables this.
2. Another reason for operating directly on transformers config is that we enable long context models via RoPE scaling (200K context and so on).

IMO we don't need to make this exhaustive, but at least supporting Mistral/Llama and their fine-tunes should unlock considerable momentum here.

I also agree that a conversion script is still necessary, especially for 4-bit. But if we can make changes such that we don't need converters for fp16 then it should work out for the best.

Edit: To clarify, we should avoid conversion for fp16 and directly use transformers safetensors weights.

llms/hf_llm/convert.py
```diff
@@ -0,0 +1,178 @@
...
+    # Download model, config and tokenizer from HF
+    model = transformers.AutoModelForCausalLM.from_pretrained(
+            hf_path,
+            trust_remote_code=True,
+            torch_dtype=getattr(torch, dtype),
+            ).state_dict()
+    config = transformers.AutoConfig.from_pretrained(hf_path)
+    tokenizer = transformers.AutoTokenizer.from_pretrained(
+            hf_path,
+            trust_remote_code=True,
```
Vaibhavs10, line 30: Hmm. Why do we need this? This could lead to remote code execution, and we should set it to false by default and let the end-user explicitly set True if they need it.

llms/hf_llm/convert.py
```diff
@@ -0,0 +1,178 @@
...
+        type=int,
+        default=64,
+    )
+    parser.add_argument(
+        "--q-bits",
+        help="Bits per weight for quantization.",
+        type=int,
+        default=4,
+    )
+    parser.add_argument(
```
Vaibhavs10, line 155: To make it easier for people to upload quants on the Hub via a flag. It should be just adding couple lines of code similar to this: 

```python

from huggingface_hub import HfApi
from huggingface_hub import logging

logging.set_verbosity_info()

api = HfApi()

api.upload_folder(folder_path="<LOCAL FOLDER PATH>", 
                  repo_id="mlx-community/<MODEL_NAME>", 
                  repo_type="model", 
                  multi_commits=True, 
                  multi_commits_verbose=True)
```

pcuenca (commented):
Looks great after a quick pass; will test it in practice later.

As @Vaibhavs10 and @mzbac said, it'd be awesome if we could avoid the conversion step for unquantized models, as this would greatly reduce friction for users, and we'd immediately get MLX support for hundreds of models: all Llama models could be automatically tagged with `mlx` on the Hub. Given that there's already support for `safetensors`, would the blocker for this be the conversion time, or something else?

If we adopted this strategy, we could just turn `convert` into a quantization script, and I expect we could use similar load functions for both `convert` and `generate`.

I really appreciate the design goal to make self-contained and very clear and self-documented examples. However, creating a specific example that follows the `transformers` naming conventions could get a lot of traction for the reasons explained by the others – this is a `hf_` example folder after all :)

I also agree with trying to make quantized models easily uploadable to the Hub.

Thanks a lot for working on this, it's awesome! 🚀

awni:
Ok everyone I made a few changes:

1. Using the transformer config / model names. No more key mappings which is great 👍 
2. Added upload option to the conversion script using @Vaibhavs10 suggestion
3. The `generate.py` can use pre existing models on the Hub (llama style for now). Now conversion needed 🎊 !

llms/hf_llm/README.md
```diff
@@ -0,0 +1,63 @@
...
+Run `python generate.py --help` to see all the options.
+
+
+### Models
+
+The following models (and variants) are supported:
+
+Hugging Face Repo | Model Size |
+----------------- | ---------- |
+N/A | N/A
```
pcuenca, line 33:
```suggestion
Llama- and Mistral-based models, including fine-tunes, that are available in `safetensors` format should work out of the box. You can use the following filters to browse and find models in the Hub:

- [Llama](https://huggingface.co/models?library=transformers,safetensors&other=llama&sort=trending)
- [Mistral](https://huggingface.co/models?library=transformers,safetensors&other=mistral&sort=trending)
```

llms/hf_llm/README.md
```diff
@@ -0,0 +1,63 @@
...
+python convert.py --hf-model <hf_repo>
+```
+
+To make a 4-bit quantized model, use `-q`. For more options run:
+
+```
+python convert.py
+```
+
+The conversion script supports Hugging Face format Llama-style models.  If the
```
pcuenca, line 53:
```suggestion
The conversion script supports Llama-style models in Hugging Face format, with weights exported as `safetensors`.  If the
```

llms/hf_llm/convert.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def fetch_from_hub(hf_path: str):
+    model_path = snapshot_download(
+        repo_id=hf_path,
+        allow_patterns=["*.json", "*.safetensors", "tokenizer.model"],
+    )
+    weight_files = glob.glob(f"{model_path}/*.safetensors")
+    if len(weight_files) == 0:
+        raise FileNotFoundError("No weights found in {}".format(model_path))
```
pcuenca, line 28:
```suggestion
        raise FileNotFoundError("No safetensors weights found in {}".format(model_path))
```

llms/hf_llm/models.py
```diff
@@ -0,0 +1,255 @@
...
+        )
+
+    with open(model_path / "config.json", "r") as f:
+        config = json.loads(f.read())
+        quantization = config.get("quantization", None)
+        model_args = ModelArgs.from_dict(config)
+
+    weight_files = glob.glob(str(model_path / "*.safetensors"))
+    if len(weight_files) == 0:
+        raise FileNotFoundError("No weights found in {}".format(model_path))
```
pcuenca, line 223:
```suggestion
        raise FileNotFoundError("No safetensors weights found in {}".format(model_path))
```

pcuenca (approved):
Amazing! 🔥

llms/hf_llm/convert.py
```diff
@@ -0,0 +1,136 @@
+# Copyright © 2023 Apple Inc.
+
+import argparse
+import collections
```
pcuenca, line 4:
```suggestion
```

llms/hf_llm/convert.py
```diff
@@ -0,0 +1,136 @@
...
+import copy
+import glob
+import json
+import shutil
```
pcuenca, line 8:
```suggestion
```

llms/hf_llm/convert.py
```diff
@@ -0,0 +1,136 @@
...
+from pathlib import Path
+
+import mlx.core as mx
+import mlx.nn as nn
+import numpy as np
+import torch
```
pcuenca, line 14:
```suggestion
```

llms/hf_llm/convert.py
```diff
@@ -0,0 +1,157 @@
...
+    return shards
+
+
+def upload_to_hub(path: str, name: str):
+    from huggingface_hub import HfApi, logging
+
+    logging.set_verbosity_info()
+
+    api = HfApi()
+
```
pcuenca, line 79:
```suggestion

    repo_id = f"mlx-community/{name}"
    api.create_repo(repo_id=repo_id, exist_ok=True)
```

llms/hf_llm/convert.py
```diff
@@ -0,0 +1,157 @@
...
+    api.upload_folder(
+        folder_path=path,
+        repo_id=f"mlx-community/{name}",
```
pcuenca, line 82:
```suggestion
        repo_id=repo_id,
```

pcuenca (commented):
Creating repo before upload

llms/hf_llm/convert.py
```diff
@@ -0,0 +1,159 @@
...
+
+    repo_id = f"mlx-community/{name}"
+    api.create_repo(repo_id=repo_id, exist_ok=True)
+    api.upload_folder(
+        folder_path=path,
+        repo_id=repo_id,
+        repo_type="model",
+        multi_commits=True,
+        multi_commits_verbose=True,
+    )
```
pcuenca, line 88: Updated snippet to create a model card that refers to the original model, reuses metadata and adds the `mlx` tag:

```suggestion
def upload_to_hub(path: str, name: str, hf_path: str):
    import os
    from huggingface_hub import HfApi, logging, ModelCard

    repo_id = f"mlx-community/{name}"

    card = ModelCard.load(hf_path)
    card.data.tags = ["mlx"] if card.data.tags is None else card.data.tags + ["mlx"]
    card.text = f"""
# {name}

This model was converted to MLX format from [`{hf_path}`]().
Please, refer to the [original model card](https://huggingface.co/{hf_path}) for more details on the original model.

## Use with mlx

```bash
pip install mlx
git clone https://github.com/ml-explore/mlx-examples.git

cd mlx-examples/llms/hf_llm
python generate.py --model {repo_id} --prompt "My name is"
```
"""
    card.save(os.path.join(path, "README.md"))

    logging.set_verbosity_info()

    api = HfApi()
    api.create_repo(repo_id=repo_id, exist_ok=True)
    api.upload_folder(
        folder_path=path,
        repo_id=repo_id,
        repo_type="model",
    )
```

Comment (pcuenca) [inline llms/hf_llm/convert.py:88]: Updated snippet to create a model card that refers to the original model, reuses metadata and adds the `mlx` tag: https://gist.github.com/pcuenca/4f4839325e799183bc6b81c668a24161

(Sorry, can't submit inline as the preview gets confused with the backticks)

**Comment by mzbac** [comment]
Nice work! 🥳

File: llms/hf_llm/convert.py [hunk 13]
```diff
@@ -0,0 +1,175 @@
...
+    return weights, config.to_dict(), tokenizer
+
+
+def quantize(weights, config, args):
+    quantized_config = copy.deepcopy(config)
+
+    # Load the model:
+    model = Model(ModelArgs.from_dict(config))
+    weights = tree_map(mx.array, weights)
```
angeloskath, line 42: When would the weights not be an `mx.array`? Loaded with numpy or sth?

llms/hf_llm/convert.py
```diff
@@ -0,0 +1,175 @@
...
+
+    repo_id = f"mlx-community/{name}"
+
+    card = ModelCard.load(hf_path)
+    card.data.tags = ["mlx"] if card.data.tags is None else card.data.tags + ["mlx"]
+    card.text = f"""
+# {name}
+This model was converted to MLX format from [`{hf_path}`]().
+Refer to the [original model card](https://huggingface.co/{hf_path}) for more details on the model.
+## Use with mlx
```
angeloskath, line 86: ❤️

llms/hf_llm/generate.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def generate(
+    model: models.Model,
+    tokenizer: transformers.AutoTokenizer,
+    prompt: str,
+    max_tokens: int,
+    temp: float = 0.0,
+):
+    prompt = tokenizer(args.prompt, return_tensors="np", return_attention_mask=False,)[
```
angeloskath, line 18: Completely nitpicky but that probably doesn't pass black cause of the last comma that remained when moving it from multi-line to single line.

llms/hf_llm/generate.py
```diff
@@ -0,0 +1,82 @@
...
+    tokens = []
+    skip = 0
+    for token, n in zip(
+        models.generate(prompt, model, args.temp),
+        range(args.max_tokens),
+    ):
+        if token == tokenizer.eos_token_id:
+            break
+
+        if n == 0:
```
angeloskath, line 33: That should be after `token.item()` to properly measure the timing of the prompt processing.

angeloskath (approved):
Left a couple of comments mostly questions except one small bug. Excited to play with this!

llms/hf_llm/generate.py
awni: Actually the `==` above forces the eval. Sorry it's very implicit...

llms/hf_llm/generate.py
awni: Huh that is weird. I ran `pre-commit run --all-files` and it didn't catch that... you are right though black doesn't like it.

awni: Ok I'm bumping the pre-commit black version in our config. It properly catches this now.

llms/hf_llm/convert.py
awni: Oversight, good catch.

llms/hf_llm/generate.py
angeloskath: Oh nice! Completely overlooked that 🤦‍♂️

singhaki:
Small typo in the readme for conversion it's hf-path instead of hf-model

Vaibhavs10:
Raised a quick PR to fix it: https://github.com/ml-explore/mlx-examples/pull/221

PR #219, opened 2024-01-03
Qlora

awni:
Enables QLoRa with the LoRA example.

### DIFF
--- a/lora/models.py
+++ b/lora/models.py
@@ -1,12 +1,31 @@
 # Copyright © 2023 Apple Inc.
-
 import math
 from dataclasses import dataclass
 from typing import List, Optional, Tuple
 
 import mlx.core as mx
 import mlx.nn as nn
+import numpy as np
 from mlx.utils import tree_map, tree_unflatten
+
+
+# TODO remove this for v0.0.7
+def patch_ql(self, x):
+    x = mx.quantized_matmul(
+        x.astype(mx.float16),
+        self.weight.T,
+        scales=self.scales,
+        biases=self.biases,
+        group_size=self.group_size,
+        bits=self.bits,
+    )
+    x = mx.stop_gradient(x)
+    if "bias" in self:
+        x = x + self.bias
+    return x
+
+
+nn.QuantizedLinear.__call__ = patch_ql
 
 
 @dataclass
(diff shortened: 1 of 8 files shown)


lora/models.py
```diff
@@ -25,6 +24,8 @@ class LoRALinear(nn.Module):
     @staticmethod
     def from_linear(linear: nn.Linear, rank: int = 8):
         output_dims, input_dims = linear.weight.shape
+        if isinstance(linear, nn.QuantizedLinear):
+            input_dims *= 2 * linear.bits
```
angeloskath, line 28: That should be `input_dims *= 32 // linear.bits` so that it works with bits other than 4. But obviously this points towards adding the following properties in `nn.QuantizedLinear` and `nn.Linear`:
- input_dims
- output_dims
- dtype

angeloskath: Maybe we should add a TODO to simplify those when/if the above properties are added in a future release.

lora/models.py
```diff
@@ -47,7 +48,10 @@ def __init__(
         self.lora_b = mx.zeros(shape=(lora_rank, output_dims))
 
     def __call__(self, x):
-        y = self.linear(x.astype(self.linear.weight.dtype))
+        if isinstance(self.linear, nn.QuantizedLinear):
```
angeloskath, line 51: Why not cast in the case of `nn.QuantizedLinear` too?
```python
dtype = self.linear.weight.dtype
if isinstance(self.linear, nn.QuantizedLinear):
    dtype = self.linear.scales.dtype
y = self.linear(x.astype(dtype))
```

angeloskath (approved):
Very nice!

I left a comment for a tiny sneaky bug that would appear only when `bits != 4`.

lora/models.py
awni: We should definitely add those properties. No reason not to.. it's also easy to get the weight shapes mixed up.

lora/models.py
awni: Yes, I like that. Thank you1

mzbac:
Can't wait for this get merged to master. Qlora with a specific domain dataset can easily beat ChatGPT. This will enable huge potential for the community. Thanks a lot!

awni:
Fixes #228

mzbac:
🚀, gonna have a fun weekend 😃

singhaki:
@awni  i am getting this error while using hf model  https://huggingface.co/mlx-community/Mistral-7B-Instruct-v0.2-4-bit

ValueError: [matmul] Last dimension of first input with shape (2,88,4096) must match second to last dimension of second input with shape (512,32000). 

Seems like some config issue where layers are not matching between config and weights.

mzbac:
@singhaki Not yet, I'm still reading the code. But from a quick look at the model cards, it seems that the conversion done earlier may not be compatible with the updated qlora script. Maybe try using the latest convert.py to reconvert it.

PR #234, opened 2024-01-05
refactor: merge deepseek coder example into hf_llm example

mzbac:
Since we already support the hf format model, it is better to merge the deepseek example with the hf llm example to reduce maintenance overhead. so, linear scale rope support was added in hf llm for the deepseek.


llms/hf_llm/requirements.txt
```diff
@@ -1,3 +1,4 @@
 mlx>=0.0.7
 numpy
 transformers
+protobuf
```
mzbac, line 4: This is to solve the tokenizer error for the Yi model. Otherwise, we have to disable use_fast in the tokenizer.

llms/hf_llm/README.md
```diff
@@ -45,6 +45,8 @@ Here are a few examples of Hugging Face models which work with this example:
 - [mistralai/Mistral-7B-v0.1](https://huggingface.co/mistralai/Mistral-7B-v0.1)
 - [meta-llama/Llama-2-7b-hf](https://huggingface.co/meta-llama/Llama-2-7b-hf)
 - [TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T](https://huggingface.co/TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T)
+- [deepseek-ai/deepseek-coder-6.7b-instruct](https://huggingface.co/deepseek-ai/deepseek-coder-6.7b-instruct)
+- [01-ai/Yi-6B-Chat](https://huggingface.co/01-ai/Yi-6B-Chat)
```
awni, line 49: ❤️

llms/hf_llm/models.py
```diff
@@ -73,8 +82,17 @@ def __init__(self, args: ModelArgs):
         self.k_proj = nn.Linear(dim, n_kv_heads * head_dim, bias=False)
         self.v_proj = nn.Linear(dim, n_kv_heads * head_dim, bias=False)
         self.o_proj = nn.Linear(n_heads * head_dim, dim, bias=False)
+        rope_scale = (
+            1 / args.rope_scaling.get("factor", 1.0)
+            if args.rope_scaling is not None
+            and args.rope_scaling.get("type") == "linear"
```
awni, line 88: Are there other scaling types for RoPE?

awni, line 88: Also I would just do the following since you have to have a factor if the type is linear, and you have to have a type.

```suggestion
            1 / args.rope_scaling["factor"]
            if args.rope_scaling is not None
            and args.rope_scaling["type"] == "linear"
```

awni (commented):
This is really great! I love removing unneeded code!! Than you!

Minor comment in the code. Also could you run the pre-commit hooks (see CONTRIBUTING.md) for formatting?

llms/hf_llm/models.py
mzbac: From the Transformers source code, it seems like there is another type called `dynamic`, but I haven't seen it in any of the model's configurations.
https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L338

mzbac: Yeah, it's just a habit of my defensive coding style. I'm happy to change the preferred style.

awni: Yea exactly.. and I was also thinking if we get to this point and those values don't exist.. crashing is probably a good thing.

mzbac:
Just run the pre-commit hook and update the linting fix. Maybe we could add a GitHub action to run PR checks  for pre-commit hook

llms/hf_llm/requirements.txt
awni: Should we enable `use_fast` in the tokenizer? It is not enabled right now.

awni:
> Maybe we could add a GitHub action to run PR checks for the pre-commit hook

Unfortunately we don't have permission for that. We are trying to setup a CI for external PRs but it might be a few weeks out.

mzbac:
> > Maybe we could add a GitHub action to run PR checks for the pre-commit hook
> 
> Unfortunately we don't have permission for that. We are trying to setup a CI for external PRs but it might be a few weeks out.

Alright, hope it gets sorted out soon. For now, I will try to add a local pre-commit hook :)

awni:
> Alright, hope it gets sorted out soon. For now, I will try to add a local pre-commit hook :)

If you run `pre-commit install` it will auto check any file you touch when you try to commit it. You should only need to install the hooks once.

mzbac:
> > Alright, hope it gets sorted out soon. For now, I will try to add a local pre-commit hook :)
> 
> If you run `pre-commit install` it will auto check any file you touch when you try to commit it. You should only need to install the hooks once.

Cool, thanks. I should read the documentation properly :)

llms/hf_llm/requirements.txt
mzbac: The use_fast is enabled by default. -> https://github.com/huggingface/transformers/blob/main/src/transformers/models/auto/tokenization_auto.py#L668

llms/hf_llm/models.py
awni: Oh great, thanks!

awni (approved):
🚀 thanks @mzbac !

PR #237, opened 2024-01-05
Prompt Lookup Decoding - merged under Speculative example 

LeonEricsson:
_Continuation of #202. Decided to merge the Prompt Lookup Decoding under the Speculative Decoding example._ 

This PR implements a example for the "Prompt Lookup Decoding" technique:

https://github.com/apoorvumang/prompt-lookup-decoding

- This approach works best on input-grounded tasks such as summarization, document QA, code editing etc where there is a high overlap in input and output.
- It replaces the draft model in speculative decoding with a simple ngram search of the input tokens. Similar to speculative decoding it has no impact on output.

TODO
- [x] Debug: Model output doesn't have any spaces
- [x] Add `--color` flag to SpeculativeDecoder
*Ended up being quite a messy implementation; need to deal with the fact that output can be truncated and hence only come from draft model*
- [x] Update README
- [x] Code cleaning


LeonEricsson:
@awni perhaps we can leave this as T5 and then make an attempt at swapping to Llama in a new PR? I was thinking we could adopt the model format / conversion from `hf_llm` as well while we're still at it? I could take a first crack at it.

llms/speculative_decoding/speculative.py
```diff
@@ -81,7 +78,7 @@ def main(args):
     )
     parser.add_argument(
         "--prompt",
-        default="translate English to French: Let's go to the store and buy some groceries including eggs, avocadoes, and bread.",
+        default="Translate the following from English to English: Let's go to the store and buy some groceries including eggs, avocadoes, and bread.",
```
awni, line 81: Is it a typo to have English to English here?

llms/speculative_decoding/decoder.py
```diff
@@ -91,6 +93,7 @@ def generate(
         print()
         self.model.reset_cache()
 
+    # Accept / Reject criteria (see Section 2.3 https://arxiv.org/pdf/2211.17192.pdf)
```
awni, line 96: 👍

llms/speculative_decoding/prompt_lookup.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def load_model(model_name: str):
+    config = T5Config.from_pretrained(model_name)
+    model = Model(config)
+    weights = mx.load(f"{model_name}.npz")
+    weights = tree_unflatten(list(weights.items()))
+    model.update(weights)
+    mx.eval(model.parameters())
+    return model
```
awni, line 18: Let's refactor this into `Model` since they both use the exact same functionality?

llms/speculative_decoding/speculative.py
```diff
@@ -59,12 +56,12 @@ def main(args):
...
         "--model-name",
         help="Name of the model.",
-        default="t5-small",
+        default="t5-base",
     )
     parser.add_argument(
         "--draft-model-name",
         help="Name of the draft model.",
-        default="t5-small",
+        default="t5-base",
```
awni, line 64: Shouldn't we use a large `t5-large` for the main model and a `t5-small` for the base? Otherwise doesn't seem like there is much point to specualitive decoding. I think I set bad defaults before but maybe we should fix now.

llms/speculative_decoding/prompt_lookup.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+    parser.add_argument(
+        "--temp",
+        help="The sampling temperature.",
+        type=float,
+        default=0.0,
+    )
+    parser.add_argument("--seed", type=int, default=0, help="The PRNG seed")
+    parser.add_argument(
+        "--color", type=bool, default=False, help="Color the accepted draft tokens"
```
awni, line 95: I love that, it's super cool to see visually how it works!

awni (changes requested):
Looks really nice! I think we can get this in soon. I didn't look yet at the core of the prompt decoder but left a few comments. 

Thanks a ton for refactoring them together, I think it makes a lot of sense this way.

llms/speculative_decoding/speculative.py
LeonEricsson: yes! seems I managed to push some debugging settings.

LeonEricsson:
> Looks really nice! I think we can get this in soon. I didn't look yet at the core of the prompt decoder but left a few comments.
> 
> Thanks a ton for refactoring them together, I think it makes a lot of sense this way.

Thanks! Addressed all your comments

cmcmaster1:
I've been poking around your code @LeonEricsson because I have some long summarization tasks that I'd like to speed up, but noticed a significant bottleneck from the loop. This is probably still not perfect, but I've had a go at speeding it up. This implementation is about 500x faster:
```python
def find_draft(self, input_ids):
            # Convert MLX array to NumPy for vectorized operations
            input_ids_np = np.array(input_ids)
            
            # Create a sliding window of the last ngram_max tokens
            ngram = input_ids_np[-self.ngram_max:]
            
            # Vectorized comparison of ngram with all possible sub-arrays of input_ids
            matches = np.lib.stride_tricks.sliding_window_view(input_ids_np, self.ngram_max) == ngram
            
            # Check if all elements in ngram match for each sub-array
            match_indices = np.all(matches, axis=1).nonzero()[0]
            
            # Filter out matches that are too short or overlap with the ngram itself
            match_indices = match_indices[(match_indices + self.ngram_max <= input_ids_np.size - self.ngram_max) & (match_indices >= self.ngram_min)]
            
            # Find the largest match
            if match_indices.size > 0:
                largest_match_idx = match_indices[-1]  # Assuming the last match is the largest
                start_idx = largest_match_idx + self.ngram_max
                end_idx = min(start_idx + self.n_draft, input_ids_np.size)
                candidate = input_ids_np[start_idx:end_idx]
                
                # Convert the candidate back to MLX array
                return mx.array(candidate, dtype=mx.uint32)
            
            return mx.array([], dtype=mx.uint32)
```

LeonEricsson:
> I've been poking around your code @LeonEricsson because I have some long summarization tasks that I'd like to speed up, but noticed a significant bottleneck from the loop. This is probably still not perfect, but I've had a go at speeding it up. This implementation is about 500x faster:

nice :rocket: the original implementation uses np sliding windows but I wanted to keep things pure mlx so I decided against it. That being said, with this being an example we should strive for what's best for the user and such a performance bottleneck is a big problem so I agree it should be changed. It might be possible to achieve the same speedups with mlx.core.vmap?

cmcmaster1:
That makes sense. And I'm guessing you didn't notice a huge performance gap, because you didn't try it on long texts? I'm shaving ~30 seconds off inference time. I was thinking about trying a vmap version next.

LeonEricsson:
@cmcmaster1  I've implemented a pure MLX version that should be comparable in performance to the numpy one. Would be great if you could confirm this on your end. **However**, before you do so note that your current implementation does not consider ngram matches other than of size `self.ngram_max`, which is not aligned with how Prompt Lookup was originally proposed. The idea of prompt lookup is to iteratively check for smaller ngram keys until you get to self.ngram_min. Note the for loop [here](https://github.com/ml-explore/mlx-examples/blob/9bb01a7383b9f5cbdc6648435bb8b0a99c986e44/llms/speculative_decoding/decoder.py#L247). I spent a lot of time trying to do away with this for loop and letting `mlx.core` do the work but couldn't find a way. 

@awni imo this is ready to be merged, sorry for the delay.

cmcmaster1:
@LeonEricsson oops, you're right. I somehow missed that and just tested on examples where it made no difference! Still much faster than the original and definitely comparable to the (flawed) numpy implementation.

awni:
Sorry for the delay!! I will review and get this in early next week

awni:
Yes, let's close it. We could look into making prompt lookup speculation an option in MLX LM as an alternative to using a draft model. @LeonEricsson if you're interesting in contributing that feel free to send a PR.

PR #238, opened 2024-01-05
add numpy as a requirement to run lora.py

lawwu:
Running the `convert.py` script in the lora example gives this error. I added numpy as a dependency.

```
 python convert.py \
    --torch-path mistral-7B-v0.1 \
    --mlx-path mistral-7B-v0.1-mlx
```

yields this error:
```
  File "/Users/lawrence.wu/github/mlx-examples/lora/convert.py", line 11, in <module>
    import numpy as np
```

### DIFF
--- a/lora/lora.py
+++ b/lora/lora.py
@@ -5,13 +5,13 @@ import json
 import math
 import time
 from pathlib import Path
-from typing import List, Optional, Tuple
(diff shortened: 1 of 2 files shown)


awni (approved):
Thanks!

PR #240, opened 2024-01-06
Force fp16 for non-quantized weights in quantized models

awni:
- will be much faster
h/t @angeloskath for pointing this out


angeloskath (approved):
Nice!

PR #241, opened 2024-01-06
Rename requirements.txt for windows file name support

ornfelt:
It's currently not possible to clone / do git pull on Windows due the file llms/deepseek-coder/requirements.txt having a space at the end. ".txt " instead of ".txt"


awni:
Thanks, I'm about to merge #234 which will remove that file anyway and merge it into `hf_llm`

PR #242, opened 2024-01-06
Update CODE_OF_CONDUCT.md

NinoRisteski:
few typos


awni:
Thanks for the contribution. The code of conduct is from Apple admin, so let's keep it fixed.

Issue #232, opened 2024-01-05
LoRA finetuning with model downloaded from mlx-community fails


shreyaskarnik:
I downloaded pre-converted model from mlx-community from HuggingFace using the command 
```bash 
huggingface-cli download --local-dir-use-symlinks False --local-dir tinyllama-1.1B-Chat-v0.6 mlx-community/tinyllama-1.1B-Chat-v0.6
``` 
and tried to run 
```bash
python lora.py --model ./tinyllama-1.1B-Chat-v0.6 \
               --train \
               --iters 600
```
and got this error 
```python
Loading pretrained model
Traceback (most recent call last):
  File "/PATH/mlx-examples/lora/lora.py", line 360, in <module>
    model, tokenizer = load_model(args.model)
                       ^^^^^^^^^^^^^^^^^^^^^^
  File "/PATH/mlx-examples/lora/lora.py", line 339, in load_model
    model_args = ModelArgs(**config)
                 ^^^^^^^^^^^^^^^^^^^
TypeError: ModelArgs.__init__() got an unexpected keyword argument 'rope_traditional'
```

Is there a way to add support for fine-tuning the models from mlx-community?

singhaki:
For this issue you can remove this param from config file. 'rope_traditional' = False for tinyllama and you might also find issue with head_dim. You can check code how it's calculated from convert script. it's config["dim"] // n_heads.

awni:
Thanks @singhaki for the input.

Regarding the rope_traditional flag, I wouldn't remove it as the model won't work as well. We can add the config param to our model so that it interprets it correctly, that's an easy fix.

The other config options you can add to the json as needed to make sure they are there. I am planning to move the LoRA example to be more like our HF LLM example to simplify model conversions etc. Should solve this problem in a cleaner way.

singhaki:
I removed it from config and added it in code. I tried like 1 week back before hf integration.

shreyaskarnik:
Thanks for the responses @singhaki and @awni really appreciate it

shreyaskarnik:
I was able to work around and get a model loaded from mlx-community for fine-tuning here is the modified `models.py` which I used https://gist.github.com/shreyaskarnik/dbe75ebba5d4819419f9ea2526650ba0

mzbac:
[...]

Issue #212, opened 2024-01-01
How can I merge the Lora weight back to the original model weight?


mzbac:
Maybe I'm missing something, but I can't find any information on how to merge Lora weight back into the original model. Running the model with a Lora adapter will add additional memory overhead and make it slightly more difficult to distribute. I wonder if we could provide the script for merging Lora weight back into the model?

awni:
Yea that's missing in our example. I think it would be nice to have an option on the LoRA layer which merges the adapters and the linear weights after the adapters are trained or when using them in generation.

mzbac:
Thanks for sharing your thoughts. I plan to use MLX to fine-tune some of my models. I will try to make it work and contribute back if possible.

mzbac:
@awni I only managed to create a merge function(https://github.com/mzbac/mlx-lora/blob/main/models.py#L92-L107) in loraLinear and loop through all the named modules to get the merged linear layer, then update the modules(https://github.com/mzbac/mlx-lora/blob/main/utils.py#L22-L28). However, this is not efficient because I have to make copies of the linear layers and update them. I am wondering if mlx have a method that allows us to map the module with lambda so we can replace the layer without making additional copies?

awni:
You needn't worry about this being inefficient:

```
        self.linear.weight += (self.lora_a @ self.lora_b).T * 2.0
        new_linear = nn.Linear(input_dims, output_dims, bias=False)
        new_linear.weight = self.linear.weight
```

The `new_linear.weight` is not doing a deep copy (under the hood) it will just point to the same data as `self.linear.weight`.

Can I ask: what is your intention with merging? My understanding is it is pretty uncommon to save the fully merged model (because you can easily restore it from the original model and the adapters). 

What is more common is to merge dynamically to avoid the additional expense of forming the low rank update when you are using the model. From that perspective it might make more sense to have an "eval mode" on the LoRALinear layer that merges it.

mzbac:
> You needn't worry about this being inefficient:
> 
> ```
>         self.linear.weight += (self.lora_a @ self.lora_b).T * 2.0
>         new_linear = nn.Linear(input_dims, output_dims, bias=False)
>         new_linear.weight = self.linear.weight
> ```
> 
> The `new_linear.weight` is not doing a deep copy (under the hood) it will just point to the same data as `self.linear.weight`.
> 
[...]

PR #252, opened 2024-01-07
Move lora example to use the same model format / conversion as `hf_llm`

awni:
So now you can just run it with an HF repo or convert directly from an HF repo (and optionally upload).

Should work with a lot more models out of the box.

FYI right now this basically copies the convert from `hf_llm` and the model as well. In the future it might/probably makes sense to merge the two (especially if it gets subsumed into a package).

### DIFF
--- a/lora/README.md
+++ b/lora/README.md
@@ -33,8 +33,8 @@ pip install -r requirements.txt
 
 ### Convert
 
-This step is optional if you want to quantize (for QLoRA) or change the default data type
-of a pre-existing model.
+This step is optional if you want to quantize (for QLoRA) or change the default
+data type of a pre-existing model.
 
 You convert models using the `convert.py` script. This script takes a Hugging
 Face repo as input and outputs a model directory (which you can optionally also
--- a/lora/README.md
+++ b/lora/README.md
@@ -18,6 +18,7 @@ be general should you wish to use a custom dataset.
   * [Evaluate](#Evaluate)
   * [Generate](#Generate)
 * [Results](#Results)
+* [Fuse and Upload](#Fuse-and-Upload)
 * [Custom Data](#Custom-Data)
 * [Memory Issues](#Memory-Issues)
 
@@ -132,6 +133,38 @@ training and validation loss at a few points over the course of training.
 | 1000      |    1.070   |      1.230      |
 
 The model trains at around 475 tokens per second on an M2 Ultra.
+
+## Fuse and Upload
+
+You can generate a fused model with the low-rank adapters included using the
+`fuse.py` script. This script also optionally allows you to upload the fused
+model to the [Hugging Face MLX
+Community](https://huggingface.co/mlx-community).
+
+To generate the fused model run:
+
+```
+python fuse.py
+```
+
+This will by default load the base model from `mlx_model/`, the adapters from
+`adapters.npz`,  and save the fused model in the path `lora_fused_model/`. All
+of these are configurable. You can see the list of options with:
+
+```
+python fuse.py --help
+```
+
+To upload a fused model, supply the `--upload-name` and `--hf-path` arguments
+to `fuse.py`. The latter is the repo name of the original model which is
+useful for the sake of attribution and model versioning.
+
+For example to fuse and upload a model derived from Mistral-7B-v0.1, you can
+run:
+
+```
+python fuse.py --upload My-4-bit-model --hf-repo mistralai/Mistral-7B-v0.1
+```
 
 ## Custom Data
 
--- a/lora/README.md
+++ b/lora/README.md
@@ -156,11 +156,10 @@ python fuse.py --help
 ```
 
 To upload a fused model, supply the `--upload-name` and `--hf-path` arguments
-to `fuse.py`. The latter is the repo name of the original model which is
+to `fuse.py`. The latter is the repo name of the original model, which is
 useful for the sake of attribution and model versioning.
 
-For example to fuse and upload a model derived from Mistral-7B-v0.1, you can
-run:
+For example, to fuse and upload a model derived from Mistral-7B-v0.1, run: 
 
 ```
 python fuse.py --upload My-4-bit-model --hf-repo mistralai/Mistral-7B-v0.1
--- a/lora/README.md
+++ b/lora/README.md
@@ -45,8 +45,8 @@ To make a 4-bit quantized model, run:
 python convert.py --hf-path <hf_repo> -q
 ```
 
-For example the following will make a 4-bit quantized Mistral 7B and by default
-store it in `mlx_model` 
+For example, the following will make a 4-bit quantized Mistral 7B and by default
+store it in `mlx_model`:
 
 ```
 python convert.py --hf-path mistralai/Mistral-7B-v0.1 -q
@@ -59,12 +59,12 @@ python convert.py --help
 ```
 
 You can upload new models to the [Hugging Face MLX
-Community](https://huggingface.co/mlx-community) by specifying `--upload-name``
+Community](https://huggingface.co/mlx-community) by specifying `--upload-name`
 to `convert.py`.
 
 ## Run
 
-The main script is `lora.py`. To see a full list of options run
+The main script is `lora.py`. To see a full list of options run:
 
 ```
 python lora.py --help
@@ -94,7 +94,7 @@ You can resume fine-tuning with an existing adapter with `--resume-adapter-file
 
 ### Evaluate
 
-To compute test set perplexity use
+To compute test set perplexity use:
 
 ```
 python lora.py --model <path_to_model> \
@@ -104,7 +104,7 @@ python lora.py --model <path_to_model> \
 
 ### Generate
 
-For generation use
+For generation use:
 
 ```
 python lora.py --model <path_to_model> \
@@ -176,7 +176,7 @@ For example, for a machine with 32 GB the following should run reasonably fast:
 
 ```
 python lora.py \
-   --model <path_to_model> \
+   --model mistralai/Mistral-7B-v0.1 \
    --train \
    --batch-size 1 \
    --lora-layers 4
--- a/lora/README.md
+++ b/lora/README.md
@@ -45,7 +45,7 @@ To make a 4-bit quantized model, run:
 python convert.py --hf-path <hf_repo> -q
 ```
 
-For exampe the following will make a 4-bit quantized Mistral 7B and by default
+For example the following will make a 4-bit quantized Mistral 7B and by default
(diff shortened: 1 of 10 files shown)


awni:
CC @pcuenca @Vaibhavs10

lora/README.md
```diff
@@ -58,6 +70,9 @@ The main script is `lora.py`. To see a full list of options run
 python lora.py --help
 ```
 
+Note, in the following the `--model` argument can be any compatable Hugging
+Face repo or a local path to a converted mdoel.
```
pcuenca, line 74:
```suggestion
Note, in the following the `--model` argument can be any compatible Hugging
Face repo or a local path to a converted model. 
```

lora/README.md
```diff
@@ -28,27 +30,37 @@ Install the dependencies:
...
 ```
-curl -O https://files.mistral-7b-v0-1.mistral.ai/mistral-7B-v0.1.tar
-tar -xf mistral-7B-v0.1.tar
+python convert.py --hf-path <hf_repo> -q
 ```
 
-If you do not have access to the Llama weights you will need to [request
-access](https://ai.meta.com/resources/models-and-libraries/llama-downloads/)
-from Meta.
+For exampe the following will make a 4-bit quantized Mistral 7B and by default
```
pcuenca, line 48:
```suggestion
For example, the following will make a 4-bit quantized Mistral 7B and by default
```

pcuenca (commented):
Very cool! 🔥 I'll take a closer look and test it in the morning, thanks a lot!

mzbac:
Nice work! Just a quick question: for the best fine-tune result, would it be possible to provide a flag that allows people to apply Lora to all linear layers (except lm_head) instead of q and v layer? Just try to avoid giving people the wrong impression when they try mlx-example lora and don't get good outcomes, causing them to give up.

awni:
> Nice work! Just a quick question: for the best fine-tune result, would it be possible to provide a flag that allows people to apply Lora to all linear layers (except lm_head) instead of q and v layer? Just try to avoid giving people the wrong impression when they try mlx-example lora and don't get good outcomes, causing them to give up.

We can definitely provide that. But I should add that there are a lot of ways LoRA training can go wrong (the learning rate may need to be tuned, the rank may be too low, overfitting (too many layers / too high rank)).  As I'm writing this, I'm realizing it's worth maybe adding a section in the README on tuning LoRA for best results :)

mzbac:
> > Nice work! Just a quick question: for the best fine-tune result, would it be possible to provide a flag that allows people to apply Lora to all linear layers (except lm_head) instead of q and v layer? Just try to avoid giving people the wrong impression when they try mlx-example lora and don't get good outcomes, causing them to give up.
> 
> We can definitely provide that. But I should add that there are a lot of ways LoRA training can go wrong (the learning rate may need to be tuned, the rank may be too low, overfitting (too many layers / too high rank)). As I'm writing this, I'm realizing it's worth maybe adding a section in the README on tuning LoRA for best results :)

Yeah, makes sense. I would like to help with the documentation once this is merged. After it gets merged, I am thinking to generalizing the LoRa parameters, such as lora_alpha, r andlora_dropout in loraLinear instead of hardcoded. It would be worth to update the documentation with these changes.

Vaibhavs10 (approved):
Love it! Two things that I think can amplify the impact:

1. Allow users to checkpoint their adapter files over to the Hub. They can do so via their Hub IDs (ofc optional) - https://github.com/ml-explore/mlx-examples/blob/1ab4a1620e892974e9135a290d24693ad8fce0f2/lora/lora.py#L327C48-L327C48 -> This way no matter what happens they'd have proper versioning of their adapter checkpoints through epochs too.

2. IMO we can create a `fuse.py` file which essentially pushes a fused model (adapters + base merged) over to the Hub so that people can use it with `hf_llm` package too. This would essentially be the logic here: https://github.com/ml-explore/mlx-examples/blob/1ab4a1620e892974e9135a290d24693ad8fce0f2/lora/models.py#L317 + pushing to hub.

Vaibhavs10:
Definitely agree on this: 
> We can definitely provide that. But I should add that there are a lot of ways LoRA training can go wrong (the learning rate may need to be tuned, the rank may be too low, overfitting (too many layers / too high rank)). As I'm writing this, I'm realizing it's worth maybe adding a section in the README on tuning LoRA for best results :)

awni:
I added a fuse.py which fuses the model and optionally uploads it to the hub. H/T @mzbac for starting that.

awni:
Closes #232 and #212

lora/models.py
```diff
@@ -32,14 +90,58 @@ def from_linear(linear: nn.Linear, rank: int = 8):
...
+        linear = self.linear
+        bias = "bias" in linear
+        weight = linear.weight
+        is_quantized = isinstance(linear, nn.QuantizedLinear)
+
+        # Use the same type as the linear weight if not quantized
+        dtype = weight.dtype
+
+        if is_quantized:
+            dtype = mx.float16
```
angeloskath, line 103: This could be `dtype = linear.scales.dtype` but I think it might be better to just fix it to `float16` for quantized layers anyway.

lora/models.py
```diff
@@ -32,14 +90,58 @@ def from_linear(linear: nn.Linear, rank: int = 8):
...
+        return fused_linear
+
     def __init__(
-        self, input_dims: int, output_dims: int, lora_rank: int = 8, bias: bool = False
+        self,
+        input_dims: int,
+        output_dims: int,
+        lora_rank: int = 8,
+        bias: bool = False,
+        scale: float = 20.0,
```
angeloskath, line 135: More of a curiosity, is the 20 a better default than 2?

angeloskath (approved):
Looks perfect!

lora/models.py
awni: It trains much faster with QLoRA and fusing didn’t work at all with a smaller scale. The low rank update was getting completely wiped out in the dequantize - requantize part.

We should let it be configurable as an arg, but so far 20 was consistently better in the cases I tried so it’s a step in the right direction.

mzbac: I have noticed in the past that when I try to merge lora weight using transformers' peft library, I had some performance degradation. Now I am wondering if it was caused by this issue. However, by quickly checking the source code, it seems like they are using a quick small scaling (16/64). I'm not sure how it works. 
https://github.com/artidoro/qlora/blob/main/qlora.py#L177-L185
https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/layer.py#L90-L93

awni: That's interesting 🤔  .. maybe they initialize the adapters differently?

mzbac: I will read their code a bit more today to see if I can find anything.

mzbac: @awni I just double checked, you are right. It's using kaiming_uniform to initialize lora a. but, I am not sure if that will cause such differences.
FYI: https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/layer.py#L120

awni: It's not inconceivable the initialization would make a difference. I can't tell just from a glance at that code what the scale of the initialization is. But if the vector started out with larger numbers in general that could offset the smaller scale.

Then again there could be a totally different explanation..

lora/fuse.py
```diff
@@ -0,0 +1,69 @@
...
+        if isinstance(m, models.LoRALinear)
+    ]
+
+    model.update_modules(tree_unflatten(fused_linears))
+    weights = dict(tree_flatten(model.parameters()))
+    utils.save_model(args.save_path, weights, tokenizer._tokenizer, config)
+    if args.upload_name is not None:
+        if args.hf_path is None:
+            raise ValueError("Must provide original Hugging Face repo to upload model")
+        utils.upload_model(args.save_path, args.upload_name, args.hf_path)
```
pcuenca, line 69:
```suggestion
        utils.upload_to_hub(args.save_path, args.upload_name, args.hf_path)
```

lora/fuse.py
```diff
@@ -0,0 +1,69 @@
...
+import models
+import utils
+from mlx.utils import tree_flatten, tree_unflatten
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="LoRA or QLoRA finetuning.")
+    parser.add_argument(
+        "--model",
+        default="mlx_model",
+        help="The path to the local model directory.",
```
pcuenca, line 15:
```suggestion
        help="The path to the local model directory or Hugging Face repo in the Hub to use as the base model.",
```

lora/fuse.py
```diff
@@ -0,0 +1,69 @@
...
+    parser.add_argument(
+        "--save-path",
+        default="lora_fused_model",
+        help="The path to save the fused model.",
+    )
+    parser.add_argument(
+        "--adapter-file",
+        type=str,
+        default="adapters.npz",
+        help="Save/load path for the trained adapter weights.",
```
pcuenca, line 26:
```suggestion
        help="Path to the trained adapter weights, in numpy or safetensors format.",
```

lora/fuse.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+    parser.add_argument(
+        "--hf-path",
+        help="Path to the original Hugging Face model (required for upload).",
+        type=str,
+        default=None,
+    )
```
pcuenca, line 33: This was a bit confusing to me, especially when you use a remote base model in `--model`. Perhaps we could allow it to be `None`, default to `--model`, and verify that it's a HF remote repo. We can do that in a followup PR.

pcuenca (commented):
Thanks for the great feature! There appears to be a typo in the invocation to the upload function :)

The workflow can be a bit confusing if you just want to fuse existing LoRAs from the Hub. Some ideas for future improvement that could be discussed in future PRs:
- Allow the adapter to be downloaded from the Hub. It must be a safetensors file, though.
- Many LoRAs point to the base model using the `base_model` metadata tag: https://huggingface.co/nminnie/zephyr-7b-sft-lora/blob/d320cc976b311d12e00ace3d033a32891d69c77d/README.md?code=true#L3. We could leverage that and just indicate the adapter to fuse.
- If you want to fuse a model in the Hub and upload the fused copy, you have to indicate the base model name twice. We could default `--hf-path` to `--model` and try to do the right thing if we can, or error when it's not possible.

lora/fuse.py
awni: This is a really good point. I didn't even consider that `--model` could actually be an HF repo. What I could do is check if `model` exists as a path. If it does, then it is a local path that we loaded from and we require `hf-path`. If it doesn't then it must be an HF repo and we can safely set `hf-path` to it.

PR #253, opened 2024-01-08
refactor: make the phi2 example can be directly load the model from hf without convert needed

mzbac:
Just try to make the phi2 example compatible with the phi2 model structure on HF, so people can directly use it without needing to convert. Additionally, given the small size of the model, it will be beneficial if people could directly load the full precision on the model without converting it first. 
Also, currently the open source LLM community prefers distributing fine-tuned models with lora weight merged, so keep the model structure consistent could avoid confusion for those distributing mlx fine-tuned models later.

But please feel free to dismiss this PR if it doesn't fit the purpose.

### DIFF
--- a/llms/phi2/README.md
+++ b/llms/phi2/README.md
@@ -7,62 +7,51 @@ GPT-4 outputs and clean web text.
 Phi-2 efficiently runs on Apple silicon devices with 8GB of memory in 16-bit
 precision.
 
-## Setup
+### Setup
 
-Download and convert the model:
-
-```sh
-python convert.py
-```
-
-To generate a 4-bit quantized model use the `-q` flag:
+Install the dependencies:
 
 ```
-python convert.py -q
+pip install -r requirements.txt
 ```
 
-By default, the conversion script will make the directory `mlx_model` and save
-the converted `weights.npz`, and `config.json` there.
-
-> [!TIP] Alternatively, you can also download a few converted checkpoints from
-> the [MLX Community](https://huggingface.co/mlx-community) organization on
-> Hugging Face and skip the conversion step.
-
-
-## Generate
-
-To generate text with the default prompt:
-
-```sh
-python phi2.py
+### Run
 ```
-
-Should give the output:
+python generate.py --model <model_path> --prompt "hello"
+```
+For example:
 
 ```
-Answer: Mathematics is like a lighthouse that guides us through the darkness of
-uncertainty. Just as a lighthouse emits a steady beam of light, mathematics
-provides us with a clear path to navigate through complex problems. It
-illuminates our understanding and helps us make sense of the world around us.
+python generate.py --model microsoft/phi-2 --prompt "hello"
+```
+The `<model_path>` should be either a path to a local directory or a Hugging
+Face repo with weights stored in `safetensors` format. If you use a repo from
+the Hugging Face Hub, then the model will be downloaded and cached the first
+time you run it. 
 
-Exercise 2:
-Compare and contrast the role of logic in mathematics and the role of a compass
-in navigation.
+Run `python generate.py --help` to see all the options.
 
-Answer: Logic in mathematics is like a compass in navigation. It helps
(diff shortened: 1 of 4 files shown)


llms/phi2/README.md
```diff
@@ -7,63 +7,52 @@ GPT-4 outputs and clean web text.
...
 
+### Convert new models 
 
-## Generate
+You can convert (change the data type or quantize) models using the
+`convert.py` script. This script takes a Hugging Face repo as input and outputs
+a model directory (which you can optionally also upload to Hugging Face).
 
-To generate text with the default prompt:
+For example, to make 4-bit quantized a model, run:
```
awni, line 40: Nit `to make a`

awni (approved):
I think this is great, we should definitely have it this way. Avoiding the required pre-conversion is a nice feature!

awni:
One thing I'm wondering about is if we should attempt to merge with `hf_llm` so that it can load and generate with Phi2 models as well? What do you think @mzbac ?

We'd probably would need to check the `config["model_type"]` field when constructing the model. From what I can see the main differences are the fused qkv matmul and layer norm instead of RMS norm. Not sure if it makes to have a whole separate model file or to try and make those configurable. (I am a bit scared of configuration condition bloat which can be hard to maintain).

  Either way this PR is a step in that direction so I'm going to merge it!

mzbac:
> One thing I'm wondering about is if we should attempt to merge with `hf_llm` so that it can load and generate with Phi2 models as well? What do you think @mzbac ?
> 
> We'd probably would need to check the `config["model_type"]` field when constructing the model. From what I can see the main differences are the fused qkv matmul and layer norm instead of RMS norm. Not sure if it makes to have a whole separate model file or to try and make those configurable. (I am a bit scared of configuration condition bloat which can be hard to maintain).
> 
> Either way this PR is a step in that direction so I'm going to merge it!

It would be great if we could merge all those hf llm models into the `hf_llm`. From what I can see in the transformer source code, it seems like they have separate models for each model and use architectures to identify the model class when instantiating them. may be we can follow the similar approach?https://github.com/huggingface/transformers/blob/main/src/transformers/models/auto/auto_factory.py#L392

awni:
> From what I can see in the transformer source code, it seems like they have separate models for each model and use architectures to identify the model class when instantiating them. may be we can follow the similar approach?

From a simplicity standpoint I think that's the right call. There will probably be a lot of code dup. To keep it organized we could have a `models` subdir and a file for each model type

mzbac:
> > From what I can see in the transformer source code, it seems like they have separate models for each model and use architectures to identify the model class when instantiating them. may be we can follow the similar approach?
> 
> From a simplicity standpoint I think that's the right call. There will probably be a lot of code dup. To keep it organized we could have a `models` subdir and a file for each model type

Hi @awni, I just quickly tried out the approach in one of my repositories. It looks alright and not as repetitive as I originally thought. Let me know what you think. here is the repo: https://github.com/mzbac/mlx-llm/tree/refactor/llm.

awni:
> Hi @awni,I quickly tried out the approach in one of my repositories. It looks alright and not as repetitive as I originally thought. The flow for adding a new model would be to simply add the model args and model class, then update the mapping in utils.py. I feel that the developer experience is kind of smooth.

I love that!  I would say to keep the example relatively flat and understandable, it would be better to have your models in the directory `models/llama.py` and just put everything llama specific file (e.g. including the llama specific model args). That makes even more straightforward to add a new model (1 model is 1 file).  (We still trying to strike a good balance between simplicity and functionality in our examples and sometimes they can be at tensions with one another 😄 )

Wdyt? If that sounds good, would be great to have you PR it?

mzbac:
@awni  Yeah, sounds good to me. The reason I put it in the folder was to provide some flexibility for future extension, but I'm happy to go with one file per model for now. We can always separate it later if needed. I'll start drafting something, hopefully we can get it done soon and enable us to make `hf_llm` a dedicated package.

PR #257, opened 2024-01-08
Updated CIFAR-10 ResNet example to use BatchNorm instead of LayerNorm

menzHSE:
This PR updates the CIFAR-10 ResNet example to use BatchNorm instead of LayerNorm. It is tested with mlx@026ef9a but should probably wait until https://github.com/ml-explore/mlx/pull/385 (remove `retain_graph` flag) appears in an official release. I am assuming, this will be in mlx 0.0.8. 

**Changes**
* Replaced ```nn.LayerNorm``` by ```nn.BatchNorm``` (test accuracy 0.807 after 100 epochs -> 0.833 after 30 epochs)
* Decreased default number of epochs from 100 to 30 (no improvement observed after approx. 25 epochs)
* Updated code comments and ```README.md```
* Updated ```requirements.txt``` to require ```mlx>=0.0.8```

**TODO**

- [x] Test with next official mlx release and verify ```mlx>=0.0.8``` in ```requirements.txt``` is still valid

### DIFF
--- a/cifar/README.md
+++ b/cifar/README.md
@@ -40,8 +40,8 @@ After training with the default `resnet20` architecture for 30 epochs, you
 should see the following results:
 
 ```
-Epoch: 29 | avg. Train loss 0.290 | avg. Train acc 0.899 | Throughput: 279.91 images/sec
-Epoch: 29 | Test acc 0.833
+Epoch: 29 | avg. Train loss 0.294 | avg. Train acc 0.897 | Throughput: 270.81 images/sec
+Epoch: 29 | Test acc 0.841
 ```
 
 Note this was run on an M1 Macbook Pro with 16GB RAM.
--- a/cifar/README.md
+++ b/cifar/README.md
@@ -36,16 +36,15 @@ python main.py --help
 
 ## Results
 
-After training with the default `resnet20` architecture for 100 epochs, you
+After training with the default `resnet20` architecture for 30 epochs, you
 should see the following results:
 
 ```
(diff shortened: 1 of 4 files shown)


awni:
@menzHSE did you record the "Previous" results on the same machine? In general it's expected that there can be considerable variability between machines.

menzHSE:
@awni I did not record the "Previous" results. It is taken from the original ```README.md```. I have tried to reproduce it with earlier mlx versions (up to 0.0.2) but am always getting the same throughput. So, whoever reported the initial results must have a better M1 MBPro than mine ;-)

awni:
Well that is very strange... let me time it on my machine..

awni:
I am on main right now with a 32GB M1 Max. Clearly that seems to help a lot...  I don't have access to a 16GB M1 pro, but I will also plan try on an M1 Mini and/or M2 Mini

```
Epoch 00 [020] | Train loss 2.215 | Train acc 0.223 | Throughput: 624.82 images/second
```

awni (approved):
Looks great, thanks for updating this.

I will wait until 0.0.8 to merge it.

Also the perf for this example needs a lot of work 😞 , even the 400 im/s was really bad...

menzHSE:
Tested with mlx 0.0.9. Updated requirements.txt to require mlx>=0.0.9. Rebased and force-pushed. I think it is ready to be merged.

menzHSE:
Tested with mlx 0.0.9. Ready to be merged.

awni:
Thanks @menzHSE !

PR #260, opened 2024-01-09
Add -local flag to llms/hf_llm/convert.py for reading source HF models from filesystem.

kerekovskik:
This PR is to add a local flag (-l, --local) to convert.py to enable the reading of HF Models already on the filesystem during a conversion operation. Currently the script has to download the model weights from HF directly before beginning the conversion to MLX format. 

Setting the --local flag disables uploads to huggingface during conversion as the hf_path argument would not specify the original model in the modelcard.

### DIFF
--- a/llms/hf_llm/convert.py
+++ b/llms/hf_llm/convert.py
@@ -14,15 +14,20 @@ from mlx.utils import tree_flatten
 from models import Model, ModelArgs
 
 
-def fetch_from_hub(hf_path: str):
-    model_path = snapshot_download(
-        repo_id=hf_path,
-        allow_patterns=["*.json", "*.safetensors", "tokenizer.model"],
-    )
+def fetch_from_hub(hf_path: str,local:bool):
+    
+    if not local:
+        model_path = snapshot_download(
+            repo_id=hf_path,
+            allow_patterns=["*.json", "*.safetensors", "tokenizer.model"],
+        )
+    else: 
+        model_path = hf_path
+
     weight_files = glob.glob(f"{model_path}/*.safetensors")
     if len(weight_files) == 0:
         raise FileNotFoundError("No safetensors found in {}".format(model_path))
-
+    
     weights = {}
     for wf in weight_files:
         weights.update(mx.load(wf).items())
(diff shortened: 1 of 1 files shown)


llms/hf_llm/convert.py
```diff
@@ -33,6 +33,22 @@ def fetch_from_hub(hf_path: str):
     )
     return weights, config.to_dict(), tokenizer
 
+def fetch_from_local(local_path):
+    weight_files = glob.glob(f"{local_path}/*.safetensors") + glob.glob(f"{local_path}/*.bin")
+    if len(weight_files) == 0:
+        raise FileNotFoundError("No safetensors or bin files found in {}".format(local_path))
```
awni, line 39: What is `*.bin`?

llms/hf_llm/convert.py
```diff
@@ -33,6 +33,22 @@ def fetch_from_hub(hf_path: str):
...
+
+    weights = {}
+    for wf in weight_files:
+        if wf.endswith(".safetensors"):
+            weights.update(mx.load(wf).items())
+        else:
+            weights[Path(wf).name] = mx.ndarray.load(wf)
```
awni, line 46: I don't think this will run.. `mx.ndarray`?

llms/hf_llm/convert.py
```diff
@@ -170,5 +198,5 @@ def upload_to_hub(path: str, name: str, hf_path: str):
     with open(mlx_path / "config.json", "w") as fid:
         json.dump(config, fid, indent=4)
 
-    if args.upload_name is not None:
+    if args.upload_name is not None and args.local is not False:
```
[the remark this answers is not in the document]
kerekovskik: Yep, I will take a look when I get a chance to day. Thank you for reviewing this.

llms/hf_llm/convert.py
kerekovskik: The .bin files are regular pytorch files Like the ones you would find in on HF that don't use safetensors. For example, WizardLM/WizardMath-7B-V1.1

awni: Makes sense, thanks. We do not support PyTorch format, so I would remove that part of the code.

kerekovskik: Sure thing, I removed it.

llms/hf_llm/convert.py
kerekovskik: I removed this since only safetensors is supported.

kerekovskik:
I consolidated the fetch_from_local and fech_from_hub functions into one. Though I kept the fetch_from_hub function name as-is. Let me know if you'd like to change that to just load_model or something more neutral as it is not technically fetching from the hub if local=True is passed. 

I also removed all the code related to non-safetensors.

kerekovskik:
@awni , I think I've incorporated everything from your initial review. Please let me know if there is anything else for this PR.

awni:
Thanks a ton! I will check it shortly!

awni (approved):
Thanks!!!

PR #263, opened 2024-01-09
Embeddings/MLX sentence transformers

andersonbcdefg:
Create a sentence-transformers like BERT model that can run MTEB with any BERT-based text embedding model. Direct integration with HuggingFace with no on-disk conversion step. Tests on bge-small to confirm ~similar scores to running with SentenceTransformers. Works with quantization.

Doesn't seem to be numerically identical to the HF model, but MTEB scores are similar so I think it's working.


andersonbcdefg:
I might have messed this up github-wise because it's showing all the changes you made to my previous speculative decoding example. LMK if this is a problem and I can re-try this PR. But the only things that should be changed are my 2 new files, plus the fix to the BERT pooler in the BERT example bc that was wrong.

embeddings/mlx_embeddings.py
```diff
@@ -0,0 +1,205 @@
...
+    key = key.replace(".self.query.", ".query_proj.")
+    key = key.replace(".self.value.", ".value_proj.")
+    key = key.replace(".attention.output.dense.", ".attention.out_proj.")
+    key = key.replace(".attention.output.LayerNorm.", ".ln1.")
+    key = key.replace(".output.LayerNorm.", ".ln2.")
+    key = key.replace(".intermediate.dense.", ".linear1.")
+    key = key.replace(".output.dense.", ".linear2.")
+    key = key.replace(".LayerNorm.", ".norm.")
+    key = key.replace("pooler.dense.", "pooler.")
+    return key
```
awni, line 28: Could you rename the modules in the individual layers to avoid this key mapping? (I know some other examples e.g. BERT do this, but I've been trying to move everything to use the same structure as Transformers as it really simplifies things.

embeddings/mlx_embeddings.py
```diff
@@ -0,0 +1,205 @@
+# Based on the MLX BERT implementation from mlx-examples, but
```
awni, line 1: Let's just call this file "embeddings.py"

embeddings/mlx_embeddings.py
```diff
@@ -0,0 +1,205 @@
...
+            print("Keeping in fp32 precision")
+        else:
+            print(f"Quantizing to {precision_nbits} bits")
+            nn.QuantizedLinear.quantize_module(mlx_model, bits=precision_nbits)
+        tokenizer = BertTokenizer.from_pretrained(model_path)
+
+        return mlx_model, tokenizer
+
+
+class MLXEmbeddingModel:
```
awni, line 139: Let's just call this "EmbeddingModel" for simplicity.

awni (commented):
This is awesome!

What do you think about naming the example "sentence_embeddings" instead? I think that will be a more helpful descriptive name.

Also are you open to adding a README for usage info + some pointers to MTEB?

andersonbcdefg:
Yep, happy to address all the points you've raised! :D

embeddings/mlx_embeddings.py
```diff
@@ -0,0 +1,205 @@
...
+        if attention_mask is not None:
+            # convert 0's to -infs, 1's to 0's, and make it broadcastable
+            attention_mask = mx.log(attention_mask)
+            attention_mask = mx.expand_dims(attention_mask, (1, 2))
+
+        y = self.encoder(x, attention_mask) # shape: B, L, D
+        return y, mx.tanh(self.pooler(y[:, 0])) # shape: B, D
+    
+    @classmethod
+    def from_hugging_face(cls, model_path: str, precision_nbits: int = 8):
```
Vaibhavs10, line 115: Quick comment: Can this just be `from_pretrained`? It sounds more in sync with HF lingo and would also feel more familiar to people.

Vaibhavs10 (commented):
Very cool that this PR should add support for roughly 1500 public checkpoints on the Hub! 🚀

awni:
> For renaming the modules so that keys match, how would you suggest handling cases where the Transformers BERT model has more modular/nested modules? e.g. separate BERTIntermediate and BERTOutput layers? I don't think I can just stick periods into the names of modules, so it would be a choice between creating more modules than is strictly needed, vs. having some names that don't map over. I'll think about this too just curious your thoughts.

@andersonbcdefg sorry for the delayed reply, I didn't notice your question.

You are right,  you can't just put periods in the names. If avoiding the name mapping means a proliferation of silly little classes then I would just do the name mapping. E.g. I wouldn't bend over backwards to avoid remapping names. 

Other than that, how is this PR coming along? Let me know if I can help at all, I'm excited to get it in!!

andersonbcdefg:
Hey, just made some updates responding to the comments here. I changed the setup of BERT so that fewer keys have to be swapped (still some, but fewer.) :) Also changed names of stuff as requested.

awni:
Great! We also need a readme. I can help with that just let me know you're plan / when I should review.

andersonbcdefg:
Ok, ready for review!

andersonbcdefg:
Hey @awni when do you think you'll get around to this one?

awni:
Hey! I will take a look shortly (next 1-2 days), sorry for the delay!

embeddings/embeddings.py
```diff
@@ -0,0 +1,242 @@
...
+                result = pooled
+            else:
+                result = np.concatenate([result, pooled], axis=0)
+            if pbar is not None:
+                pbar.update(len(batch))
+
+        return result
+
+def test_embeddings():
+    mx_model, tokenizer = Bert.from_hugging_face("BAAI/bge-small-en-v1.5", precision_nbits=16)
```
sciencecw, line 224: Is this implemented

andersonbcdefg: Good catch on the renamed thing! I might need help to get the 16-bits version working. I found that when I naively cast all the weights of the model to 16 bits upon loading it, performance plummeted, which is why I only made 32 bit or quantized an option (the quantization utils didn't support 16 bit "quantization" because normally you don't need special scales etc. for 16 bits).

sciencecw: When using 32 bits there's still a small difference between sentence-transformers and the MLX version, ~2% in cosine similarity. Is this expected?

andersonbcdefg: I'm not sure why that is. All I verify in my testing is that MTEB performance is preserved on a subset of tasks.

awni:
@andersonbcdefg sorry for the delay. I rebased this and ran the formatting. I'm doing a little work on it now.

Just curious, what were the results you were getting? For the MLX model in 32-bit I'm seeing:

```
{'Banking77Classification': 0.807987012987013, 'STS12': 0.7597192919985193}
```

andersonbcdefg:
Lower precision results were decent/usable I thought! But it's not super important to support given how small these models are; the overhead of quantizing/dequantizing probably isn't necessary here. I'll let you make the call!

awni:
Cool, what about F32, it's about 1% worse than the torch version. Did you see the same? I can spend a little time investigating, but I also want to make sure I am not doing anything wrong on my end.

andersonbcdefg:
See if it looks better on your end with normalize=False. It doesn't affect metrics based on cosine similarity, but does seem to make a difference for e.g. BankingClassification.

awni:
I think it helped, now I see:

```
{'Banking77Classification': 0.8325974025974027, 'STS12': 0.7584972019004673}
```

The STS12 is still a bit worse than the torch model.. but the banking classification is better..

andersonbcdefg:
Yup, that's what I was getting too. Not sure what explains the remaining discrepancy.

awni:
@andersonbcdefg could you comment a bit on what this example adds beyond the original MLX Bert example? Is it mostly the MTEB evaluation? If so, maybe the right call is to modify the MLX Bert example and add the sample evaluation script? What do you think?

andersonbcdefg:
Yeah, my thought is that it makes sentence embeddings usable in MLX, which
requires the pooling, handling batches, truncation, etc. and also makes it
easy to load the models. This would be ok as a BERT wrapper + evaluation
script too.

sciencecw:
I support embedding being a separate example. `sentence-transformers` models are hosted on huggingface hub but the info about the pooling layers are not stored in the .bin file

Ideally, instead of specifying pooling strategy in the constructor, the pooling layers should be constructed by reading config.json and the layer folders (1_Pooling, 2_Dense, 3_Normalize ... etc)

sweetcard:
The tokenizer by pytorch has poor performance and uses too much memory.
Replacing it with mlx will be good news.😄

andersonbcdefg:
Well, while this is stuck I'm working on a separate repo for embeddings in MLX.

awni:
@andersonbcdefg sorry I got kind of stuck on this myself w.r.t. to how it should integrate with our BERT example.

Maybe the answer is that it shouldn't..but then it doesn't add a ton of MLX-specific functionality beyond that. The pooling into a single embedding and evals on MTEB are very nicely done and I can very much see how they are useful, but I'm not sure it's unique enough to merit a standalone example here.

Maybe it makes more sense as a separate community built repo? That would allow you to move faster adding new features and decouple it from the constraints of MLX Examples. If you are going in that route, we should include it in our featured projects and I'd be happy to point people to it! 

Let me know what you think!

andersonbcdefg:
works for me! i'll ping you when it's ready to share

awni:
Awesome can't wait!

PR #265, opened 2024-01-09
Fix use of undefined args in generate function in phi-2 example

alwint3r:
Currently, the `generate.py` accidentally works because `generate` is called in `main` where `args` is within the scope of the main function.

### DIFF
--- a/llms/phi2/generate.py
+++ b/llms/phi2/generate.py
@@ -16,7 +16,7 @@ def generate(
     temp: float = 0.0,
 ):
     print("[INFO] Generating with Phi-2...", flush=True)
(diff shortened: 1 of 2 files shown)


mzbac:
Nice catch, I manually tested the changes on my local machine. It looks good.

awni (approved):
Thanks!! We had the same bug in the `hf_llm` example, should have caught it here 🤦

alwint3r:
I noticed it too. I can file another PR if that's okay with you.

awni:
I'm not sure what you are referring to but if you see a bug, please definitely send a PR!

PR #270, opened 2024-01-09
chore(lora): remove quantization config for fuse the quant model

mzbac:
For the fuse qlora adapter weight, the quantization configuration will cause the fused model to fail to load. I haven't tested it locally yet, but this is the issue I had in the past when playing around with merging Lora.

### DIFF
--- a/lora/fuse.py
+++ b/lora/fuse.py
@@ -66,6 +66,10 @@ if __name__ == "__main__":
 
     model.update_modules(tree_unflatten(fused_linears))
     weights = dict(tree_flatten(model.parameters()))
+
+    if quantization := config.get("quantization", None):
+        config.pop("quantization")
+
     utils.save_model(args.save_path, weights, tokenizer._tokenizer, config)
 
     if args.upload_name is not None:


lora/fuse.py
```diff
@@ -66,6 +66,10 @@
 
     model.update_modules(tree_unflatten(fused_linears))
     weights = dict(tree_flatten(model.parameters()))
+
+    if quantization := config.get("quantization", None):
+        config.pop("quantization")
+
```
awni, line 72: Maybe I miss something, but I don't think that's correct... we requantize the model after it's fused?

https://github.com/ml-explore/mlx-examples/blob/main/lora/models.py#L120-L125

mzbac: @awni  You are right. Sorry, I didn't notice that.

mzbac: @awni Any thoughts on making the requantize linear optional, so people can fuse models in fp16 and requantize them using other advanced quantization methods such as GPTQ, AWQ, GGUF?

awni: Yes that would be totally fine... you probably are aware but that would also require dequantizing the rest of the model (not just the LoRA layers).

mzbac: Yeah, it's a bit tricky for non full layers lora..

PR #271, opened 2024-01-10
fix(lora): tokenizer return incompatible mx array

mzbac:
The custom tokenizer returns an mx array, and it will throw an error at https://github.com/ml-explore/mlx-examples/blob/main/lora/lora.py#L191 when trying to assign the mx array to the npx array.  also, the transformer tokenizer defaults include special tokens (bos, eos). According to https://huggingface.co/docs/transformers/main_classes/tokenizer#transformers.PreTrainedTokenizer.__call__.add_special_tokens, we don't really need the custom tokenizer for adding eos.

### DIFF
--- a/lora/README.md
+++ b/lora/README.md
@@ -162,7 +162,7 @@ useful for the sake of attribution and model versioning.
 For example, to fuse and upload a model derived from Mistral-7B-v0.1, run: 
 
 ```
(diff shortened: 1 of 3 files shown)


awni (approved):
This is a nice simplification, thank you

PR #272, opened 2024-01-10
lora.py - Incorrect assignment when padding zeros

gavi:
When zeros are padded to training samples assignment is being made between numpy array and mlx array. This causes an exception. This PR simply corrects it.


awni:
Thanks @gavi ! This was just fixed in #271 so I will close this. Sorry for breaking that.

PR #274, opened 2024-01-10
Update txt2image.py

metacritical:
mx not np.


stable_diffusion/txt2image.py
```diff
@@ -52,5 +52,5 @@
     x = (x * 255).astype(mx.uint8)
 
     # Save them to disc
-    im = Image.fromarray(np.array(x))
```
awni, line 55: I think we want to make it a numpy array as `Image.fromarray` doesn't work with `mx.array`? (Maybe you are using an old version of `mlx`?

PR #276, opened 2024-01-10
Stable Diffusion: Input image downsampling

### LINKED ISSUE #267: Img2Img in mlx-examples/stable_diffusion fails with reshape mismatch
I have only just started exploring all this wonderful MLX stuff but when I try to run the `image2image.py` script on a simple `.png` I am getting an error that seems to arise during the `unet.py` reshaping of `x` process using `[B, H, W, C]`:
```
ValueError: All the input array dimensions must match exactly except for the concatenation axis. However, the provided shapes are (8,18,18,1280), (8,17,17,1280), and the concatenation axis is -1
```
Somehow we go from `[8,17,17,1280]` to `[8,18,18,1280]` during `def __call__()` but I can't see an obvious reason or fix. Since I am only using one `[528, 528, 3]` image I don't think I can have caused this, but anything is possible! My bad if so.

Incidentally, the `im2im.png` in the repo can't be read as `original.py` as in the sample prompt because of the addition of the other fireplace images.

LeonEricsson:
closes #267 

U-Net requires that it's input is evenly divisible by the number of down / up blocks used as to guarantee that the skip connections can be added properly (avoiding shape mismatch). This is a well documented limitation [1](https://stackoverflow.com/questions/66028743/how-to-handle-odd-resolutions-in-unet-architecture-pytorch) [2](https://github.com/CompVis/stable-diffusion/issues/268) [3](https://github.com/CompVis/stable-diffusion/issues/191) of both UNet and SD. 

One solution is to just throw an error and let the user resize their input, another is to resize it for them. I followed the reasoning in [HuggingFace diffusers](https://github.com/huggingface/diffusers/blob/3be7c96e28495d2e45af5a5f8488a9bb41d66ccb/src/diffusers/image_processor.py#L787) which downsamples the input image using nearest neighbor interpolation. I wrote the downsampling function myself as I was uncertain if we wanted to include external libraries. Given that Stable Diffusion already uses Pillow one alternative is to use Pillow's own resize. This would also give the user more options when it comes to what kind of interpolation algorithm they want to use.  

Running SD on my own computer is very slow, I've made sure that the downsampling works but if someone could confirm SD output that would be great.

### DIFF
--- a/stable_diffusion/README.md
+++ b/stable_diffusion/README.md
@@ -67,7 +67,7 @@ Image 2 Image
 There is also the option of generating images based on another image using the
 example script `image2image.py`. To do that an image is first encoded using the
 autoencoder to get its latent representation and then noise is added according
-to the forward diffusion process and the `strength` parameter. A `stregnth` of
+to the forward diffusion process and the `strength` parameter. A `strength` of
 0.0 means no noise and a `strength` of 1.0 means starting from completely
 random noise.
 
@@ -78,6 +78,7 @@ The command to generate the above images is:
 
     python image2image.py --strength 0.5 original.png 'A lit fireplace'
 
+*Note: `image2image.py` will automatically downsample your input image to guarantee that its dimensions are divisible by 64. If you want full control of this process, resize your image prior to using the script.*
 
 Performance
 -----------
--- a/stable_diffusion/image2image.py
+++ b/stable_diffusion/image2image.py
@@ -27,7 +27,7 @@ if __name__ == "__main__":
 
     sd = StableDiffusion()
 
-    # Read the image    
+    # Read the image
     img = Image.open(args.image)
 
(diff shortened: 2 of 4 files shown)


angeloskath (commented):
This looks great, thanks! But out of curiosity why not throw in the `generate_latents_*` and resize the image in the script using PIL (which we use to read the image anyway)?

LeonEricsson:
> This looks great, thanks! But out of curiosity why not throw in the `generate_latents_*` and resize the image in the script using PIL (which we use to read the image anyway)?

yeah I realised after the entire implementation that PIL was already being used, I was working of off the assumption that I shouldn't import stuff unless absolutely necessary. Sorry not quite sure what you mean by _why not throw in the generate_latents_*_

angeloskath (approved):
Looks great!

PR #278, opened 2024-01-10
Fix Tokenizer save error.

Vaibhavs10:
Currently when one runs `python fuse.py` (assuming you have a `mlx_model` folder and a `adapters.npz` file.
It throws an error:

```
AttributeError: 'tokenizers.Tokenizer' object has no attribute 'save_pretrained'. Did you mean: 'from_pretrained'?
```

This PR fixes that.


Vaibhavs10:
You can test this by running `python fuse.py`

awni (approved):
Sorry for breaking that! Thanks @Vaibhavs10 !

PR #284, opened 2024-01-10
Fix use of undefiend args variable inside generate function in hf_llm example

alwint3r:
Similar to the one in [this PR](https://github.com/ml-explore/mlx-examples/pull/265).
The `generate` function uses `args` defined in the main function instead of using its function parameters.


awni (approved):
👍 thanks!!

PR #290, opened 2024-01-11
Phixtral

marcelbischoff:
Adding support for MoE models based on Phi-2 aka Phixtral 
- https://huggingface.co/mlabonne/phixtral-4x2_8
- https://huggingface.co/mlabonne/phixtral-2x2_8
This is an almost verbatim copy of the Phi-2 code and Mixtral MoE block

```sh
python llms/phixtral/generate.py --prompt "def fib(n: int) -> int:" \
--model mlabonne/phixtral-2x2_8
```

### DIFF
--- a/llms/README.md
+++ b/llms/README.md
@@ -84,7 +84,7 @@ You can upload new models to Hugging Face by specifying `--upload-repo` to
 python -m mlx_lm.convert \
     --hf-path mistralai/Mistral-7B-v0.1 \
     -q \
-    --upload-repo mlx-community/my-4bit-mistral \
+    --upload-repo mlx-community/my-4bit-mistral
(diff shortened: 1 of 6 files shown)


awni:
Thanks for moving it. No rush, but you forgot to include the `phixtral.py` file 😄

awni:
@marcelbischoff any reason this is draft? Should I go ahead and review it? (Also would be good to add a README, I can help with that if you like, lmk).

marcelbischoff:
Hey, was mostly a draft because of missing README.

marcelbischoff:
Should be ready to review

awni (approved):
Very nice, works well for me :)!!

PR #291, opened 2024-01-11
prompt parameter

anjor:
The prompt parameter was unused in the generate function. The function should use the passed in parameter instead of accessing `args.prompt` directly.


awni (approved):
Thanks!

PR #292, opened 2024-01-11
fix(hf_llm): undefined hf_path

mzbac:
I haven't tested it locally, I just quickly looked at the code. It seems that after introducing the local flag, hf_path becomes undefined.


awni (approved):
🤦 sorry for not catching that and thanks for the fix!

PR #293, opened 2024-01-11
refactor(hf_llm): moving phi2 example into hf_llm

mzbac:
Moving the phi2 example into hf_llm, starting to unify the hf model examples. This will enable us later to create a separate package for hf_llm so that people can import it into their projects instead of copying and pasting the example from the repository.

@awni Would you please help review the PRI? I have pretty much completed the implementation and tested it locally for generating and converting (not uploading, but I haven't made any changes in that part so it shouldn't break). However, I am keeping the PR as a draft to reduce noise. In case there are some back-and-forth changes.


llms/hf_llm/convert.py
```diff
@@ -106,63 +170,11 @@ def upload_to_hub(path: str, name: str, hf_path: str):
...
-        action="store_true",
-        help="Whether the hf-path points to a local filesystem.",
-        default=False,
-    )
-
+    parser = configure_parser()
     args = parser.parse_args()
 
     print("[INFO] Loading")
-    weights, config, tokenizer = fetch_from_hub(args.hf_path, args.local)
```
mzbac, line 165: The local flag is not really needed. We can check if the path exists or not to determine if it's a local path.

llms/hf_llm/utils.py
```diff
@@ -0,0 +1,146 @@
...
+
+    Returns:
+        Path: The path to the model.
+    """
+    model_path = Path(path_or_hf_repo)
+    if not model_path.exists():
+        model_path = Path(
+            snapshot_download(
+                repo_id=path_or_hf_repo,
+                allow_patterns=["*.json", "*.safetensors", "*.py", "tokenizer.model"],
```
mzbac, line 59: Include the `*.py` file to download patterns. In case a model has a custom config mapping, it will throw an error at `autoConfig.from_pretrained(model_path)`. user will still be prompted whether they want to trust remote code when loading the model/config from_pretrained.

llms/hf_llm/convert.py
awni: You're right, that's much nicer!

awni:
This looks great really nice! I will review it today. Also thanks for adding the docstrings that is a big improvement!

llms/hf_llm/models/phi2.py
```diff
@@ -1,48 +1,32 @@
...
 
 @dataclass
-class ModelArgs:
-    max_sequence_length: int = 2048
-    num_vocab: int = 51200
-    model_dim: int = 2560
-    num_heads: int = 32
-    num_layers: int = 32
+class ModelArgs(BaseModelArgs):
+    n_positions: int = 2048
```
mzbac, line 12: Update the argument name to match the configuration name in config.json.

llms/hf_llm/convert.py
```diff
@@ -1,52 +1,97 @@
...
 import mlx.core as mx
 import mlx.nn as nn
 import transformers
-from huggingface_hub import snapshot_download
 from mlx.utils import tree_flatten
-from models import Model, ModelArgs
+from utils import get_model_path, load
+
+MAX_FILE_SIZE_GIBIBYTE_DEFAULT = 15
+REPO_PREFIX = "mlx-community"
```
Vaibhavs10, line 15: Maybe we relax this to allow people to upload under there namespace too?

awni:
@mzbac I pushed some changes. The Phi2 stuff was broken I think because safetensors is in a different format.

awni (approved):
@mzbac I pushed some changes directly:

- There were a couple bugs with phi2 / llama, generation works now
- Updated README with a couple extra pointers 
- Use `model_type` instead of the `architecture` field. @Vaibhavs10 and @pcuenca confirmed this is a better approach.
- Upload to any repo instead of hardcoding `mlx_community`

I'm going to go ahead and merge it and start working on packaging it.

Thanks for this amazing change it's really nicely done! And continues to simplify the proliferation of LLM examples which I am really happy about!

mzbac:
> @mzbac I pushed some changes directly:
> 
> * There were a couple bugs with phi2 / llama, generation works now
> * Updated README with a couple extra pointers
> * Use `model_type` instead of the `architecture` field. @Vaibhavs10 and @pcuenca confirmed this is a better approach.
> * Upload to any repo instead of hardcoding `mlx_community`
> 
> I'm going to go ahead and merge it and start working on packaging it.
> 
> Thanks for this amazing change it's really nicely done! And continues to simplify the proliferation of LLM examples which I am really happy about!

I think the phi2 hf repo was just updated 5 hours ago, maybe that broke the example, Thanks for helping to review and fix the bugs for the PR.

PR #294, opened 2024-01-11
fix: Change hf_path to model_path

katopz:


awni:
Thanks! I just merge #292 which has the same fix 😄 !

PR #301, opened 2024-01-12
Mlx llm package

### LINKED ISSUE #220: Can we make hf_llm as a Python package?
First of all, thank you for making it available to us. I would love to use it for my project, but currently I have to manually copy the source code and put it in the project. I'm wondering if it's possible to make it a Python package so that people can simply import and use it.

awni:
- Rename `hf_llm` to `mlx_lm`
- Setup.py for python package
- Slight refactor to make API more friendly

Closes #220


mzbac:
The phi2 modelArgs may need to be updated to match the updated config.json. I am happy to work on that in a separate PR. Maybe also make some updates for supporting phi1 since they share the model_type "phi".

mzbac:
I played around a bit trying to bring the qwen model to mlx-lm, noticed there are so many different variants of the tokenizer, some requiring trusted remote code or needing to specify the eos_token. Maybe we should remove the tokenizer from mlx-llm or let the user instantiate it themselves. For example: https://github.com/ml-explore/mlx-examples/blob/main/llms/qwen/qwen.py#L201-L203
Any thoughts? @awni

angeloskath (approved):
👍

awni:
@mzbac I am merging as is. We can do some follow up on the tokenizer.

> The phi2 modelArgs may need to be updated to match the updated config.json.

Also not sure what you mean here? The phi2 model ran ok for me.

mzbac:
> Interesting point @mzbac. Do you think it's ok to do both 🤔? It's nice to have the tokenizer loaded in common / simple cases. It makes it super easy to do use the package.
> 
> But indeed one may need to customize the tokenizer in which case they can easily do the `Autotokenizer` bit themselves?
> 
> Note the `generate` function takes a model and a tokenizer so it's already nicely decoupled in that sense.

Yeah, the current load function will throw an error when loading a custom autokenizer with trust remote code disabled. Perhaps a better approach would be to add a parameter "trust_remote_code" in the load function. Once users get the tokenizer instance, they can add a special tokenizer or do any configuration of the tokenzier before passing the tokenizer to the generate function. I will have a look at the Autotokenizer source code to see how flexible it is for configuring the tokenizer after initialization.

mzbac:
> @mzbac I am merging as is. We can do some follow up on the tokenizer.
> 
> > The phi2 modelArgs may need to be updated to match the updated config.json.
> 
> Also not sure what you mean here? The phi2 model ran ok for me.

Yeah, currently it works because we have all the default values for the mode arguments. However, in the current model args, it doesn't read the settings from the config.json file. Fixing that would make the model more flexible in case there are some further variants of the series, I will create a PR to update it :)

PR #303, opened 2024-01-12
Add PLaMo-13B model as an LLM example

mitmul:
Thank you for developing this amazing library, MLX. This PR adds an example of operating a Llama 2-based model, named PLaMo using MLX, which is trained on both Japanese and English languages in a balanced manner. PLaMo is open source, and its weights are distributed on Hugging Face. This example includes scripts for downloading and converting the weights of PLaMo, and for generating text using the converted model. Please see the detail of PLaMo here: https://tech.preferred.jp/en/blog/llm-plamo/

This example is created in a directory separate from `llms/hf_llm` and `llms/llama` because PLaMo includes some differences from Llama implementation in several aspects. For example, replacing the rotary positional encoding used in PLaMo with `nn.RoPE` implemented in MLX did not work properly, so I had to implement it with more primitive operators (it implies something is different from the original RoPE also in the original PLaMo code). Additionally, the implementation of grouped query attention was slightly different from what is present in `llms/llama/llama.py`, and PLaMo adopted the "parallel layers" technique in its transformer block, so the architecture itself is slightly different from the original Llama 2. It was also necessary to modify the names of variables within the model to directly load the weights distributed on huggingface.

Due to these reasons, although much of the code is common with code in `llms/llama`, I created new scripts for this model based on the original implementation: [modeling_plamo.py](https://huggingface.co/pfnet/plamo-13b/blob/main/modeling_plamo.py) in a separate directory at `llms/plamo`.

I would appreciate it if you could kindly review this.


awni:
This is cool! Thanks for the contribution! At this point I think it makes more sense to start putting new models in `mlx-lm` rather than as standalone models. Would you be open to moving it there?

mitmul:
Sure! Just as I started working on this PR, the introduction of mlx-lm coincided in timing, so I adopted a strategy of separating it into a different directory. However, now it seems more appropriate to include it under mlx-lm. I will make such changes to this PR as well.

awni:
Thank you!! Let me know when it's ready for review!

mitmul:
@awni I have updated it to a state ready for review. Could you please take the time to review this PR? I would greatly appreciate it.

llms/mlx_lm/utils.py
```diff
@@ -188,5 +190,9 @@ def load(path_or_hf_repo: str) -> Tuple[nn.Module, PreTrainedTokenizer]:
     model.load_weights(list(weights.items()))
 
     mx.eval(model.parameters())
-    tokenizer = AutoTokenizer.from_pretrained(model_path)
+    if model.__class__ is plamo.Model:
+        trust_remote_code = resolve_trust_remote_code(None, config["tokenizer_class"], False, True)
```
awni, line 194: @mzbac what do you think about this as a stopgap for the qwen model also?

llms/mlx_lm/generate.py
```diff
@@ -43,7 +44,10 @@ def main(args):
     model, tokenizer = load(args.model)
     print("=" * 10)
     print("Prompt:", args.prompt)
-    prompt = tokenizer.encode(args.prompt)
+    if model.__class__ is plamo.Model:
+        prompt = tokenizer.encode(args.prompt, add_special_tokens=False)
```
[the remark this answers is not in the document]
awni: I'm confused by this. For the prompt we should never be adding an EOS token in any case right?

llms/mlx_lm/utils.py
mzbac: My understanding is the new version of Transformer will automatically prompt the user if they try to load a custom tokenizer, so we don't have to manually call `resolve_trust_remote_code()`. Additionally, I am concerned that people may use the mlx-lm package not only as a CLI tool but also integrate it into their service. The terminal prompt may not be visible to them. Personally, I feel it would be better to provide an option in the load function where users can explicitly enable remote code and pass specific token configuration if they want.

awni: That's a good point re using it not as a CLI. So maybe we should go with a config option in the load function? That sounds pretty simple to me?

llms/mlx_lm/convert.py
```diff
@@ -69,7 +70,8 @@ def fetch_from_hub(
         weights.update(mx.load(wf).items())
 
     config = transformers.AutoConfig.from_pretrained(model_path)
-    tokenizer = transformers.AutoTokenizer.from_pretrained(model_path)
+    trust_remote_code = resolve_trust_remote_code(None, config.tokenizer_class, False, True)
```
mzbac, line 73: I think the new version of the transformer should fix the issue of not prompting the user to load a custom tokenizer, so we don't have to manually call `resolve_trust_remote_code`. Would you try installing the latest transforms as suggested in this link and see if it works without resolve_trust_remote_code? [Link](https://github.com/huggingface/transformers/issues/28501#issuecomment-1892237713)

llms/mlx_lm/utils.py
mzbac: Yeah, something like https://github.com/ml-explore/mlx-examples/blob/25404dd0008aafea2aebce68fafe70238ef2ff66/llms/mlx_lm/utils.py#L188-L211 but maybe using an explicit option as `tokenizer_config` instead of kwargs.

awni: Yea I think that's very reasonable. That will fix the use as an API and the new transformers will make it work nicely with the CLI.

Are you interested in adding it to your qwen PR and then we can rebase this one on that?

mzbac: Sure thing, I just double checked and the new version of transformers hasn't been released yet. We either have to wait until the release or use transformers's master branch in requirement.txt. Do you have any preference? @awni

awni: I think the release should be out any day now, they are due for a release, so let's plan to wait.

awni: (keeps it simple from a requirements perspective)

awni:
This looks great, and thanks for moving it to MLX LM.

Just to recap @mitmul, @mzbac will add the tokenizer config option in #312 to support the Qwen model in MLX LM after the new Transformers package is out.

After that you can rebase this on #312 after its merged and then we can land this as a quick follow up!

Let me know if I missed anything! Will be great to have these two new models, especially excited about the multilingual support 👍

llms/mlx_lm/generate.py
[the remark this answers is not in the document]
mitmul: Sorry, this issue is not due to the difference in versions of the transformers library, but rather due to the difference in the `add_eos_token` setting in the `tokenizer_config.json` distributed with each of plamo-13b and plamo-13b-instruct[-nc]. In the `tokenizer_config.json` of plamo-13b-instruct[-nc], this setting is `true`, which seems to cause the EOS to be automatically added during encoding. This significantly degrades the performance of the chat response, so I am thinking of reporting this issue to the authors of PLaMo. I believe this is not something to be handled in this `generate.py`, so I will remove this part.

llms/mlx_lm/convert.py
mitmul: Thank you for pointing to the instruction, and I tried with the latest transformers library installed from github repo. Then, I confirmed that I can remove `resolve_trust_remote_code` from this code. I applied the modification in 6828f51, so I'd appreciate it if you would check this part again.

llms/mlx_lm/generate.py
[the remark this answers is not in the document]
mitmul: I removed this part in https://github.com/ml-explore/mlx-examples/pull/303/commits/ff63bc71661e03f9d527290ace35fa0327342f0f.

llms/mlx_lm/generate.py
```diff
@@ -3,6 +3,7 @@
 
 import mlx.core as mx
 
+from .models import plamo
```
mzbac, line 6: nit: unused import :)

llms/mlx_lm/utils.py
```diff
@@ -8,9 +8,10 @@
 import mlx.nn as nn
 from huggingface_hub import snapshot_download
 from transformers import AutoTokenizer, PreTrainedTokenizer
+from transformers.dynamic_module_utils import resolve_trust_remote_code
```
mzbac, line 11: nit: unused import: )

mitmul: Thanks! removed in this commit: https://github.com/ml-explore/mlx-examples/pull/303/commits/d213195ee8c55861a904d92256461dea6aad5abf

llms/mlx_lm/generate.py
mitmul: Thanks! removed in this commit: https://github.com/ml-explore/mlx-examples/pull/303/commits/d213195ee8c55861a904d92256461dea6aad5abf

awni:
Hi! #312 has landed! Would you mind rebasing this on main and then we can merge it also?

mitmul:
> Hi! #312 has landed! Would you mind rebasing this on main and then we can merge it also?

Thank you for letting me know it, @awni . I merged the latest main and this PR became a simple one that almost just adds PLaMo model code and update README. I'd appreciate it if you would check this PR again.

llms/README.md
```diff
@@ -111,6 +111,17 @@ Here are a few examples of Hugging Face models that work with this example:
 - [microsoft/phi-2](https://huggingface.co/microsoft/phi-2)
 - [mistralai/Mixtral-8x7B-Instruct-v0.1](https://huggingface.co/mistralai/Mixtral-8x7B-Instruct-v0.1)
 - [Qwen/Qwen-7B](https://huggingface.co/Qwen/Qwen-7B)
+- [pfnet/plamo-13b](https://huggingface.co/pfnet/plamo-13b)
+- [pfnet/plamo-13b-instruct](https://huggingface.co/pfnet/plamo-13b-instruct)
+    - Please note that the prompt must be formatted according to the format written in the Usage section of the follwoing page: [Usage](https://huggingface.co/pfnet/plamo-13b-instruct#usage)
```
awni, line 116: I'm wondering if this is not relevant anymore now that we default to use the Hugging Face tokenizer chat template? (See #338 )

Also it looks like the comment about Qwn needing the special EOS and trust remote code option was dropped in the rebase. Could you re-add that?

mitmul: @awni Thank you for pointing out it. I removed the unnecessary notes from README in: https://github.com/ml-explore/mlx-examples/pull/303/commits/0381d15da6df3b09fe26aeadb7023197b0cf5b10

Well, I think the description about the special EOS for Qwen model still remains here: https://github.com/ml-explore/mlx-examples/blob/0381d15da6df3b09fe26aeadb7023197b0cf5b10/llms/README.md?plain=1#L126-L138

Could you check this update again?

awni: My mistake, thanks for the update!

awni (approved):
Thank you for the contribution, this is great!!

PR #306, opened 2024-01-12
Fix TypeError in whisper benchmark script

aboucaud:
Executing `python benchmark.py` from the `whisper` directory yields
```tb
TypeError: DecodingOptions.__init__() got an unexpected keyword argument 'model_path'
```

This PR aims at fixing the issue by adding the appropriate placeholder for `model_path` in the `DecodingOptions` dataclass.

### DIFF
--- a/whisper/benchmark.py
+++ b/whisper/benchmark.py
@@ -66,7 +66,7 @@ def decode(model, mels):
 
 
 def everything(model_path):
-    return transcribe(audio_file, model_path=model_path)
+    return transcribe(audio_file, path_or_hf_repo=model_path)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


awni (changes requested):
Thanks for catching this.

Actually I think the right fix is to change the references of `model_path` in `transcribe` and `load_model` to be `hf_path_or_repo`

As here https://github.com/ml-explore/mlx-examples/blob/main/whisper/whisper/transcribe.py#L65
And here https://github.com/ml-explore/mlx-examples/blob/main/whisper/whisper/load_models.py#L16

Here are the references to update:
https://github.com/ml-explore/mlx-examples/blob/main/whisper/benchmark.py#L69
and
https://github.com/ml-explore/mlx-examples/blob/main/whisper/benchmark.py#L131

awni (approved):
Thanks!!

PR #308, opened 2024-01-12
Change tuple type definitions to use Tuple

angeloskath:
Annotating types using `list`, `tuple`, `dict` was introduced in python 3.9 and we are supporting from 3.8 . I guess since nobody complained I may be the last person using 3.8 .

I also bumped the version for the `mlx_lm` package so it can be uploaded to PyPI.

### DIFF
--- a/bert/model.py
+++ b/bert/model.py
@@ -1,7 +1,7 @@
 import argparse
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


awni (approved):
>  I guess since nobody complained I may be the last person using 3.8 .
🤣 

Thank you

PR #310, opened 2024-01-12
Add an argument to lora.py for saving the adapter file regularly

Maverobot:
Thank you for providing such a great library! 

Currently there is a risk of losing all progress in the event of an interruption during the finetuning process (`lora.py`). 

This pull request introduces the `--save-every N` argument, allowing to save adapter weights to the disk every N iterations during the finetuning process. This addition draws inspiration from a similar argument in [finetune in llama.cpp](https://github.com/ggerganov/llama.cpp/tree/master/examples/finetune).

I would appreciate any feedback.

### DIFF
--- a/lora/lora.py
+++ b/lora/lora.py
@@ -97,6 +97,12 @@ def build_parser():
         help="Save/load path for the trained adapter weights.",
     )
     parser.add_argument(
+        "--save-every",
+        type=int,
+        default=100,
+        help="Save the model every N iterations.",
+    )
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


gladjoyhub:
Yes, fine-tuning uses a lot of computing, energy and time, it would be nice to save progress every iter.

ivanfioravanti:
Nice one! I have lost time and energy sometimes.

awni (approved):
Looks great, thanks for adding that!!

PR #311, opened 2024-01-13
chore(mlx-lm): update phi2 model args to sync with hf config format.

mzbac:
Just quickly update the model args to match config.json in Phi2's official HF repository configuration.

### DIFF
--- a/llms/README.md
+++ b/llms/README.md
@@ -61,7 +61,7 @@ text using the given prompt.
 For a full list of options run:
 
 ```
-python -m mlx_lm generate --help
+python -m mlx_lm.generate --help
 ```
 
 To quantize a model from the command line run:
--- a/llms/mlx_lm/models/phi2.py
+++ b/llms/mlx_lm/models/phi2.py
@@ -1,6 +1,5 @@
 import math
 from dataclasses import dataclass
-from typing import Tuple
 
 import mlx.core as mx
 import mlx.nn as nn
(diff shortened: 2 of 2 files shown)


llms/README.md
```diff
@@ -61,7 +61,7 @@ text using the given prompt.
 For a full list of options run:
 
 ```
-python -m mlx_lm generate --help
+python -m mlx_lm.generate --help
```
mzbac, line 64: Small typo I found while working on updating the phi2 model args.

llms/mlx_lm/models/phi2.py
```diff
@@ -10,43 +9,87 @@
...
+    intermediate_size: int = 10240
+    layer_norm_eps: float = 1e-5
+    rope_theta: float = 10000.0
+
+    def __post_init__(self):
+        if self.num_key_value_heads is None:
+            self.num_key_value_heads = self.num_attention_heads
 
 
 class LayerNorm(nn.LayerNorm):
```
mzbac, line 28: I have tried to replace the LayerNorm with the native `nn.LayerNorm`, but for some reason, not converting to fp32 during the forward pass led to a noticeable decline in inference.

awni (approved):
Thanks!

llms/README.md
awni: 👍

llms/mlx_lm/models/phi2.py
awni: Yes for any operation which does a sum (or reduction over many elements, fp16 will typically not work well). We cast up to fp32 in RMSNorm and LayerNorm (as well as softmax for the denominator).

It might be nice to make this an option for our normalization layers so we can avoid so much code duplication..

llms/mlx_lm/models/phi2.py
```diff
@@ -10,43 +10,87 @@
...
     def __call__(self, x: mx.array) -> mx.array:
         return super().__call__(x.astype(mx.float32)).astype(x.dtype)
 
 
-class RoPEAttention(nn.Module):
-    def __init__(self, dims: int, n_head: int, rotary_dim: int):
+class PhiAttention(nn.Module):
```
awni, line 34: 👍

awni:
This is very nicely done @mzbac thank you!

llms/mlx_lm/models/phi2.py
mzbac: That makes sense. Thank you for the detailed explanation.

PR #312, opened 2024-01-13
refactor(qwen): moving qwen into mlx-lm

mzbac:
Remove the conversion step for the qwen model so that it can directly load the model from HF and move it to the mlx-lm package. Also, fix a small bug in the implementation of the qwen model.


llms/mlx_lm/generate.py
```diff
@@ -21,6 +21,17 @@ def setup_arg_parser():
         default="mlx_model",
         help="The path to the local model directory or Hugging Face repo.",
     )
+    parser.add_argument(
```
mzbac, line 24: I am not very sure if this is the best implementation for supporting enabling trust remote code and specifying a special token for the tokenizer.

llms/mlx_lm/models/qwen.py
```diff
@@ -95,7 +97,7 @@ def __init__(self, args: ModelArgs):
             args.hidden_size, args.intermediate_size // 2, bias=not args.no_bias
         )
         self.w2 = nn.Linear(
-            args.intermediate_size // 2, args.hidden_size, bias=not args.no_bias
+            args.hidden_size, args.intermediate_size // 2, bias=not args.no_bias
```
mzbac, line 100: This is a bug I found in the original Qwen implementation, fixed it according to the HF repo's implementation. 
https://huggingface.co/Qwen/Qwen-1_8B-Chat/blob/main/modeling_qwen.py#L558-L568

llms/README.md
```diff
@@ -101,10 +101,14 @@ Here are a few examples of Hugging Face models that work with this example:
...
 
 Most
 [Mistral](https://huggingface.co/models?library=transformers,safetensors&other=mistral&sort=trending),
 [Llama](https://huggingface.co/models?library=transformers,safetensors&other=llama&sort=trending),
-and
 [Phi-2](https://huggingface.co/models?library=transformers,safetensors&other=phi&sort=trending)
 style models should work out of the box.
+For [Qwen](https://huggingface.co/models?library=transformers,safetensors&other=qwen&sort=trending) style models, it is required to enable the `trust_remote_code` option and specify the `eos_token`. This ensures proper functioning of the tokenizer with Qwen models. You can do this by passing `--trust-remote-code` and `--eos-token "<eos_token>"` in the command line, or by setting these options in the Python API:
+```python
+model, tokenizer = load("qwen/Qwen-7B", tokenizer_config={"eos_token": "<eos_token>", "trust_remote_code": True})
```
angeloskath, line 113: 'eos_token' needs to be `<|endoftext|>` rather than `<eos_token>`

llms/mlx_lm/utils.py
```diff
@@ -134,12 +140,14 @@ def generate(
     return tokens
 
 
-def load(path_or_hf_repo: str) -> Tuple[nn.Module, PreTrainedTokenizer]:
+def load(path_or_hf_repo: str, **kwargs) -> Tuple[nn.Module, PreTrainedTokenizer]:
```
angeloskath, line 143: Personally I would make this an explicit keyword argument called `tokenizer_config`. If they start to become a lot we can refactor to kwargs dict.

angeloskath (changes requested):
This looks good to me. The `eos_token` needs fixing but after that it is mergeable.

@awni let us know if you think that allowing people to enable `trust_remote_code` is a bad idea.

llms/README.md
mzbac: Nice catch, will update it.

llms/mlx_lm/utils.py
mzbac: Yeah, I was thinking of making it a bit flexible. If we find that the usage of kwargs is getting out of control, we can always extract it from kwargs without breaking the contract. But I am happy to go either way.

llms/mlx_lm/convert.py
mzbac, line 1: The changes in convert.py are required to load qwen's custom tokenizer. I'm not sure if it's the best way to do that, I'm open to suggestions.

llms/mlx_lm/utils.py
```diff
@@ -57,7 +57,13 @@ def get_model_path(path_or_hf_repo: str) -> Path:
         model_path = Path(
             snapshot_download(
                 repo_id=path_or_hf_repo,
-                allow_patterns=["*.json", "*.safetensors", "*.py", "tokenizer.model"],
+                allow_patterns=[
+                    "*.json",
+                    "*.safetensors",
+                    "*.py",
+                    "tokenizer.model",
+                    "*.tiktoken",
```
mzbac, line 65: Add allow tiktoken for supporting Qwen's tokenizer.

mzbac:
@awni This PR is ready for review, but I will keep it as a draft in case we need to change the way the tokenizer_config gets passed around.

llms/mlx_lm/models/qwen.py
awni: Interesting, I wonder how it was working before? Maybe we were not loading in strict mode so the weights were just getting set correctly anyway...

awni:
I'm just brainstorming a bit, not sure this is a good idea. But one way to avoid needing to configure the tokenizer in `convert.py` which is kind of annoying is:

1. Don't load the tokenizer in the `convert.py` script (would require a flag to `load`, like `load_tokenizer=False`)
2. Simply copy the relevant tokenizer files over in the conversion script

Another thought: is it sensible to read the [tokenizer class field](https://huggingface.co/Qwen/Qwen-1_8B/blob/main/config.json#L30) (if it's there) in the model specific file and try to set the config based on that? From an API standpoint it might keep things pretty clean, not sure it's flexible enough though.

@mzbac what do you think?

mzbac:
> I'm just brainstorming a bit, not sure this is a good idea. But one way to avoid needing to configure the tokenizer in `convert.py` which is kind of annoying is:
> 
> 1. Don't load the tokenizer in the `convert.py` script (would require a flag to `load`, like `load_tokenizer=False`)
> 2. Simply copy the relevant tokenizer files over in the conversion script
> 
> Another thought: is it sensible to read the [tokenizer class field](https://huggingface.co/Qwen/Qwen-1_8B/blob/main/config.json#L30) (if it's there) in the model specific file and try to set the config based on that? From an API standpoint it might keep things pretty clean, not sure it's flexible enough though.
> 
> @mzbac what do you think?

Yeah, I agree it would be a better approach to avoid loading the tokenizer in concert.py just to save it into the converted model folder. I will check and see if there is an easy way to find all the tokenizer dependencies and simply copy them over.

To read the tokenizer config from the model repository, it may not solve the issue here. We still need to let the user explicitly enable trust remote code and correct the wrong config. The reason we have to manually set the EOS token for qwen is that I think the tokenizer config wasn't right.

llms/mlx_lm/models/qwen.py
mzbac: Yeah, maybe the current example requires aconversation. so it make the mismatch dimension less obvious and example is using Chinese prompt and the performance degradation is not noticeable to most non-Chinese speakers.

llms/mlx_lm/tokenizer_utils.py
```diff
@@ -0,0 +1,66 @@
...
+
+# Constants copied from transformers library https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_utils_base.py#L128-L135
+SPECIAL_TOKENS_MAP_FILE = "special_tokens_map.json"
+ADDED_TOKENS_FILE = "added_tokens.json"
+TOKENIZER_CONFIG_FILE = "tokenizer_config.json"
+FULL_TOKENIZER_FILE = "tokenizer.json"
+RE_TOKENIZER_FILE = re.compile(r"tokenizer\.(.*)\.json")
+
+# Hardcoded mapping of model type to vocab file pattern
+VOCAB_FILE_MAPPING = {"qwen": "*.tiktoken"}
```
mzbac, line 14: If we want to avoid executing remote code, a hardcoded mapping for the vocabulary may be the best option for now. Otherwise, we may have to dynamically load the remote script and retrieve the name of the vocabulary file from https://huggingface.co/Qwen/Qwen-7B-Chat/blob/main/tokenization_qwen.py#L60.

It will be something similar to https://github.com/huggingface/transformers/blob/main/src/transformers/dynamic_module_utils.py#L380.

@awni What do you think? Personally, I feel that the behavior of the load tokenizer from_pretrained not prompting the user to allow remote code is inconsistent in transformers. I might try raising a PR there and see if they can enable prompting for allowing remote code in the tokenizer as with loading models from_pretrained. Then we would need to hardcode mapping in mlx_lm.

mzbac:
@awni We may just want the new version of the transformer released that fixed the tokenizer's trust remote code prompt, so we don't have to add any additional logic into convert.py. The user will be prompted whether they want to trust the remote code or not.
FYI, https://github.com/huggingface/transformers/issues/28501

awni:
> We may just want the new version of the transformer released that fixed the tokenizer's trust remote code prompt, so we don't have to add any additional logic into convert.py. The user will be prompted whether they want to trust the remote code or not.

Yes! That is a nice and simple solution. I don't mind limiting `mlx_lm` to that >= the new latest Transformer release.

mzbac:
@awni The PR is ready for review and merge. Transformers 4.37.0 has just been released. For Qwen, you still need to provide the trusted remote code and EOS token in cli, otherwise, it will throw an error when checking if the token equals EOS in generate_step.
Once this is merged, https://github.com/ml-explore/mlx-examples/pull/303 will be unblocked.

awni:
@mzbac why do you need to provide trust remote code in the CLI? I thought the new Transformer will prompt the user if it's not provided?

mzbac:
> @mzbac why do you need to provide trust remote code in the CLI? I thought the new Transformer will prompt the user if it's not provided?

@awni, The Transformer will prompt the user if they don't provide trusted remote code. However, without specifying an EOS token, it will throw an error in the generation step, since the user has to manually add "--eos" in the CLI, I thought of providing a "trust remote code" option for convenience.

awni:
Makes sense! Just FYI I pushed some nits to the README on your branch. Merging now!

awni (approved):
Thanks for adding this!!

angeloskath (approved):

Issue #143, opened 2023-12-18
ViT + CLIP


gboduljak:
Would it be worth implementing ViT and CLIP example?

awni:
Yea that ones on our list of examples to add!

gboduljak:
> Yea that ones on our list of examples to add! Are you interested in contributing it? If so which model would you use?

I would like to contribute :) I would use models from the official CLIP repository: https://github.com/openai/CLIP. If you have an alternative idea, please let me know.

nkasmanoff:
@gboduljak I submitted a PR to your existing PR, which creates a local implementation of the CLIPImageProcessor. https://github.com/gboduljak/mlx-examples/pull/1

This should eliminate the dependency on transformers, aside from using it for downloading the model & tokenizer.

gboduljak:
@nkasmanoff Thanks for the help. I will take a look at your work now.

gboduljak:
@nkasmanoff I merged your PR, corrected the nits and I refactored your implementation so that everything is in ``preprocessing`` folder. Many thanks for the help. In future, we might drop this 'copy-paste' implementation from HuggingFace. Ideally, we should use [mlx-data](https://github.com/ml-explore/mlx-data). If you have time, it would be awesome to have ``mlx-data`` implementation of ``CLIPImageProcessor``.

PR #315, opened 2024-01-13
CLIP (ViT)

gboduljak:
This is an implementation of CLIP (ViT).  Closes https://github.com/ml-explore/mlx-examples/issues/143.

Implemented: 
- [x] CLIP Model + CLIP Loss
- [x] CLIP Model tests
- [x] CLIP Vision Encoder tests
- [x] CLIP Text Encoder tests
- [x] Partial implementation of `CLIPTokenizer` 
- [x] Partial implementation of `CLIPImageProcessor` 
- [x] Initialization 
- [x] ``convert.py`` 
- [x] README.md

This PR is ready for review. My only concern is the correctness of CLIP initialization. I am not sure how to test that.
I would like to generate `HuggingFace` weights and configuration after the review. It is already possible to generate and upload `HuggingFace` weights and configuration. However, I decided to wait because those may change if the code changes.


clip/model.py
```diff
@@ -0,0 +1,447 @@
...
+def quick_gelu(x: mx.array) -> mx.array:
+    """
+    Applies GELU approximation that is fast but somewhat inaccurate. See: https://github.com/hendrycks/GELUs
+    """
+    return x * mx.sigmoid(1.702 * x)
+
+
+def clip_loss(logits_per_text: mx.array, logits_per_image: mx.array) -> mx.array:
+    [N, _] = logits_per_text.shape
+    [M, _] = logits_per_image.shape
```
angeloskath, line 27: This feels nitpicky but I think it is slightly more readable and far more common to omit the brackets from this destructuring. Basically, `N, _ = logits_per_text.shape` or alternatively `N = len(logits_per_text)`.

clip/model.py
```diff
@@ -0,0 +1,447 @@
...
+            activation=quick_gelu,
+            norm_first=True,
+        )
+        self.attention.query_proj.bias = mx.zeros(hidden_dim)
+        self.attention.key_proj.bias = mx.zeros(hidden_dim)
+        self.attention.value_proj.bias = mx.zeros(hidden_dim)
+        self.attention.out_proj.bias = mx.zeros(hidden_dim)
+
+    def init_weights(self, factor: float, num_hidden_layers: int):
+        [embed_dim, _] = self.linear2.weight.shape
```
angeloskath, line 75: Same as above.

clip/model.py
```diff
@@ -0,0 +1,447 @@
...
+    def __call__(self, x: mx.array) -> CLIPTextOutput:
+        # Extract some shapes
+        B, N = x.shape
+        eot_tokens = mx.argmax(x, axis=-1)
+        # Look up embeddings
+        x = self._embed(x)
+        # Compute the causal mask
+        mask = nn.MultiHeadAttention.create_additive_causal_mask(N, x.dtype)
+        # Push through the transformer
+        x = reduce(lambda x, l: l(x, mask), self.layers, x)
```
angeloskath, line 141: Love it :-) . However, I will have to be a bit pedantic and annoying and ask you to change it to

```python
for l in self.layers:
    x = l(x, mask)
```

it is a bit simpler when one needs to add a breakpoint or save some intermediate value or sth. In this particular case it is also not that much more verbose.

clip/model.py
```diff
@@ -0,0 +1,447 @@
...
+            CLIPEncoderLayer(
+                config.hidden_size, config.intermediate_size, config.num_attention_heads
+            )
+            for _ in range(config.num_hidden_layers)
+        ]
+        self.post_layernorm = nn.LayerNorm(config.hidden_size)
+        self.init_weights(config.initializer_range, config.initializer_factor)
+
+    def _embed(self, x: mx.array) -> mx.array:
+        [batch_size, _, _, _] = x.shape
```
angeloskath, line 224: Same as above.

clip/model.py
```diff
@@ -0,0 +1,447 @@
...
+        embeddings += self.position_embedding
+        return embeddings
+
+    def __call__(self, x: mx.array) -> CLIPVisionOutput:
+        # Look up patch embeddings
+        x = self._embed(x)
+        # Prenorm
+        x = self.pre_layernorm(x)
+        # Push through transformer
+        x = reduce(lambda x, l: l(x, mask=None), self.layers, x)
```
angeloskath, line 247: Same as above.

clip/cats.jpeg
angeloskath, line 1: I would move the images into an `example-images` folder or something named similarly and add a README in that folder that is attributing the image creator and reporting the license on these. For instance something like the following

```
example-images/cats.jpeg by Person Who Made It CC BY 4.0
```

see https://wiki.creativecommons.org/wiki/best_practices_for_attribution for more examples.

angeloskath: If they are your images, 1) I envy you 2) add yourself

angeloskath (changes requested):
@gboduljak this looks amazing! Thanks for the contribution.

I started the review, haven't finished it yet but I am submitting the 1st pass and will get back to it again later today to review the conversion, preprocessing and hub download parts.

The current comments are mostly stylistic and image attribution. So far the PR looks great so feel free to address the review comments when the 2nd pass of review is submitted.

clip/model.py
```diff
@@ -0,0 +1,447 @@
...
+            dims=hidden_dim,
+            mlp_dims=intermediate_dim,
+            num_heads=num_heads,
+            activation=quick_gelu,
+            norm_first=True,
+        )
+        self.attention.query_proj.bias = mx.zeros(hidden_dim)
+        self.attention.key_proj.bias = mx.zeros(hidden_dim)
+        self.attention.value_proj.bias = mx.zeros(hidden_dim)
+        self.attention.out_proj.bias = mx.zeros(hidden_dim)
```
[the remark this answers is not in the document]
gboduljak: Thanks for the suggestion. I will go for it. It is much cleaner than the current one.

clip/model.py
gboduljak: I will rewrite to match your suggestion.

clip/model.py
gboduljak: I will rewrite to match your suggestion.

clip/model.py
gboduljak: I agree. The `loop` solution is much better for debugging :)

gboduljak:
> @gboduljak this looks amazing! Thanks for the contribution.
> 
> I started the review, haven't finished it yet but I am submitting the 1st pass and will get back to it again later today to review the conversion, preprocessing and hub download parts.
> 

Thank you for the code review and the constructive feedback. I will address your requested changes after your second pass.

gboduljak:
> As you mentioned one could use `mlx-data` for the preprocessing and tokenization steps but we don't currently have a `BPETokenizer` in MLX Data. Let me know if you feel like tackling that cause I think it would be very useful.

This looks interesting to me. Please take a look at ``clip/preprocessing/tokenizer.py``. This is based on the tokenizer from [Stable Diffusion example](https://github.com/ml-explore/mlx-examples/blob/main/stable_diffusion/stable_diffusion/tokenizer.py). I would to port this to `mlx-data`.

gboduljak:
> > As you mentioned one could use `mlx-data` for the preprocessing and tokenization steps but we don't currently have a `BPETokenizer` in MLX Data. Let me know if you feel like tackling that cause I think it would be very useful.
> 
> This looks interesting to me. Please take a look at `clip/preprocessing/tokenizer.py`. This is based on the tokenizer from [Stable Diffusion example](https://github.com/ml-explore/mlx-examples/blob/main/stable_diffusion/stable_diffusion/tokenizer.py). I would to port this to `mlx-data`.

@angeloskath please see https://github.com/ml-explore/mlx-data/issues/36 :)

mzbac:
@gboduljak Great work! I can't wait for this to be merged, it will enable us to port Llava-like models to mlx 🚀

gboduljak:
@angeloskath Could you take a second look at this PR? I would like to complete this soon :)

awni:
@gboduljak can you rebase on main and resolve the conflict? I can spend some time on this also as I'd like to get it in in the nearish future.

gboduljak:
> @gboduljak can you rebase on main and resolve the conflict? I can spend some time on this also as I'd like to get it in in the nearish future.

@awni done :)

clip/cats.jpeg
awni, line 1: Can you put these test files in a subdirectory like `assets/.`?

awni:
@gboduljak my high level comment is there is a lot of code in this example. Examples are intended to be pretty minimal so they can be 1. instructive 2. easily hackable/customizable, and 3. easy for us to maintain. If we want to include this in `mlx-examples` we should aim to reduce the footprint and simplify as much as possible

- If that means removing some config options to make it less flexible that is totally fine!
- If that means hardcoding some model preprocessing steps, that is totally fine!
- Reducing the number of small files would also be good.

Another path is to make this a separate package that is external to mlx-examples which is also great! And maybe makes more sense for your goals with this, I am not sure.

If we push to include it in mlx-examples though let's aim to simplify as much as possible. Let me know which route your prefer and if you want to aim to get it in mlx-examples (which I support) are you up for taking a stab at simplifying it a bit?

gboduljak:
> @gboduljak my high level comment is there is a lot of code in this example. Examples are intended to be pretty minimal so they can be 1. instructive 2. easily hackable/customizable, and 3. easy for us to maintain. If we want to include this in `mlx-examples` we should aim to reduce the footprint and simplify as much as possible
> 
> * If that means removing some config options to make it less flexible that is totally fine!
> * If that means hardcoding some model preprocessing steps, that is totally fine!
> * Reducing the number of small files would also be good.
> 
> Another path is to make this a separate package that is external to mlx-examples which is also great! And maybe makes more sense for your goals with this, I am not sure.
> 
> If we push to include it in mlx-examples though let's aim to simplify as much as possible. Let me know which route your prefer and if you want to aim to get it in mlx-examples (which I support) are you up for taking a stab at simplifying it a bit?

The largest amount of the code is preprocessing, which is not really related to CLIP implementation. My goal was to reduce dependency on `transformers`, but some things are missing in existing MLX ecosystem (e.g. BPETokenizer) which makes this PR massive. 

I suggest we remove preprocessing completely and rely on `transformers`, until we do not implement necessary features in `mlx-data`. This makes sense because this is a repository of examples. By relying on `transformers` preprocessing (i.e. tokenizer + image preprocessing), we are left with just a few files (e.g. CLIP implementation, CLIP tests and convert.py).

To save your time, please take a look at the model implementation (`model.py`, `convert.py`) and the tests (`test.py`).

awni:
> I suggest we remove preprocessing completely

I'm in favor of removing it's all in NumPy anyway there is no need to reimplement it here. I don't love having a dependency on Torch though but maybe we take it out once we have the image preprocessing in mlx-data.

- Regarding image preprocessing - just curious is the image preprocessing in Transformers in Torch or numpy? Also do you know what is missing to use mlx-data for the image parts, I thought we had most of those functions?

- Definitely tokenizations using Transformers is fine, let's do that. Once we can do BPE whatever is needed in mlx-data we can update it.

gboduljak:
> * Regarding image preprocessing - just curious is the image preprocessing in Transformers in Torch or numpy? Also do you know what is missing to use mlx-data for the image parts, I thought we had most of those functions?

I think it is mostly `numpy` and `Pillow`. I think we are missing image normalization in `mlx-data`. But that is easy to implement anyway.

Let me take a look at this later today and come up with some simplified implementation (e.g. removing preprocessing and trying to implement something basic in `mlx`). I will let you know when that is done.

awni:
Awesome thank you!!! 

The normalization is pretty easy to do in `mlx-data`. Here is an exampel: https://github.com/ml-explore/mlx-examples/blob/main/cifar/dataset.py

awni:
@gboduljak let me know when this is ready for re-review.

gboduljak:
> @gboduljak let me know when this is ready for re-review.

@awni here is the status update.

I removed preprocessing for images and I just wrapped image preprocessing from `transformers`. I would like to try `mlx-data` for image preprocessing instead of this 'patchy' solution.

The implementation of CLIP model, the weight conversion script and the tests are final and I do not intend to change them. These will not change during the image preprocessing re-implementation. Thus, please review `model.py`, `convert.py` and `test.py`.

gboduljak:
Hi @awni, @angeloskath :)

@angeloskath I addressed your first pass comments. 

This PR is now ready for another review. I removed existing preprocessing code and I implemented a very simple port of [CLIPImageProcessor](https://github.com/huggingface/transformers/blob/main/src/transformers/models/clip/image_processing_clip.py).  The 'simple port' is implemented using `Pillow` and `numpy`. This means we can drop dependency on `transformers`. I decided to keep the dependency because it is useful for tests. However, `transformers` are not necessary for inference.

I tried using `mlx-data` for the image processor. In particular, I tried 
```python

def normalize(x: mx.array, mean: mx.array, std: mx.array) -> mx.array:
    x = x.astype("float32") / 255.0
    return (x - mean) / std

mean = mx.array(conf["image_mean"]).reshape((1, 1, 3))
std = mx.array(conf["image_std"]).reshape((1, 1, 3))
norm = partial(normalize, mean=mean, std=std)

dset = (
    # Make a buffer (finite length container of samples) from the python list
    dx.buffer_from_vector([{'image': b'assets/cat.jpeg'},{'image': b'assets/dog.jpeg'}])
    # Shuffle and transform to a stream
    .to_stream()
    # CLIP image pipeline
    .load_image("image")
    .image_resize_smallest_side("image", 224)
    .image_center_crop("image", 224, 224)
    # Accumulate into batches
    .batch(2)
    # Normalize
    .key_transform("image", norm)
    # Finally, fetch batches in background threads
    .prefetch(prefetch_size=2, num_threads=2)
)
cat, dog = [sample["image"] for sample in dset]
``` 
However, I was getting outputs very different from the [reference CLIPImageProcessor](https://github.com/huggingface/transformers/blob/main/src/transformers/models/clip/image_processing_clip.py). I suspect this is because CLIPImageProcessor uses Bicubic sampler for resizing. I am not sure whether we have Bicubic sampler in `mlx-data`. If we have the Bicubic sampler  in `mlx-data`, please suggest how to use it.

chigkim:
Sorry for the digression, but thank you so much for working on this!
I've developed [VOCR](https://github.com/chigkim/vocr/releases), a tool designed to assist blind users in capturing screenshots and extracting information from images.
VOCR has been utilizing only Apple Vision Kit, but I'm currently working on V2 to incorporate multimodal models with various engines such as GPT-4V, Llama.cpp, and Ollama. I can't wait to try MLX with multimodal vision-language models!
Although still in its beta phase, users are finding VOCR helpful for a wide range of purposes. These include accessing information from inaccessible software, analyzing pictures and videos on social media, and obtaining subtitles, among others.
it would be amazing if it becomes possible in the future to finetune multimodal models on Macs with Apple silicon!
As a blind person myself, I just wanted to express my appreciation to folks involved in the open-source projects related to multimodal vision-language models!
Thanks so much!

clip/README.md
```diff
@@ -0,0 +1,104 @@
...
+```
+python download.py openai/clip-vit-base-patch32
+```
+If the download is successful, the folder ```weights``` will appear. In the ```weights``` folder, there will be the HuggingFace configuration and weights for ```openai/clip-vit-base-patch32```.
+
+Now, convert the PyTorch weights to MLX with:
+
+```
+python convert.py --torch-path weights/openai/clip-vit-base-patch32 --mlx-path weights/mlx/openai/clip-vit-base-patch32
+```
```
awni, line 29: Why not do this in one step instead of two?

clip/README.md
```diff
@@ -0,0 +1,104 @@
...
+from image_processor import CLIPImageProcessor
+
+MODEL: str = "openai/clip-vit-base-patch32"
+CONVERTED_CKPT_PATH: str = f"weights/mlx/{MODEL}"
+
+# Load pretrained MLX CLIPModel
+mlx_clip = CLIPModel.from_pretrained(Path(CONVERTED_CKPT_PATH))
+# Load input tokenizer and transformers image (pre)processor
+tokenizer = CLIPTokenizer.from_pretrained(Path(CONVERTED_CKPT_PATH))
+img_processor = CLIPImageProcessor.from_pretrained(Path(CONVERTED_CKPT_PATH))
```
awni, line 54: We shoudl simplify this to a single call to a `load` like this:

```python
model, tokenizer, image_processor = load(string_path)
```

And use a `string_path` to load the model not a `Path`. Simplifies a lot for the caller.

clip/download.py
```diff
@@ -0,0 +1,51 @@
+"""Taken from https://github.com/ml-explore/mlx-examples/issues/155#issuecomment-1868120534"""
+import argparse
+import os
+
+import requests
+from tqdm import tqdm
+
```
awni, line 7: I think it's cleaner to use `snapshot_download` and we already have a dependency for it. Check this [example](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/utils.py#L57-L82).

Concretely you could merge the download and convert scripts very easily.
1. Take a HF repo and snapshot download the files you need. This will cache the files and give you a path.
2. Load what you need from the path and convert to MLX

clip/tokenizer.py
```diff
@@ -0,0 +1,128 @@
...
+import json
+from pathlib import Path
+from typing import Any, Union
+
+import mlx.core as mx
+import regex
+from huggingface_hub import hf_hub_download
+
+
+class CLIPTokenizer:
```
awni, line 12: Why not just use the `ClipTokenizer` from Transformers directly instead of copying the code here?

awni:
This is really coming along nicely! I left a few higher level comments / suggestions? Could you take a look and make changes as needed?

Once that's done I'll plan to do some more nitty gritty checking and run it myself as well.

awni:
> I suspect this is because CLIPImageProcessor uses Bicubic sampler for resizing. I am not sure whether we have Bicubic sampler in mlx-data.

It's not currently exposed. Could you file an issue in `mlx-data`? I think it should not be too difficult to add.

gboduljak:
> This is really coming along nicely! I left a few higher level comments / suggestions? Could you take a look and make changes as needed?
> 
> Once that's done I'll plan to do some more nitty gritty checking and run it myself as well.

Thank you for the feedback. I am to address your comments today or tomorrow. I will also file an issue in mlx-data for the bicubic sampler.

clip/README.md
gboduljak: I will implement something similar to the suggestion :)

clip/README.md
gboduljak: I did not know that `snapshot_download` exists. There is no reason why we should use two steps instead of one.

gboduljak: Let's use the one step.

clip/download.py
gboduljak: This is a great idea. I will implement something similar to the example.

clip/tokenizer.py
gboduljak: This is a good question. You could also ask 'Why not just use the `CLIPImageProcessor` from `transformers` instead of implementing the new one? My goal was to remove dependency on `transformers` by implementing simple and hackable implementations of the tokenizer and the image processor. These implementations are 'good enough' for demo purposes. In particular, their outputs are equal to the corresponding `transformers` outputs up to numerical accuracy.  Please see the tests.

We can remove custom implementations, but we will need `transformers`. I do not have a strong opinion regarding this. Please suggest whether we want simpler, custom implementations or a dependency on `transformers`.

gboduljak:
> Once that's done I'll plan to do some more nitty gritty checking and run it myself as well.

@awni I addressed your comments. Please take a look.

awni:
> This is a good question. You could also ask 'Why not just use the CLIPImageProcessor from transformers instead of implementing the new one? 

I like the idea of fewer dependencies. It is a bit of an art deciding where to draw the line. If the tokenizer is all that's standing in the way from that then it seems ok to me to keep the tokenizer you extracted.

gboduljak:
> I like the idea of fewer dependencies. It is a bit of an art deciding where to draw the line. If the tokenizer is all that's standing in the way from that then it seems ok to me to keep the tokenizer you extracted.

Thanks for fixing the nits and removing unnecessary inits. We need **both** `tokenizer.py` and `image_processor.py` to remove `transformers` dependency.

angeloskath (approved):
Looks great.

Hopefully we 'll be able to remove a bunch of those dependencies (image preprocessing and BPE tokenizer) using mlx-data in the near future :-)

awni (approved):
🚀

PR #318, opened 2024-01-15
feat(mlx_lm): add mixtral support in mlx_lm

mzbac:
Added the example for Mixtral in HF format, in preparation to support LoRa/QLoRa for Mixtral. 
Uploaded the quant model [here](https://huggingface.co/mlx-community/Mixtral-8x7B-Instruct-v0.1-hf-4bit-mlx).

### DIFF
--- a/llms/mlx_lm/convert.py
+++ b/llms/mlx_lm/convert.py
@@ -10,7 +10,7 @@ import mlx.nn as nn
 import transformers
 from mlx.utils import tree_flatten
 
-from .utils import get_model_path, load
+from .utils import get_model_path, linear_class_predicate, load
 
 MAX_FILE_SIZE_GB = 15
 
@@ -94,11 +94,10 @@ def quantize_model(
     model, _ = load(hf_path)
     model.load_weights(list(weights.items()))
 
-    nn.QuantizedLinear.quantize_module(model, q_group_size, q_bits)
-    quantized_config["quantization"] = {
-        "group_size": q_group_size,
-        "bits": q_bits,
-    }
+    nn.QuantizedLinear.quantize_module(
+        model, q_group_size, q_bits, linear_class_predicate=linear_class_predicate
+    )
+    quantized_config["quantization"] = {"group_size": q_group_size, "bits": q_bits}
     quantized_weights = dict(tree_flatten(model.parameters()))
 
     return quantized_weights, quantized_config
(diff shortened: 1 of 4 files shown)


llms/mlx_lm/convert.py
```diff
@@ -94,11 +94,10 @@ def quantize_model(
     model, _ = load(hf_path)
     model.load_weights(list(weights.items()))
 
-    nn.QuantizedLinear.quantize_module(model, q_group_size, q_bits)
-    quantized_config["quantization"] = {
-        "group_size": q_group_size,
-        "bits": q_bits,
-    }
+    nn.QuantizedLinear.quantize_module(
+        model, q_group_size, q_bits, linear_class_predicate=linear_class_predicate
```
mzbac, line 98: Add the linear_class_predicate so that we don't quantize any layer with dimensions that can't be divided by 32.

llms/mlx_lm/utils.py
```diff
@@ -10,16 +10,21 @@
...
 # Constants
 MODEL_MAPPING = {
     "llama": llama,
     "mistral": llama,  # mistral is compatible with llama
+    "mixtral": mixtral,
     "phi": phi2,
 }
 
+linear_class_predicate = (
+    lambda m: isinstance(m, nn.Linear) and m.weight.shape[0] % 32 == 0
```
mzbac, line 25: Using the same assertion (0th dimension) in mlx quantize, would go with `all (dim % 32 == 0 for dim in m.weight.shape)`, but it feels a bit overkill.

awni: @mzbac just FYI we are adding support for non 32 multiple weights in https://github.com/ml-explore/mlx/pull/458 so we will be able to remove this very soon.

awni (approved):
This is beautifully done, thank you for adding it.

What do you think, should we remove `llms/mixtral` or keep it there as a reference standalone MOE model?

llms/mlx_lm/utils.py
mzbac: > @mzbac just FYI we are adding support for non 32 multiple weights in [ml-explore/mlx#458](https://github.com/ml-explore/mlx/pull/458) so we will be able to remove this very soon.

🚀 🚀 🚀

mzbac:
> This is beautifully done, thank you for adding it.
> 
> What do you think, should we remove `llms/mixtral` or keep it there as a reference standalone MOE model?

I am a bit concerned that removing it will cause the old Mixtral model in mlx-community to no longer work. Given the size of the model, Maybe keeping it as a standalone reference for the MOE model is not a bad idea for now, but it may cause some confusion,  I'm not really sure. :(

awni:
You're right, let's keep it for now as it is one of the more used models in HF. We could replace that repo in the future if we decide to remove it.

PR #320, opened 2024-01-15
Missing requirements needed for convert script

bojanbabic:
This example requires pandas. Also added eos param that would stop generating the text if EOS token generated.

### DIFF
--- a/llms/llama/llama.py
+++ b/llms/llama/llama.py
@@ -234,9 +234,8 @@ def generate(args):
             # Actually perform the computation to measure the prompt processing time
             mx.eval(token)
             prompt_processing = toc("Prompt processing", start)
-        cur = tokenizer.decode([t.item() for t in tokens])
 
-        if len(tokens) >= args.max_tokens or cur.endswith(args.eos):
+        if len(tokens) >= args.max_tokens:
             break
 
         elif (len(tokens) % args.write_every) == 0:
@@ -385,7 +384,6 @@ if __name__ == "__main__":
     parser.add_argument(
         "--temp", type=float, default=0.0, help="The sampling temperature"
     )
-    parser.add_argument("--eos", type=str, default="\n\n\n", help="End of sequence")
     parser.add_argument("--seed", type=int, default=0, help="The PRNG seed")
 
     args = parser.parse_args()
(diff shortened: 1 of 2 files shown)


llms/llama/requirements.txt
```diff
@@ -1,3 +1,4 @@
 mlx>=0.0.6
 sentencepiece
 torch
+pandas
```
awni, line 4: I don't think it requires pandas..

bojanbabic: actually it requires `numpy`. let me change that.

llms/llama/llama.py
```diff
@@ -384,6 +385,7 @@ def load_model(model_path):
     parser.add_argument(
         "--temp", type=float, default=0.0, help="The sampling temperature"
     )
+    parser.add_argument("--eos", type=str, default="\n\n\n", help="End of sequence")
```
awni, line 388: Any reason to use an argument here instead of getting the EOS token from the tokenizer?

bojanbabic: Sometimes we want to stop generating on a predefined sequence, instead of EOS from the tokenizer.

mzbac: End of sequence may be application concerns, but it should not be the main focus of the example. I think we could include it as part of mlx-lm package support. However, personally, I feel we should keep the original llama example concise demonstrates how to run llama via mlx without adding too much unnecessary noise there.

awni: I agree with @mzbac on this. This example is definitely not meant to be feature complete, it's meant to be simple and readable. New features such as this one make a lot more sense in mlx-lm.

bojanbabic: Makes sense. I'll remove change that adds new arg and supporting code. requirements change should stay since it breaks conversion script. Let me know if you have any other concerns.

awni: Perfect thanks!

bojanbabic:
@awni @mzbac as per your suggestion, I've decided to leave only missing requirement change. This changes the nature of the PR, so I've changed the title of the PR as well in order to reflect what is in the PR.

mzbac (approved):

awni (approved):
Thanks!

PR #326, opened 2024-01-16
Fix import order of the `normalizing_flow` example

AtomicVar:
This repo fails to run `pre-commit run --all-files` -- the `isort` hook fails. This PR fixes import order of the `normalizing_flow` example. After this, the repo will stay in a clean state.

By the way, we should add a `pull_request_template.md` file to this repo (like the one in mlx repo) so that contributors can be checked to run pre-commit hooks.


awni (approved):
Thanks!!

And yea, a PR template is a good idea!

PR #333, opened 2024-01-18
fix the import path issue for mlx_lm

iLoveBug:
the pull request fix the problem when I try to run convert.py and generate.py, it reports the following issue:

  File "/Users/zhangfan/Documents/AI/mlx-examples/llms/mlx_lm/generate.py", line 6, in <module>
    from .utils import generate_step, load
ImportError: attempted relative import with no known parent package

### DIFF
--- a/llms/mlx_lm/convert.py
+++ b/llms/mlx_lm/convert.py
@@ -10,7 +10,7 @@ import mlx.nn as nn
 import transformers
 from mlx.utils import tree_flatten
 
-from .utils import get_model_path, linear_class_predicate, load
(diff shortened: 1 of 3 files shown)


awni:
mlx_lm is meant to be used as a package, so your fix will break that. Check out the [docs](https://pypi.org/project/mlx-lm/) for more on how it should be used.

Here is a sample command for the CLI:

```
python -m mlx_lm.generate --model mistralai/Mistral-7B-v0.1 --prompt "hello"
```

If you want to run it without building first, you should be able to just run that from outside the `mlx_lm` directory.

PR #337, opened 2024-01-19
feat: move lora into mlx-lm

mzbac:
This PR is demonstrating the idea of moving the Lora example into the mlx-lm package. The lora.py and fuse.py files behavours are being kept as is for ensuring a consistent user experience. However, in this PR, I am trying to modularize Lora example and make sure it can be decoupled from data loading, model loading, and generation. 

related issues:
https://github.com/ml-explore/mlx-examples/pull/213
https://github.com/ml-explore/mlx-examples/pull/235

Happy to change the implementation or dismiss it if we feel that it doesn't fit the purpose.


llms/mlx_lm/lora.py
```diff
@@ -0,0 +1,245 @@
...
+    args = parser.parse_args()
+
+    np.random.seed(args.seed)
+
+    print("Loading pretrained model")
+    model, tokenizer = load(args.model)
+
+    # Freeze all layers other than LORA linears
+    model.freeze()
+    for l in model.model.layers[len(model.model.layers) - args.lora_layers :]:
```
mzbac, line 166: Ideally, when applying Lora linear, it should be moved into the trainer. The user only needs to specify the number of Lora layers and provide a targeted linear layer name list. The trainer should then apply Lora linear based on those specifications. However, I couldn't figure out a good way to do that. Maybe @awni you have some thoughts on how to do it?

llms/mlx_lm/tuner/trainer.py
```diff
@@ -0,0 +1,224 @@
...
+        self,
+    ):
+        if not self.save_dir.is_dir():
+            raise ValueError(f"The provided path {self.save_dir} is not a directory.")
+
+        flattened_tree = tree_flatten(self.model.trainable_parameters())
+
+        mx.savez(self.save_dir / self.adapter_file, **dict(flattened_tree))
+        # Save the LoRA layer names
+        param_names = list(
```
mzbac, line 212: Save the additional Lora layer parameter name in the JSON file, along with the adapter weight. So, if we can find a way to apply the Lora linear according to the JSON later, that way, in fuse.py or when generating text from model with an adapter, we won't have to manually apply the Lora linear. We should be able to read from JSON and attach Lora to the base model. This will make distributing Lora and inference easier.

llms/mlx_lm/utils.py
```diff
@@ -142,42 +142,40 @@ def generate(
     return tokens
 
 
-def load(path_or_hf_repo: str) -> Tuple[nn.Module, PreTrainedTokenizer]:
+def load_model(model_path: Path) -> nn.Module:
```
mzbac, line 145: Just move the `load_model` out of the `load` function in case we only want to load the model and leave it flexible for users to initialize their own tokenizer. The tokenizer is currently very variable in LLM space at the moment.

awni:
This looks awesome! I'm looking forward to reviewing. I may not get to it this weekend depending on scheduling but if not, then Monday!

mzbac:
> This looks awesome! I'm looking forward to reviewing. I may not get to it this weekend depending on scheduling but if not, then Monday!

No rush, enjoy the weekend:)

ivanfioravanti:
great work @mzbac I was looking forward for lora mlx-lm integration!

llms/mlx_lm/models/base.py
```diff
@@ -4,6 +4,8 @@
 
 @dataclass
 class BaseModelArgs:
+    train: bool = False
```
mzbac, line 7: Add this flag to the arguments because It needs to initialize different model layers based on whether it is for training or inference. This will prevent the need for conditional checking in the forwarding pass, which can cause some performance degradation.

llms/mlx_lm/models/mixtral.py
```diff
@@ -174,14 +175,60 @@ def __call__(self, x: mx.array) -> mx.array:
...
 class MixtralDecoderLayer(nn.Module):
     def __init__(self, args: ModelArgs):
         super().__init__()
         self.hidden_size = args.hidden_size
 
         self.self_attn = MixtralAttention(args)
 
-        self.block_sparse_moe = MixtralSparseMoeBlock(args)
+        self.block_sparse_moe = (
+            MixtralSparseMoeTrainingBlock(args)
```
mzbac, line 228: For training/inferencing, we use different models layers to avoid performance degradation.

llms/mlx_lm/tuner/utils.py
```diff
@@ -0,0 +1,22 @@
+import mlx.core as mx
+from mlx.utils import tree_unflatten
+
+from .linear import LoRALinear
+
+
+def apply_lora_layers(model, adapter_file: str):
```
mzbac, line 7: With this helper, we can recover the lora layers and update weights from a single adapter file.

llms/mlx_lm/models/mixtral.py
awni: It might be cleaner to to use the same __call__ and check the last dimension of `y` just before the expert loop and branch based on that. That way we could also use the batching version prompts. Wdyt?

awni: Indeed that’s exactly why I didn’t do it in the first place for inference only. I think it has to do with memory use being different but I didn’t figure it out yet.

but sounds good let’s keep it separate for now until perf issues are debugged.

awni: Another alternative is to use the Module.training flag to branch? I am not sure what the implications are for the rest of the code

mzbac: > Another alternative is to use the Module.training flag to branch? I am not sure what the implications are for the rest of the code

I have considered using `model.train()`, but I need to construct different layers during model initialization. So, `model.train` still act as a runtime behavior change that is equivalent to an if-else statement in the forward pass. Maybe I misunderstood.

awni: When you notice the perf degradation that was for inference right? The degradation is almost certainly not due to just checking the condition but due to having a different path for prompt processing and single token generation.

If you use model.training to branch instead of the shape of the input in the `__call__` function you will have the same path for prompt processing and token generation (the loop) and a different branch for training (the batching over experts).

mzbac: > When you notice the perf degradation that was for inference right? The degradation is almost certainly not due to just checking the condition but due to having a different path for prompt processing and single token generation.
> 
> If you use model.training to branch instead of the shape of the input in the `__call__` function you will have the same path for prompt processing and token generation (the loop) and a different branch for training (the batching over experts).

Sounds good, I will give it a try. If that works, it would be great. I don't really like the current implementation and adding an additional train flag in model arguments.:)

mzbac: module.training works well. I didn't find any performance impact from my local test. Thank you for pointing out the better solution 🚀

llms/mlx_lm/lora.py
ProjectProgramAMark: Have we considered adopting an option for a yaml configuration file, such as #235 was working on? I think there can be a way to cleanly grab from a config file such as this, provide defaults in the case of missing params, but also allow users to pass their own layers without a yaml config file, for example in the case of rapid prototyping on a notebook? Also, we can probably have layer checking in a separate function somewhere. That would keep the logic somewhat clean. Let me know what you think

mzbac: Yeah, reading from a YAML config file is definitely a good option. Once this PR is merged, we can start exploring the options and determine the best way to proceed. Ideally, I think it would be great if we could provide a similar API like peft lib has in loraconfig which takes target_modules as input and automatically applies the LoRa based on that.
https://huggingface.co/docs/peft/tutorial/peft_model_config

awni:
Excellent thanks! This is next on my list! Very excited about it and sorry for the delay!

mzbac:
> Excellent thanks! This is next on my list! Very excited about it and sorry for the delay!

No rush, al good. Thank you for helping with the review. :)

llms/phixtral/phixtral.py
```diff
@@ -198,7 +198,7 @@ def __call__(
         x: mx.array,
         mask: mx.array = None,
         cache: mx.array = None,
-    ) -> tuple[mx.array, mx.array]:
+    ) -> Tuple[mx.array, mx.array]:
```
awni, line 201: 👍

llms/mlx_lm/lora.py
awni: - I think it's a good idea to do it in steps (this PR followed by refactor improvements
- I agree this part of the code belongs in the tuner. 
- I would say a high level spec in a json or yaml (I happen to prefer json 😄 ) is a good idea.

llms/mlx_lm/lora.py
```diff
@@ -0,0 +1,249 @@
...
+            optimizer=opt,
+            loss=default_loss,
+            train_dataset=train_set,
+            val_dataset=valid_set,
+            test_dataset=test_set,
+        )
+        # Train model
+        trainer.train()
+        # Save adapter weights
+        trainer.save_adapter()
```
awni, line 214: Since the trainer saves the adapters, why not just have it make a call to save at the end of `train()` and avoid the additional API for the caller?

llms/mlx_lm/lora.py
```diff
@@ -0,0 +1,249 @@
...
+            args.batch_size,
+            num_batches=args.test_batches,
+        )
+        test_ppl = math.exp(test_loss)
+
+        print(f"Test loss {test_loss:.3f}, Test ppl {test_ppl:.3f}.")
+
+    if args.prompt is not None:
+        print("Generating")
+        generate(
```
awni, line 242: Great that this is shared now 👍

llms/mlx_lm/lora.py
```diff
@@ -0,0 +1,249 @@
...
+
+    print("Loading datasets")
+    train_set, valid_set, test_set = load_dataset(args)
+
+    # Resume training the given adapters.
+    if args.resume_adapter_file is not None:
+        print(f"Loading pretrained adapters from {args.resume_adapter_file}")
+        model.load_weights(args.resume_adapter_file, strict=False)
+    # init training args
+    training_args = TrainingArguments(
```
awni, line 187: Nit: how about `TrainingArgs` for consistency with `ModelArgs`?

llms/mlx_lm/lora.py
```diff
@@ -0,0 +1,249 @@
...
+        print("Training")
+        model.train()
+        opt = optim.Adam(learning_rate=args.learning_rate)
+        # init trainer
+        trainer = LoraTrainer(
+            model=model,
+            tokenizer=tokenizer,
+            args=training_args,
+            optimizer=opt,
+            loss=default_loss,
```
awni, line 206: I would remove the default loss here since it's the default, and until we have a need to specialize it I don't think we need to expose it as a parameter. It makes the calling code a little more verbose.

llms/mlx_lm/lora.py
mzbac: I didn't really think about this. My intuition was to make it explicit, but now that we save_per_step during training, I don't feel there is a reason why we can't move it into the train step.

llms/mlx_lm/tuner/linear.py
```diff
@@ -0,0 +1,88 @@
+import math
```
awni, line 1: I feel like the name of this file should be a bit more descriptive. Maybe `lora_layers.py` or even `lora.py`?

llms/mlx_lm/tuner/trainer.py
```diff
@@ -0,0 +1,204 @@
...
+        iterate_batches(dataset, tokenizer, batch_size, max_seq_length),
+    ):
+        losses, toks = loss(model, *batch)
+        all_losses.append((losses * toks).item())
+        ntokens += toks.item()
+
+    return np.sum(all_losses) / ntokens
+
+
+class LoraTrainer:
```
awni, line 106: This is a really general training loop, it might be better to name it `Trainer` rather than `LoraTrainer`.

I'm also wondering why you prefer it as a class? It doesn't seem like holding state is necessary and from an implementation and use standpoint it might be simpler to just have a `trainer.train` function (also more consistent with `evaluate`)

llms/mlx_lm/lora.py
```diff
@@ -0,0 +1,249 @@
...
+            train_dataset=train_set,
+            val_dataset=valid_set,
+            test_dataset=test_set,
```
awni, line 209: I would remove the test set from the trainer. It is unused (as far as I can tell) and it encourage good data hygiene to not let it be used durin gtraining.

awni (changes requested):
This is super awesome 🚀 !! I left mostly minor comments. 

I think the only slightly non-simple comment is around the decision to make `LoraTrainer` a class versus just a function `train` (either way I would not put `Lora` in the name since it is quite general and could be reused). I am curious what you think about that.

llms/mlx_lm/lora.py
mzbac: It's mostly the args for the training. Making it "model args" feels a bit misleading.

llms/mlx_lm/lora.py
mzbac: Yeah, I'm happy to remove it. It was put there as an example for people who try to read the example and know that they can pass a custom loss function, but I think we can mention it in the API documentation.

llms/mlx_lm/tuner/linear.py
mzbac: either of them makes sense to me, I will update it.

llms/mlx_lm/tuner/trainer.py
mzbac: Initially, my intention was to encapsulate all elements related to training within the trainer object. However, after making some changes, I feel that both a class and a function could be good options. Considering the `evaluate` used by testing, it now seems more logical to use the "train" function. I am happy to refactor it into a function.

llms/mlx_lm/lora.py
mzbac: I will update it after refactoring the trainer class.

llms/mlx_lm/lora.py
awni: Sorry I think you misunderstood:

`TrainingArgument` -> `TrainingArgs`

mzbac: > Sorry I think you misunderstood:
> 
> `TrainingArguments` -> `TrainingArgs`
> 
> That way we use `Args` consistently

Sorry for the misunderstanding. It's definitely a good suggestion, and I will update it accordingly.

gladjoyhub:
When can we pip install mlx-lm with lora? Can't wait! Thanks!

awni:
As soon as this lands I will bump the version and upload the new package.

mzbac:
@awni I have addressed all the comments. Let me know if there is anything else I missed.

gladjoyhub:
@mzbac
With your newest lora utils.py, I tried:

model, tokenizer = load("/mlx_model")
model.load_weights("/mlx_adapter/lora_adapter.npz, strict=False)
response = generate(model, tokenizer, prompt=llama_prompt, max_tokens=50)

But the response is same as base model's response, the loaded adapter weights didn't seem to take effect. Please help.

The reason I didn't use fuse, is because I'd like to use one base model, but want to be able to switch among different trained adaptors, hence model.load_weights()

Is this possible in the coming mlx-lm?

llms/mlx_lm/lora.py
```diff
@@ -0,0 +1,252 @@
...
+if __name__ == "__main__":
+    parser = build_parser()
+    args = parser.parse_args()
+
+    np.random.seed(args.seed)
+
+    print("Loading pretrained model")
+    model, tokenizer = load(args.model)
+
+    if model.__class__ not in SUPPORTED_MODELS:
```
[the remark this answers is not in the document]
awni: Seems like a safe option

mzbac:
> @mzbac With your newest lora utils.py, I tried:
> 
> from mlx_lm import load, generate model, tokenizer = load("/mlx_model") model.load_weights("/mlx_adapter/lora_adapter.npz, strict=False) response = generate(model, tokenizer, prompt=llama_prompt, max_tokens=50)
> 
> But the response is same as base model's response, the loaded adapter weights didn't seem to take effect. Please help.
> 
> The reason I didn't use fuse, is because I'd like to use one base model, but want to be able to switch among different trained adaptors, hence model.load_weights()
> 
> Is this possible in the coming mlx-lm?

Yeah,when you load the Lora weight, you have to make sure your model has all replaced Lora layers. to make sure your model structure in sync with the lora weight, It's better to use this helper function to load the adapter: [link](https://github.com/ml-explore/mlx-examples/blob/3a2dde67ff0d2617ea127f47c0522e69501c4571/llms/mlx_lm/tuner/utils.py#L7)
```python
model, tokenizer = load("/mlx_model")
apply_lora_layers(model,"/mlx_adapter/lora_adapter.npz")
response = generate(model, tokenizer, prompt=llama_prompt, max_tokens=50)
```

awni:
I think it looks awesome!! I am just going to push a couple changes about usage and then I will merge it.

gladjoyhub:
> > @mzbac With your newest lora utils.py, I tried:
> > from mlx_lm import load, generate model, tokenizer = load("/mlx_model") model.load_weights("/mlx_adapter/lora_adapter.npz, strict=False) response = generate(model, tokenizer, prompt=llama_prompt, max_tokens=50)
> > But the response is same as base model's response, the loaded adapter weights didn't seem to take effect. Please help.
> > The reason I didn't use fuse, is because I'd like to use one base model, but want to be able to switch among different trained adaptors, hence model.load_weights()
> > Is this possible in the coming mlx-lm?
> 
> Yeah,when you load the Lora weight, you have to make sure your model has all replaced Lora layers. to make sure your model structure in sync with the lora weight, It's better to use this helper function to load the adapter: [link](https://github.com/ml-explore/mlx-examples/blob/3a2dde67ff0d2617ea127f47c0522e69501c4571/llms/mlx_lm/tuner/utils.py#L7)
> 
> ```python
> from mlx_lm import load, generate
> model, tokenizer = load("/mlx_model")
> apply_lora_layers(model,"/mlx_adapter/lora_adapter.npz")
> response = generate(model, tokenizer, prompt=llama_prompt, max_tokens=50)
> ```
> 
> And I think adding support to load the adapter from the load function is definitely something doable. I will see if I can add it after this PR gets merged.

Yes, apply_lora_layers worked. load_adapter would be great! Thanks!

awni (approved):
Ok! I added a good enough doc for the CLI `LORA.md`. We can also update it to include the API usage once we expose the high level functions to the API.

PR #338, opened 2024-01-19
add an option to apply the tokenizer chat template

batmac:
Hi,
here is a small tweak to mlx_lm to be able to apply the defined chat_template (in tokenizer_config.json) when using it to quickly and correctly generate in cli.

I love your work, 😍

### DIFF
--- a/llms/mlx_lm/generate.py
+++ b/llms/mlx_lm/generate.py
@@ -64,15 +64,14 @@ def main(args):
 
     model, tokenizer = load(args.model, tokenizer_config=tokenizer_config)
 
-    if not args.ignore_chat_template:
+    if not args.ignore_chat_template and (
+        hasattr(tokenizer, "apply_chat_template")
+        and tokenizer.chat_template is not None
+    ):
         messages = [{"role": "user", "content": args.prompt}]
-        if (
-            hasattr(tokenizer, "apply_chat_template")
-            and tokenizer.chat_template is not None
-        ):
-            prompt = tokenizer.apply_chat_template(
-                messages, tokenize=False, add_generation_prompt=True
-            )
+        prompt = tokenizer.apply_chat_template(
+            messages, tokenize=False, add_generation_prompt=True
+        )
     else:
         prompt = args.prompt
 
--- a/llms/mlx_lm/generate.py
+++ b/llms/mlx_lm/generate.py
@@ -46,6 +46,12 @@ def setup_arg_parser():
         "--temp", type=float, default=DEFAULT_TEMP, help="Sampling temperature"
     )
     parser.add_argument("--seed", type=int, default=DEFAULT_SEED, help="PRNG seed")
+    parser.add_argument(
+        "--apply_chat_template",
+        "-a",
+        action="store_true",
+        help="apply the tokenizer's chat template on the prompt",
+    )
     return parser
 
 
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/generate.py
```diff
@@ -35,15 +35,31 @@ def setup_arg_parser():
...
 
 def main(args):
     mx.random.seed(args.seed)
     model, tokenizer = load(args.model)
+    if args.apply_chat_template:
+        messages = [{"role": "user", "content": args.prompt}]
+        # test if the tokenizer has a chat template
+        if not hasattr(tokenizer, "apply_chat_template"):
+            raise ValueError(f"{args.model} does not have a chat template")
+        prompt = tokenizer.apply_chat_template(
```
mzbac, line 55: Looks like a good improvement. However, could you explain why you would like to disable `add_special_tokens`? I quickly looked at the documentation and it seems that there are more usages such as `tokenizer.apply_chat_template(messages, tokenize=False, add_generation_prompt=True)`.

batmac: oh I think you're right. My brain has swapped them I suppose ! Good catch. Let me fix that.

batmac: (for reference, here is the code in the transformers lib https://github.com/huggingface/transformers/blob/3f69f415adcbdaedec154ba8eac220ef3276975d/src/transformers/tokenization_utils_base.py#L1678 )

mzbac (approved):
LGTM, and I have seen people get confused when they try to interact with mlx-lm using the wrong chat template.

llms/mlx_lm/generate.py
```diff
@@ -35,15 +35,31 @@ def setup_arg_parser():
...
     return parser
 
 
 def main(args):
     mx.random.seed(args.seed)
     model, tokenizer = load(args.model)
+    if args.apply_chat_template:
+        messages = [{"role": "user", "content": args.prompt}]
+        # test if the tokenizer has a chat template
+        if not hasattr(tokenizer, "apply_chat_template"):
```
mzbac, line 53: nit: could add one more check if `tokenizer.chat_template` is set. Otherwise, `apply_chat_template` may use the default template, which can cause some unexpected behavior. https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_utils_base.py#L1732-L1737

batmac: I've now added a message to warn about that , owever while testing it I see that the transformers lib also prints a warning, so it may be too much? Thank you for your reviews and tell me what you think.

example of missing template:
```python
❯ python -m mlx_lm.generate --model TinyLlama/TinyLlama-1.1B-Chat-v0.3 --max-tokens 20 --prompt 'I have a capybara in my garden! What should I do?' -a
Fetching 8 files: 100%|███████████████████████| 8/8 [00:00<00:00, 122910.01it/s]
Special tokens have been added in the vocabulary, make sure the associated word embeddings are fine-tuned or trained.
TinyLlama/TinyLlama-1.1B-Chat-v0.3 does not define a chat template, using the default template, which may cause unexpected results

No chat template is defined for this tokenizer - using the default template for the LlamaTokenizerFast class. If the default is not appropriate for your model, please set `tokenizer.chat_template` to an appropriate template. See https://huggingface.co/docs/transformers/main/chat_templating for more information.

==========
Prompt: (...)
```

(I've also rework the other error message to better reflect the problem I think)

mzbac: Maybe only enable the chat template when the tokenizer has applied a chat template and has a tokenizer.chat_template. 
For example:
```python
if hasattr(tokenizer, "apply_chat_template") and tokenizer.chat_template:
    prompt = tokenizer.apply_chat_template(messages, tokenize=False, add_generation_prompt=True)
```

awni: I like this idea, let's use the prompt as is when no chat template exists. That seems clean to me.

awni (commented):
This seems like a great change to reduce confusion and make it easier to use the right template. 

I'm wondering why it should even be an option (especially defaulting to False). Should we not just always apply the chat template if it exists? And if it does not exist use the prompt as is?

If that is too implicit, we could have it be an option but I would suggest changing the option to default `True` (so `--ignore-chat-template` instead.

Does that seem reasonable?

batmac:
As a regular user of "python -m mlx_lm.generate ..." myself to quickly generate stuff in my terminal, I think using a template by default with an option to not insert it (to keep the ability to experiment) would be great (as I basically always use a "formatted prompt")

Let me quickly commit this change so we can test how it feels.

awni (approved):
Thank you !! 🚀

Solido:
The warning message is confusing as it also appear when launching a lora training session.
I'm using the `jsonl text` format. What should I configure? does this collide with the openAI format commit that register a specific format too?

Issue #305, opened 2024-01-12
Throws a "received parameters not in model" error when trying to convert llama2:70b hf to 4 bit npz


Satyam7166-tech:
I was trying to convert llama2:70b hf to 4bit quantisation for performing Qlora finetuning. 
> ```
> python convert.py --hf-path <model> -q
> ```

I ran this in `mlx_examples/lora` My exact command was:

`python convert.py --hf-path meta-llama/Llama-2-70b-chat-hf -q --mlx-path ./`

But I recieved an error. This is my trace:

```
INFO] Loading
config.json: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 614/614 [00:00<00:00, 5.96MB/s]
generation_config.json: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 188/188 [00:00<00:00, 781kB/s]
model-00008-of-00015.safetensors: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████| 9.80G/9.80G [54:38<00:00, 2.99MB/s]
model-00006-of-00015.safetensors: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████| 9.80G/9.80G [56:31<00:00, 2.89MB/s]
model-00002-of-00015.safetensors: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████| 9.80G/9.80G [1:02:40<00:00, 2.61MB/s]
model-00005-of-00015.safetensors: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████| 9.80G/9.80G [1:11:26<00:00, 2.29MB/s]
model-00007-of-00015.safetensors: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████| 9.97G/9.97G [1:28:44<00:00, 1.87MB/s]
model-00009-of-00015.safetensors: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████| 9.80G/9.80G [56:14<00:00, 2.90MB/s]
model-00010-of-00015.safetensors: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████| 9.80G/9.80G [58:25<00:00, 2.80MB/s]
model-00015-of-00015.safetensors: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████| 524M/524M [02:58<00:00, 2.93MB/s]
[...]

PR #340, opened 2024-01-20
chore: fix the convert.py script for weights are not sanitized and support quant for non-32 dimensions

mzbac:
Fix for: https://github.com/ml-explore/mlx-examples/issues/305

### DIFF
--- a/llms/mlx_lm/convert.py
+++ b/llms/mlx_lm/convert.py
@@ -10,7 +10,7 @@ import mlx.nn as nn
 import transformers
 from mlx.utils import tree_flatten
 
-from .utils import get_model_path, linear_class_predicate, load
+from .utils import get_model_path, linear_class_predicate, load_model
 
 MAX_FILE_SIZE_GB = 15
 
@@ -60,30 +60,23 @@ def fetch_from_hub(
 ) -> Tuple[Dict, dict, transformers.PreTrainedTokenizer]:
     model_path = get_model_path(model_path)
 
-    weight_files = glob.glob(f"{model_path}/*.safetensors")
-    if not weight_files:
-        raise FileNotFoundError(f"No safetensors found in {model_path}")
-
-    weights = {}
-    for wf in weight_files:
-        weights.update(mx.load(wf).items())
+    model = load_model(model_path)
 
     config = transformers.AutoConfig.from_pretrained(model_path)
     tokenizer = transformers.AutoTokenizer.from_pretrained(model_path)
 
-    return weights, config.to_dict(), tokenizer
+    return model, config.to_dict(), tokenizer
 
 
 def quantize_model(
-    weights: dict, config: dict, hf_path: str, q_group_size: int, q_bits: int
+    model: nn.Module, config: dict, q_group_size: int, q_bits: int
 ) -> tuple:
     """
     Applies quantization to the model weights.
 
     Args:
-        weights (dict): Model weights.
+        model (nn.Module): The model to be quantized.
         config (dict): Model configuration.
-        hf_path (str): HF model path..
         q_group_size (int): Group size for quantization.
         q_bits (int): Bits per weight for quantization.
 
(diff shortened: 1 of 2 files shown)


llms/mlx_lm/convert.py
```diff
@@ -60,39 +60,30 @@ def fetch_from_hub(
 ) -> Tuple[Dict, dict, transformers.PreTrainedTokenizer]:
     model_path = get_model_path(model_path)
 
-    weight_files = glob.glob(f"{model_path}/*.safetensors")
```
mzbac, line 63: The weights directly loaded from safetensors were not sanitized, so it will still throw an error during quantization. Therefore, change to load the model from utils which has weight sanitization.

llms/mlx_lm/convert.py
```diff
@@ -60,39 +60,30 @@ def fetch_from_hub(
...
     config = transformers.AutoConfig.from_pretrained(model_path)
     tokenizer = transformers.AutoTokenizer.from_pretrained(model_path)
 
-    return weights, config.to_dict(), tokenizer
+    return model, config.to_dict(), tokenizer
 
 
 def quantize_model(
-    weights: dict, config: dict, hf_path: str, q_group_size: int, q_bits: int
+    model: nn.Module, config: dict, q_group_size: int, q_bits: int
```
mzbac, line 72: Instead of passing the weights and reloading the model, just pass the model to avoid reloading it.

llms/mlx_lm/convert.py
```diff
@@ -183,12 +174,16 @@ def convert(
...
+    model, config, tokenizer = fetch_from_hub(hf_path)
+
+    weights = dict(tree_flatten(model.parameters()))
     dtype = mx.float16 if quantize else getattr(mx, dtype)
     weights = {k: v.astype(dtype) for k, v in weights.items()}
+
     if quantize:
         print("[INFO] Quantizing")
-        weights, config = quantize_model(weights, config, hf_path, q_group_size, q_bits)
+        model.load_weights(list(weights.items()))
```
mzbac, line 185: Reload the weights to ensure that the model's weights are converted to fp16 for better quantization.

mzbac:
@awni, this will fix the broken convert script for the llama model and enable quantization for the yayi model.

llms/mlx_lm/convert.py
awni: Good catch!

llms/mlx_lm/utils.py
```diff
@@ -22,8 +22,10 @@
 }
 
 linear_class_predicate = (
-    lambda m: isinstance(m, nn.Linear) and m.weight.shape[0] % 32 == 0
-)  # TODO remove this once we support quantization for non-multiples of 32
+    lambda m: isinstance(m, nn.Linear)
+    and m.weight.shape[0]
+    != 8  # avoid quantizing gate layers, otherwise we have to re-quant and uplaod all the mixtral models
```
awni, line 27: 👍 at some point we can remove this... but for now I think it's ok.

Nit: typo `uplaod` -> `upload`

llms/mlx_lm/utils.py
```diff
@@ -142,33 +144,33 @@ def generate(
     return tokens
 
 
-def load(path_or_hf_repo: str) -> Tuple[nn.Module, PreTrainedTokenizer]:
+def load_model(model_path: Path) -> nn.Module:
     """
-    Load the model from a given path or a huggingface repository.
+    Load and initialize the from a given path.
```
awni, line 149:
```suggestion
    Load and initialize the model from a given path.
```

awni (approved):
Looks great, thanks for the fix. I left two nits re comments, if you don't mind fixing.

Ready to go o/w 🚀

llms/mlx_lm/utils.py
mzbac: Thanks for pointing out the typo. I have fixed it :)

PR #342, opened 2024-01-20
fix the chinese character generation as same as PR #321

iLoveBug:
It's the same fix as PR #321 as it fixed the code in utils.py, but the function generation is not used in main function.
So I do it once again here.

### DIFF
--- a/llms/mlx_lm/generate.py
+++ b/llms/mlx_lm/generate.py
@@ -114,6 +114,7 @@ def main(args):
     tic = time.time()
     tokens = []
     skip = 0
+    REPLACEMENT_CHAR = "\ufffd"
     for token, n in zip(
         generate_step(prompt, model, args.temp, args.colorize), range(args.max_tokens)
     ):
@@ -125,12 +126,11 @@ def main(args):
             tic = time.time()
         tokens.append(token.item())
         s = tokenizer.decode(tokens)
-        if args.colorize:
-            colorprint_by_t0(t0,s[skip:])
-        else:
+        if REPLACEMENT_CHAR not in s:
             print(s[skip:], end="", flush=True)
-        skip = len(s)
-    print(tokenizer.decode(tokens)[skip:], flush=True)
+            skip = len(s)
+    tokens = tokenizer.decode(tokens).replace(REPLACEMENT_CHAR, "")
+    print(tokens[skip:], flush=True)
     gen_time = time.time() - tic
     print("=" * 10)
     if len(tokens) == 0:
--- a/llms/mlx_lm/generate.py
+++ b/llms/mlx_lm/generate.py
@@ -1,9 +1,8 @@
 import argparse
-import time
 
 import mlx.core as mx
 
-from .utils import generate_step, load
+from .utils import generate, load
 
 DEFAULT_MODEL_PATH = "mlx_model"
 DEFAULT_PROMPT = "hello"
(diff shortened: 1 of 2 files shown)


mzbac:
Thanks for the PR, but the fix feels a bit repetitive. Would you mind updating the main to use "generate" from utils instead? Personally, I feel that would be a better solution.

awni:
I agree with @mzbac , but the only difficulty with that is not removing the timing information from the CLI version. Some options:

1. Remove the timing information for the CLI
2. Change utils.py generate to return a dictionary of a result which includes other data/metadata (currently the timing info, but we may add more in the future like probabilities)
3. Always print timing in the utils.py generation (when in VERBOSE mode)


I think it's nice to have the timing information easily there so I like 2 or 3 but I'm also not that opposed to removing it.

iLoveBug:
It's always a good idea to reduce duplicate code as much as possible. 

I’d like to keep the timing information, and I even print the timing information in lora.py in my local copy so that I can see the total fine-tuning time used.

So I think maybe option 3 is a good idea: print timing information in verbose mode?

llms/mlx_lm/generate.py
```diff
@@ -35,41 +34,15 @@ def setup_arg_parser():
         "--temp", type=float, default=DEFAULT_TEMP, help="Sampling temperature"
     )
     parser.add_argument("--seed", type=int, default=DEFAULT_SEED, help="PRNG seed")
+    parser.add_argument('-v', '--verbose', help='Increase output verbosity', action='store_true')
```
awni, line 37: 👍

llms/mlx_lm/generate.py
```diff
@@ -35,41 +34,15 @@ def setup_arg_parser():
...
-    if len(tokens) == 0:
-        print("No tokens generated for this prompt")
-        return
-    prompt_tps = prompt.size / prompt_time
-    gen_tps = (len(tokens) - 1) / gen_time
-    print(f"Prompt: {prompt_tps:.3f} tokens-per-sec")
-    print(f"Generation: {gen_tps:.3f} tokens-per-sec")
+    output = generate(model, tokenizer, args.prompt, args.temp, args.max_tokens, args.seed, args.verbose)
+    if not args.verbose:
+        print(output)
```
awni, line 45: Let's just fix verbose to `True` for `generate.py`. I don't think anyone will want to use it without verbosity (that's meant more for the API)

llms/mlx_lm/generate.py
awni, line 37: Remove per below comment.

llms/mlx_lm/utils.py
```diff
@@ -105,6 +107,7 @@ def generate(
     prompt: str,
     temp: float = 0.0,
     max_tokens: int = 100,
+    seed: int = 0,
```
awni, line 110: Also can you revert to set the seed outside this function (e.g. before you call `generate` in `generate.py`. I think I prefer that to give the caller more control over how to seed the RNG state.

awni (changes requested):
Looks great, nice simplification! Just a couple minor changes and then it's good to merge. Thanks for the contribution!

awni:
Also could you rebase this on main as it appears there is a conflict. Thank you!

awni (approved):
Thank you looks great!

awni (approved):
I did some changes here to fix color formatting with this. It should all be mostly good now. But I changed the way we do formatting to a user provided callback to keep the main generation loop simple.

Thanks for the addition.

Issue #323, opened 2024-01-16
Moe implementation crashes during fine-tuning


mzbac:
Currently, the Moe implementation loops through selected experts per token. This works fine for inference because of the kv cache; it usually only computes one token at a time. However, it causes issues in training because it needs to loop over batch_size*sequence_length. I'm wondering if there is a way to vectorize this computation.

https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/mixtral.py#L167-L172

awni:
Definitely there is a way to vectorize it! I [tried this](https://github.com/ml-explore/mlx-examples/blob/e42682dced2d301123da3a260179b5f4bfdfad9e/mixtral/mixtral.py#L132-L141) at one point, but it was very slow. Something like it might work better for training though where batching is more useful

awni:
Though why is it crashing for you? The implicit eval might be an issue 🤔 for training..

mzbac:
> Though why is it crashing for you? The implicit eval might be an issue 🤔 for training..

When I load the quant model and use Lora with the default configuration, the memory usage goes up to over 120GB and eventually gets killed by the operating system. I can see that the model is stuck in a for loop at https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/mixtral.py#L168. My assumption is that it will take 4*2048 steps there, and we have 32 layers of the sparse MOE block. The computation required would be 32*4*2048. Perhaps that's why it crashs

awni:
Yes lol that will never work for trianing.. that part of the model was really only designed for inference.. it has to be vectorized to work with training. The implicit eval could still be an issue (it might be slow), but I think it should work.

mzbac:
> Definitely there is a way to vectorize it! I [tried this](https://github.com/ml-explore/mlx-examples/blob/e42682dced2d301123da3a260179b5f4bfdfad9e/mixtral/mixtral.py#L132-L141) at one point, but it was very slow. Something like it might work better for training though where batching is more useful

@awni I tried to replicate the transformers implementation(https://github.com/huggingface/transformers/blob/main/src/transformers/models/mixtral/modeling_mixtral.py#L810-L840), but I got stuck on expert_mask and torch.where. I can't really figure out how to implement it using mx...

mzbac:
[...]

PR #343, opened 2024-01-20
chore(lora): support mixtral in lora example

mzbac:
Quick and dirty implementation to support Mixtral while still work in progress on moving Lora to mlx-lm.
Close https://github.com/ml-explore/mlx-examples/issues/323


awni (approved):
Look great! Thank you!!

maximegmd:
Hello, thanks for implementing this, but could you also document the hardware requirements as well as commands that fit on the target hardware? I suppose one would need to train only a few layers to fit in system memory.

Issue #345, opened 2024-01-20
[Feature Request] Colorize output of mlx_lm based on token probability


ivanfioravanti:
Hi all,
let me start with: "MLX rocks"! 
What do you think of adding a way to colorize the output based on the probability of the token? 
This can be achieved integrating [@antirez](https://github.com/antirez) work in LLM-FTC-sampling repo [here](https://github.com/antirez/LLM-FTC-sampling/commit/f45c2b4086c4fed03610c889b9d16488cf5cb5cb)
We can make this optional to avoid impact on performance.

Personally I see this more useful on lora.py during fine tuning, to easily visualize how confident the model is in specific scenarios with different checkpoints save.

I see mlx_lm and lora are going to be merged soon in #337 so this can be integrated later in case.

I'll work on this one for me and create a PR to be evaluated in case

PR #347, opened 2024-01-20
Add colorized output option to generate script

ivanfioravanti:
Two new functions were added to the script that allow output to be colorized based on the T[0] probability. Changes were made to the `generate_step` function in utils.py to permit colorization. Additionally, an argument for colorization was introduced to the command-line parser.

I tried to avoid performance impact when --colorize is not passed. 

Related issue:
#345


mzbac:
Looks like a good improvement. Maybe later, we can refactor the generate code a bit so that inference and lora don't share the same function, allowing them to decouple from each other.

llms/mlx_lm/utils.py
```diff
@@ -71,7 +71,7 @@ def get_model_path(path_or_hf_repo: str) -> Path:
 
 
 def generate_step(
-    prompt: mx.array, model: nn.Module, temp: float = 0.0
+    prompt: mx.array, model: nn.Module, temp: float = 0.0, colorize: bool = False
```
awni, line 74: Can you change the name of this parameter to `return_probability`? It might be useful for other things and it's more clear what it does in that case.

awni (commented):
This is a nice feature, thanks for adding it! 

I left one comment, could you please check it?

Other than that I am happy to merge it as is. Just thinking though this would be a lot more general if `generate` took a print callback which takes as input a token and a probability. Gives the caller a lot more control over the output. We could refactor to that in the future if it is common to want to modify the output.

llms/mlx_lm/utils.py
ivanfioravanti: Should we change the --colorize command line argument too? 
I like the idea of giving more control over the output in future, if this is used 🤷‍♂️. 
People may like to change thresholds or colors (being colorblind this makes sense to me).

awni: I think it's ok the way you have it. That command line really does colorize so it's aptly named. If we eventually make the output more flexible via a callback and want to expose more option to the CLI we can change it then.

awni (approved):
This is really nice thanks for the feature!!

PR #351, opened 2024-01-22
feat(lora): add de-quantized support for fuse.py

mzbac:
Quick and dirty implementation for supporting de-quant model during fusing. Also, updated the model weight to use HF format so that once the model is fused, it can be converted by gguf for more advanced quant support. 

PS: Once the Lora example is merged into mlx-lm, we can properly add this functionality and make the fused model can be directly loaded by the transformers library to support AWQ/GPTQ.

### DIFF
--- a/lora/fuse.py
+++ b/lora/fuse.py
@@ -4,6 +4,7 @@ import argparse
 from pathlib import Path
 
 import mlx.core as mx
+import mlx.nn as nn
 import utils
 from mlx.utils import tree_flatten, tree_unflatten
 from models.lora import LoRALinear
@@ -41,6 +42,12 @@ if __name__ == "__main__":
         type=str,
         default=None,
     )
+    parser.add_argument(
+        "-d",
+        "--de-quantize",
+        help="Generate a de-quantized model.",
+        action="store_true",
+    )
 
     print("Loading pretrained model")
     args = parser.parse_args()
@@ -53,7 +60,7 @@ if __name__ == "__main__":
 
     # Freeze all layers other than LORA linears
     model.freeze()
-    for l in model.model.layers[-lora_layers:]:
+    for l in model.model.layers[len(model.model.layers) - lora_layers :]:
         l.self_attn.q_proj = LoRALinear.from_linear(l.self_attn.q_proj)
         l.self_attn.v_proj = LoRALinear.from_linear(l.self_attn.v_proj)
         if hasattr(l, "block_sparse_moe"):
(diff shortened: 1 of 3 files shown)


lora/utils.py
```diff
@@ -112,15 +112,27 @@ def make_shards(weights: dict, max_file_size_gibibyte: int = 15):
...
-def save_model(save_dir: str, weights, tokenizer, config):
+def save_model(save_dir: str, weights, tokenizer, config, de_quantize: bool = False):
     save_dir = Path(save_dir)
     save_dir.mkdir(parents=True, exist_ok=True)
-    shards = make_shards(weights)
+
+    shards = make_shards(weights, max_file_size_gibibyte=5)
+    shards_count = len(shards)
+    shard_file_format = (
+        "model-{:05d}-of-{:05d}.safetensors"
```
awni, line 122: It was a bad idea to not use this format in `mlx-lm` as well. We should update it there as well

lora/utils.py
```diff
@@ -112,15 +112,27 @@ def make_shards(weights: dict, max_file_size_gibibyte: int = 15):
     return shards
 
 
-def save_model(save_dir: str, weights, tokenizer, config):
+def save_model(save_dir: str, weights, tokenizer, config, de_quantize: bool = False):
     save_dir = Path(save_dir)
     save_dir.mkdir(parents=True, exist_ok=True)
-    shards = make_shards(weights)
+
+    shards = make_shards(weights, max_file_size_gibibyte=5)
```
awni, line 119: Curious, why did you change the max size here?

awni (commented):
Looks great, thanks for adding that option! That will be super useful! Just a minor comment. Could you address and then we can merge?

lora/utils.py
mzbac: I updated it in lora PR https://github.com/ml-explore/mlx-examples/pull/337

lora/utils.py
mzbac: > Curious, why did you change the max size here?

just try to it easier to upload. Otherwise, uploading 15GB takes too long and if anything happens during the upload, users need to re-upload. and, 5GB seems to be the most common size in the Huggingface repo.

lora/fuse.py
```diff
@@ -67,8 +74,33 @@
...
+                    linear.bias = m.bias
+                de_quantize_layers.append((n, linear))
+
+        model.update_modules(tree_unflatten(de_quantize_layers))
+
     weights = dict(tree_flatten(model.parameters()))
-    utils.save_model(args.save_path, weights, tokenizer, config)
+    utils.save_model(
+        args.save_path, weights, tokenizer, config, de_quantize=args.de_quantize
+    )
```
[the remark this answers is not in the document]
mzbac: updated: )

awni (approved):
Thank you!

Issue #348, opened 2024-01-21
Lora example's `convert.py` throws `TypeError` because of `tuple[mx.array, mx.array]`


adhulipa:
#### Description

I ran the following and observed an error

1. Created a new condo env
```
conda create -q -n mlx-examples python=3.8
conda activate mlx-examples
```
2. Ran pip install from lora dir `pip install -r requirements.txt` (from `..../mlx-examples/lora`)
3. Ran `convert.py` as `convert.py --hf-path mistralai/Mistral-7B-v0.1 -q`
4. Observed the error

```
↪ python convert.py --hf-path mistralai/Mistral-7B-v0.1 -q                                                                                                                                           (mlx-examples)

Traceback (most recent call last):
  File "convert.py", line 8, in <module>
    import utils
  File "/Users/aditya_d/projects/mlx-examples/lora/utils.py", line 13, in <module>
    import models.phi2 as phi2
  File "/Users/aditya_d/projects/mlx-examples/lora/models/phi2.py", line 120, in <module>
    class Model(nn.Module):
  File "/Users/aditya_d/projects/mlx-examples/lora/models/phi2.py", line 131, in Model
    ) -> tuple[mx.array, mx.array]:
TypeError: 'type' object is not subscriptable
```

adhulipa:
Is this as simple as editing the type hint to `    ) -> Tuple[mx.array, mx.array]:` ??

adhulipa:
Seems like I need to use python 3.9 or newer? https://stackoverflow.com/questions/41535881/how-do-i-upgrade-to-python-3-6-with-conda

PR #352, opened 2024-01-22
chore(mlx-lm): bump the mlx-lm python version requirement

mzbac:
To avoid the error https://github.com/ml-explore/mlx-examples/issues/348 after merging the lora example into mlx-lm.

### DIFF
--- a/llms/setup.py
+++ b/llms/setup.py
@@ -19,5 +19,5 @@ setup(
     license="MIT",
     install_requires=requirements,
     packages=["mlx_lm", "mlx_lm.models"],
-    python_requires=">=3.8",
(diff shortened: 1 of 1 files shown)


awni:
I think it would be nice to support > 3.8 since that's what we support with MLX as well. It should be an easy fix to replace the `tuple[...]` with `from typing import Tuple`  and `Tuple[...]`

mzbac:
> I think it would be nice to support > 3.8 since that's what we support with MLX as well. It should be an easy fix to replace the `tuple[...]` with `from typing import Tuple` and `Tuple[...]`

sounds good, I will update the tuple in lora PR.

PR #354, opened 2024-01-23
Update docs with `conda` install option

sugatoray:
Add conda installation option to the docs of `mlx-ml`.


awni (approved):
Awesome! I love that this is on conda forge as well!

Issue #346, opened 2024-01-20
Fused & Uploaded Model Losing Coherence


USMCM1A1:
I noticed today that when I use python -m mlx_lm.generate the output doesn't match what I get locally using python lora.py.  For example:
Local output using lora adapters:
```
(base) Williams-MacBook-Pro-3:lora williammarcellino$ python lora.py --model /Users/williammarcellino/mlx-examples/lora/mlx_model \
>                --adapter-file /Users/williammarcellino/mlx-examples/lora/adapters.npz \
>                --max-tokens 250 \
>                --temp .7 \
>                --prompt "
> Q: How does Aristotle define the soul?
> A: "
Loading pretrained model
Total parameters 7243.436M
Trainable parameters 1.704M
Loading datasets
Generating

Q: How does Aristotle define the soul?
A: 

Aristotle defines the soul as "the first act of a living body, which has in itself the faculty of movement and sensation."

This definition encompasses two key elements: 

1)   The soul is 
the first act or principle of life in a living being. 

2)  The soul is 
the faculties of movement and sensation in a living being. 

Let's explore these two aspects further:

1. The Soul as the First Act:

Aristotle views the soul as the first and most fundamental aspect of a living being. 
He argues that the soul is the source of life, and that it is responsible for all 
the activities and functions of a living organism. 

In this sense, the soul is not a separate entity from the body, but rather 
the animating principle that gives rise to life in a living being. 

2. The Soul as the Faculties of Movement and Sensation:

Aristotle defines the soul as the faculties of movement and sensation in a 
living being.
```
Whereas the same prompt and arguments produces a pretty weird, incoherent response:
```
(base) Williams-MacBook-Pro-3:mlx_lm williammarcellino$ python -m mlx_lm.generate \
>                --max-tokens 250 \
>                --temp .7 \
> --prompt "
> > Q: How does Aristotle define the soul?
> > A: " \
> --model mlx-community/mistral-7b-v0.2-GreeceRome-v0
Fetching 6 files: 100%|██████████████████████████████████████████████████████████████| 6/6 [00:00<00:00, 21290.88it/s]
==========
Prompt: 
> Q: How does Aristotle define the soul?
> A: 


Aristotle defines the soul as the first principle and cause of the body's life-giving activities. 

He considers the soul to be a substance that is inherent in every living body and is the actuality of all its processes. 

The soul is the form and final cause of the organism, and the efficient cause of its operations and growth. 

[...]

PR #360, opened 2024-01-24
chore(mlx-lm): add load model with adapter and fix bug in sample

mzbac:
As an alternative, when the fused model has slightly different text generation due to de-quantization, users can choose to load the adapter instead.  https://github.com/ml-explore/mlx-examples/issues/346

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -99,7 +99,7 @@ def generate_step(
     """
 
     def sample(logits: mx.array) -> Tuple[mx.array, float]:
-        softmax_logits = mx.softmax(logits / temp if temp > 0 else logits)
+        softmax_logits = mx.softmax(logits)
 
         if temp == 0:
             token = mx.argmax(logits, axis=-1)
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -13,6 +13,7 @@ from transformers import AutoConfig, AutoTokenizer, PreTrainedTokenizer
 
 # Local imports
 from .models import llama, mixtral, phi2, plamo, qwen
+from .tuner.utils import apply_lora_layers
 
 # Constants
 MODEL_MAPPING = {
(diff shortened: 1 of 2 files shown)


llms/mlx_lm/utils.py
```diff
@@ -98,11 +99,14 @@ def generate_step(
     """
 
     def sample(logits: mx.array) -> Tuple[mx.array, float]:
+        softmax_logits = mx.softmax(logits / temp if temp > 0 else logits)
```
mzbac, line 102: @awni  This fixes the error `UnboundLocalError: local variable 'prob' referenced before assignment`.

awni: Oops, I did test it carefully enough. I think we may need unit tests for this package if it keeps expandinf

awni: Also I think this is how it was before, but I wonder about the logic of returning the plain softmax if `temp == 0` but the temp version otherwise. It seems strange that `temp = 1e-4` will give you much more confident probabilities than `temp = 0` (which is the most confident). It seems more reasonable to me to always ignore the temperature in the probability computation.

llms/mlx_lm/utils.py
```diff
@@ -246,8 +250,10 @@ def load(
         model_path (Path): The path or the huggingface repository to load the model from.
         tokenizer_config (dict, optional): Configuration parameters specifically for the tokenizer.
             Defaults to an empty dictionary.
+        adapter_file (str, optional): Path to the adapter file. If provided, applies LoRA layers to the model.
```
awni, line 253: 👍

awni (approved):
Thanks, and sorry for the bug! We can merge it but I'm curious on your thoughts on the right thing to return for the probability.

llms/mlx_lm/utils.py
mzbac: My understanding was that this was meant to show how the token is selected. My intuition was that whatever token is selected, we return its probability. but I didn't really understand why `temp = 1e-4` would give us the most confident probability. :(

awni: If you use a really small temperature the max probability gets closer and closer to 1. As temp -> 0 , the max probability is exactly 1 (aka argmax sampling). Hence the inconsistency. Yea I would prefer to just set `prob = mx.softmax(logits)[token]` in all cases as I think it is more consistent / decoupled from how you sample.

awni: Would you mind making that change and then we can merge?

mzbac: Yeah, that makes sense. Sorry, I misread the `1e-4` as `1e4`. I will update it now.

awni (approved):
Thank you!

PR #361, opened 2024-01-24
change the verbose effect

iLoveBug:
I set the verbose default as "false", and keep the generate function always print the tokens. the verbose argument only enable the output of prompt and timing information.


awni:
I'm curious, why do you want verbose `False` with the CLI? I'm not against it just wondering why it's important?

llms/mlx_lm/utils.py
```diff
@@ -157,19 +158,19 @@ def generate(
...
+        s = tokenizer.decode(tokens)
+        if REPLACEMENT_CHAR not in s:
             if formatter:
                 formatter(s[skip:], prob.item())
-                skip = len(s)
-            elif REPLACEMENT_CHAR not in s:
+            else:
                 print(s[skip:], end="", flush=True)
-                skip = len(s)
+            skip = len(s)
```
awni, line 167: The formatter expects one token / prob, so this will break in some cases.

llms/mlx_lm/utils.py
```diff
@@ -157,19 +158,19 @@ def generate(
...
 
     tokens = tokenizer.decode(tokens).replace(REPLACEMENT_CHAR, "")
 
+    print(tokens[skip:], flush=True)
```
awni, line 171: We only want to print this in verbose mode.

llms/mlx_lm/utils.py
awni, line 167: We don't want to print here in general. The API shouldn't print anything when `verbose=False`

iLoveBug:
OK，I understood. Thanks for your review.

PR #363, opened 2024-01-24
add safetensors option for mistral weight conversion

davidkoski:
- add an option to write and read weights in safetensor format
- in theory these weights load faster, though in mesaurement it seems to be less than 50ms difference


awni:
@davidkoski maybe we close this given our conversation re using the native HF format?

PR #364, opened 2024-01-24
Uplaod train loss data to wandb

madroidmaq:
Use it as follows:

```
python lora.py --model <path_to_model> \
               --train \
               --wandb-log \
               --iters 600
``` 

When I use it, the training data generated on the wandb platform is as follows

image


awni:
Thanks! This is a cool feature but it doesn't make sense for this example. The LoRA example is supposed to be pretty light weight. easy to read and easy to hack for different purposes (as you've done!). Some people may want to use other visualization servers etc.

PR #365, opened 2024-01-24
feat(mlx-lm): add de-quant for fuse.py

mzbac:
Add de-quant support for mlx-lm fuse.py, manually tested locally.

### DIFF
--- a/llms/mlx_lm/fuse.py
+++ b/llms/mlx_lm/fuse.py
@@ -8,7 +8,7 @@ from typing import Any, Dict, Union
 from mlx.utils import tree_flatten, tree_unflatten
 
 from .tuner.lora import LoRALinear
-from .tuner.utils import apply_lora_layers
+from .tuner.utils import apply_lora_layers, dequantize
 from .utils import fetch_from_hub, get_model_path, save_weights, upload_to_hub
 
 
@@ -42,6 +42,11 @@ def parse_arguments() -> argparse.Namespace:
         type=str,
         default=None,
     )
+    parser.add_argument(
+        "--de-quantize",
+        help="Generate a de-quantized model.",
+        action="store_true",
+    )
     return parser.parse_args()
 
 
@@ -54,6 +59,7 @@ def main() -> None:
 
     model.freeze()
     model = apply_lora_layers(model, args.adapter_file)
+
     fused_linears = [
         (n, m.to_linear())
         for n, m in model.named_modules()
@@ -61,6 +67,11 @@ def main() -> None:
     ]
 
     model.update_modules(tree_unflatten(fused_linears))
+
+    if args.de_quantize:
+        print("De-quantizing model")
+        model = dequantize(model)
+
     weights = dict(tree_flatten(model.parameters()))
 
     save_path = Path(args.save_path)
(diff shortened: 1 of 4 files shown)


llms/mlx_lm/tuner/utils.py
```diff
@@ -6,18 +6,60 @@
...
         [name.replace(".lora_a", "").replace(".lora_b", "") for name, _ in adapters]
     )
-
     for name, module in model.named_modules():
         if name in lora_layers:
             replacement_module = LoRALinear.from_linear(module)
-            linear_replacements[name] = replacement_module
+            linear_replacements.append((name, replacement_module))
+
+    model.update_modules(tree_unflatten(linear_replacements))
```
mzbac, line 31: Refactor the code a bit to avoid unnecessary conversion of dict to list.

llms/mlx_lm/utils.py
```diff
@@ -24,7 +24,7 @@
     "qwen": qwen,
     "plamo": plamo,
 }
-MAX_FILE_SIZE_GB = 15
+MAX_FILE_SIZE_GB = 5
```
mzbac, line 27: Reduce the shard size to make it easier for uploading.

llms/mlx_lm/fuse.py
```diff
@@ -54,13 +59,19 @@ def main() -> None:
 
     model.freeze()
     model = apply_lora_layers(model, args.adapter_file)
+
     fused_linears = [
         (n, m.to_linear())
```
mzbac, line 64: TODO: disable quantization if de-quantize is true.

llms/mlx_lm/tuner/utils.py
```diff
@@ -6,18 +6,60 @@
...
+    Apply LoRA layers to the model.
+
+    Args:
+        model (nn.Module): The neural network model.
+        adapter_file (str): Path to the adapter configuration file.
+
+    Returns:
+        nn.Module: The updated model with LoRA layers applied.
+    """
+    print("load adapters :", adapter_file)
```
awni, line 19: Nit: is this print here intentionally or debugging? Probably good to avoid printing in the API, we could have a logging in the future if it's useful.

awni (commented):
LGTM, just one nit then we can merge it. Thanks for adding this!!

llms/mlx_lm/tuner/utils.py
mzbac: @awni This was intentional because I feel load adapter does not have any output to indicate whether it was successful or not. However, I am happy to remove it and use proper logging later. P.S.: I constantly make mistakes by misspelling adapters as "adapter." It's kind of helpful for troubleshooting.

awni: I see, is the below `mx.load` error if it can't find the file not useful for that? We could do a check for that instead (e.g. check if the raise error with helpful message if not?

mzbac: > I see, is the below `mx.load` error if it can't find the file not useful for that? We could do a check for that instead (e.g. check if the raise error with helpful message if not?

The mx.load error is not a very user-friendly error message. If no file is found, it just throws `ValueError: [load_npz] Input must be a zip file or a file-like object that can be opened with zipfile.ZipFile`.

awni: Oh that's so useless!!! We should fix that indeed

mzbac: @awni I removed the print adapter file path and added a file exist check before mx.load.

mzbac:
@awni Please let me know if I missed anything else so that we can merge this PR.

awni (approved):
This is great! Thank you!

PR #366, opened 2024-01-24
feat(mlx-lm): add lora hypeparameters in lora layer

mzbac:
Add LoRA hyperparameters based on the https://github.com/microsoft/LoRA/blob/main/loralib/layers.py with slight modifications to support advanced LoRA fine-tuning.

### DIFF
--- a/llms/mlx_lm/tuner/lora.py
+++ b/llms/mlx_lm/tuner/lora.py
@@ -16,7 +16,6 @@ class LoRALinear(nn.Module):
         # TODO remove when input_dims and output_dims are attributes
         # on linear and quantized linear
         output_dims, input_dims = linear.weight.shape
-        bias = "bias" in linear
         if isinstance(linear, nn.QuantizedLinear):
             input_dims *= 32 // linear.bits
         lora_lin = LoRALinear(
@@ -26,7 +25,6 @@ class LoRALinear(nn.Module):
             lora_alpha=lora_alpha,
             lora_dropout=lora_dropout,
             scale=scale,
-            bias=bias,
         )
         lora_lin.linear = linear
         return lora_lin
@@ -82,10 +80,7 @@ class LoRALinear(nn.Module):
         # Regular linear layer weights
         self.linear = nn.Linear(input_dims, output_dims, bias=bias)
 
-        if lora_dropout > 0.0:
-            self.lora_dropout = nn.Dropout(p=lora_dropout)
-        else:
-            self.lora_dropout = lambda x: x
+        self.lora_dropout = nn.Dropout(p=lora_dropout)
 
         # Scale for low-rank update
         self.scale = scale * (lora_alpha / r)
@@ -104,7 +99,5 @@ class LoRALinear(nn.Module):
         if isinstance(self.linear, nn.QuantizedLinear):
             dtype = self.linear.scales.dtype
         y = self.linear(x.astype(dtype))
-        if self.training:
-            x = self.lora_dropout(x)
-        z = (x @ self.lora_a) @ self.lora_b
+        z = (self.lora_dropout(x) @ self.lora_a) @ self.lora_b
         return y + self.scale * z
--- a/llms/mlx_lm/tuner/lora.py
+++ b/llms/mlx_lm/tuner/lora.py
@@ -6,14 +6,27 @@ import mlx.nn as nn
 
 class LoRALinear(nn.Module):
     @staticmethod
-    def from_linear(linear: nn.Linear, rank: int = 8, scale: float = 20.0):
+    def from_linear(
+        linear: nn.Linear,
+        r: int = 8,
+        lora_alpha: float = 16,
+        lora_dropout: float = 0.05,
+        scale: float = 10.0,
+    ):
         # TODO remove when input_dims and output_dims are attributes
         # on linear and quantized linear
         output_dims, input_dims = linear.weight.shape
+        bias = "bias" in linear
         if isinstance(linear, nn.QuantizedLinear):
             input_dims *= 32 // linear.bits
         lora_lin = LoRALinear(
-            input_dims=input_dims, output_dims=output_dims, rank=rank, scale=scale
+            input_dims=input_dims,
+            output_dims=output_dims,
+            r=r,
+            lora_alpha=lora_alpha,
+            lora_dropout=lora_dropout,
+            scale=scale,
+            bias=bias,
         )
         lora_lin.linear = linear
         return lora_lin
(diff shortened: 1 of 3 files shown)


llms/mlx_lm/tuner/lora.py
```diff
@@ -58,31 +71,40 @@ def __init__(
...
     ):
         super().__init__()
 
         # Regular linear layer weights
         self.linear = nn.Linear(input_dims, output_dims, bias=bias)
 
+        if lora_dropout > 0.0:
+            self.lora_dropout = nn.Dropout(p=lora_dropout)
+        else:
+            self.lora_dropout = lambda x: x
```
awni, line 88: Fwiw this is exactly what `Dropout.__call__` does if you give it a `0.0` value, so it's not really necessary. I don't mind if we keep it though if you like to be explicit :)

llms/mlx_lm/tuner/lora.py
```diff
@@ -58,31 +71,40 @@ def __init__(
         self,
         input_dims: int,
         output_dims: int,
-        rank: int = 8,
+        r: int = 8,
+        lora_alpha: float = 16,
+        lora_dropout: float = 0.0,
+        scale: float = 10.0,
```
awni, line 77: 👍 really nice to be able to configure these.

llms/mlx_lm/tuner/lora.py
```diff
@@ -58,31 +71,40 @@ def __init__(
...
         )
-        self.lora_b = mx.zeros(shape=(rank, output_dims))
+        self.lora_b = mx.zeros(shape=(r, output_dims))
 
     def __call__(self, x):
         dtype = self.linear.weight.dtype
         if isinstance(self.linear, nn.QuantizedLinear):
             dtype = self.linear.scales.dtype
         y = self.linear(x.astype(dtype))
+        if self.training:
```
awni, line 107: This should not be necessary, dropout checks `self.training` hence the reason to use `Module.train()` and `Module.eval()`

llms/mlx_lm/tuner/lora.py
```diff
@@ -6,14 +6,27 @@
...
-            input_dims=input_dims, output_dims=output_dims, rank=rank, scale=scale
+            input_dims=input_dims,
+            output_dims=output_dims,
+            r=r,
+            lora_alpha=lora_alpha,
+            lora_dropout=lora_dropout,
+            scale=scale,
+            bias=bias,
         )
         lora_lin.linear = linear
```
awni, line 31: You don't really need to pass `bias` here because it will always be overwritten by the next call to set the linear layer.

The `kwarg` is there only so people have the option to instantiate a `LoraLinear` directly (with or without a bias) rather than building from a pre-existing `Linear` layer

llms/mlx_lm/tuner/lora.py
mzbac: Does the dropout have a boundary check for p in the range of 0.0 to 1.0? I copied and pasted this from the original Lora implementation code. If mlx dropout has a boundary check for p, I don't see any reason why we would need to do this check.

awni (commented):
Thanks! This looks great, really nice to take steps to configure those params.. I left a few minor comments, please check, and then we can merge 🚀

llms/mlx_lm/tuner/lora.py
awni: Indeed it does:
https://github.com/ml-explore/mlx/blob/main/python/mlx/nn/layers/dropout.py#L21-L22

mzbac:
@awni  addressed all the comments. Thank you for helping with the review :)

awni (approved):
🚀

gladjoyhub:
@mzbac
I saw the new feature of load adapter_file:
model, tokenizer = load(
    "mlx-community/phi-2-hf-4bit-mlx", adapter_file="path_to_your_adapters.npz"
)

But how to change adapter_file without reloading base model to save loading time?

mzbac:
> @mzbac I saw the new feature of load adapter_file: model, tokenizer = load( "mlx-community/phi-2-hf-4bit-mlx", adapter_file="path_to_your_adapters.npz" )
> 
> But how to change adapter_file without reloading base model to save loading time?

Currently, we do not have the method to do that, but it should be possible. For example, before loading the Lora layers, we can check if the model already has any Lora layers. If there are Lora layers in the model, we can convert them back to linear layers and then reload the new set of Lora layers.

gladjoyhub:
> > @mzbac I saw the new feature of load adapter_file: model, tokenizer = load( "mlx-community/phi-2-hf-4bit-mlx", adapter_file="path_to_your_adapters.npz" )
> > But how to change adapter_file without reloading base model to save loading time?
> 
> Currently, we do not have the method to do that, but it should be possible. For example, before loading the Lora layers, we can check if the model already has any Lora layers. If there are Lora layers in the model, we can convert them back to linear layers and then reload the new set of Lora layers.

Yes, like a tool replace_lora_layers(model, new_adapter_file), which uses
model.update_modules(tree_unflatten(list(linear_replacements.items())))
model.update(tree_unflatten(adapters))

But I don't know how to convert lora layers back to linear layers.

mzbac:
> > > @mzbac I saw the new feature of load adapter_file: model, tokenizer = load( "mlx-community/phi-2-hf-4bit-mlx", adapter_file="path_to_your_adapters.npz" )
> > > But how to change adapter_file without reloading base model to save loading time?
> > 
> > 
> > Currently, we do not have the method to do that, but it should be possible. For example, before loading the Lora layers, we can check if the model already has any Lora layers. If there are Lora layers in the model, we can convert them back to linear layers and then reload the new set of Lora layers.
> 
> Yes, like a tool replace_lora_layers(model, new_adapter_file), which uses model.update_modules(tree_unflatten(list(linear_replacements.items()))) model.update(tree_unflatten(adapters))
> 
> But I don't know how to convert lora layers back to linear layers.

The Lora has original linear layer as a property, so you should be able to use that to recover the original model.

gladjoyhub:
> > > > @mzbac I saw the new feature of load adapter_file: model, tokenizer = load( "mlx-community/phi-2-hf-4bit-mlx", adapter_file="path_to_your_adapters.npz" )
> > > > But how to change adapter_file without reloading base model to save loading time?
> > > 
> > > 
> > > Currently, we do not have the method to do that, but it should be possible. For example, before loading the Lora layers, we can check if the model already has any Lora layers. If there are Lora layers in the model, we can convert them back to linear layers and then reload the new set of Lora layers.
> > 
> > 
> > Yes, like a tool replace_lora_layers(model, new_adapter_file), which uses model.update_modules(tree_unflatten(list(linear_replacements.items()))) model.update(tree_unflatten(adapters))
> > But I don't know how to convert lora layers back to linear layers.
> 
> The Lora has original linear layer as a property, so you should be able to use that to recover the original model.

If names don't change in lora_layers, then I don't need to do again: model.update_modules(tree_unflatten(list(linear_replacements.items()))),

I can just do
model.update(tree_unflatten(adapters)) 
to change the weights for the new adapter

Is my understanding correct?

Issue #367, opened 2024-01-24
mlx-lm 0.0.4 - ValueError: Missing parameters: lm_head.biases lm_head.scales


da-z:
Hi guys.

[mlx-ui](https://github.com/da-z/mlx-ui) users are getting this error after bumping from 0.0.3 to 0.0.4, while 0.0.3 continues to work fine, same models.

```sh
File "/Users/x/dev/mlx-ui/venv/lib/python3.11/site-packages/mlx_lm/utils.py", line 232, in load
    model = load_model(model_path)
            ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/x/dev/mlx-ui/venv/lib/python3.11/site-packages/mlx_lm/utils.py", line 205, in load_model
    model.load_weights(list(weights.items()))
  File "/Users/x/dev/mlx-ui/venv/lib/python3.11/site-packages/mlx/nn/layers/base.py", line 151, in load_weights
    raise ValueError(f"Missing parameters: {missing}.")
ValueError: Missing parameters: lm_head.biases lm_head.scales.
```

Is there anything I can do on mlx-ui side?

awni:
Sorry for that, I will take a look. Is it with the NousHermes model?

da-z:
No worries and no rush 😊 (I am using cached requirements.txt from repo now)

I've reproduced with 2 models:

```
mlx-community/dolphin-2.6-mistral-7b-dpo-laser-4bit-mlx
mlx-community/Nous-Hermes-2-Mixtral-8x7B-DPO-4bit
```

awni:
Hmm, I see the problem. We updated MLX to quantize non 32 multiples. So now we don't by default assume those layers are not quantized.

awni:
It's a problem of back compatibility. It's messy to support those old models while also supporting new models that are fully quantized. 

One option is to make a little script that can download, quantize the linear layer, and then re-upload 🤔 

@mzbac what do you think?

mzbac:
> It's a problem of back compatibility. It's messy to support those old models while also supporting new models that are fully quantized.
> 
> One option is to make a little script that can download, quantize the linear layer, and then re-upload 🤔
> 
> @mzbac what do you think?

Non-32-dimensional quantization shouldn't cause the issue. We still have that `m.weight.shape[0]!= 8`, so it may be caused by something else. I can take a look at those two models.

awni:
No that's just for gating :), this is for the LM head

[...]

PR #369, opened 2024-01-24
fix(mlx-lm): handle legacy quant models

mzbac:
Close https://github.com/ml-explore/mlx-examples/issues/367
cc @awni


awni (approved):
Nice patch! Thank you!!

PR #373, opened 2024-01-25
chore(mlx-lm): truncate the input sentence to max seq len in lora iterate_batches

mzbac:
The current iterate_batches function will return a batch with the maximum length that the training set sentence has. This can cause a memory spike when the training dataset contains long sentences, which may crash the training process.

### DIFF
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -175,6 +175,7 @@ def train(
                 tokenizer=tokenizer,
                 batch_size=args.batch_size,
                 num_batches=args.val_batches,
+                max_seq_length=args.max_seq_length,
             )
             print(
                 f"Iter {it + 1}: "
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -60,7 +60,7 @@ def iterate_batches(dataset, tokenizer, batch_size, max_seq_length, train=False)
         indices = np.random.permutation(indices)
         # Collect batches from dataset
         for i in range(0, len(indices) - batch_size + 1, batch_size):
-            # Encode batch
+            # Encode batch up to the max seq length
             batch = [
                 tokenizer.encode(dataset[indices[i + j]])[:max_seq_length]
                 for j in range(batch_size)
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/tuner/trainer.py
```diff
@@ -60,25 +60,21 @@ def iterate_batches(dataset, tokenizer, batch_size, max_seq_length, train=False)
         indices = np.random.permutation(indices)
         # Collect batches from dataset
         for i in range(0, len(indices) - batch_size + 1, batch_size):
-            # Encode batch
+            # Encode batch up to the max seq length
             batch = [
-                tokenizer.encode(dataset[indices[i + j]]) for j in range(batch_size)
+                tokenizer.encode(dataset[indices[i + j]])[:max_seq_length]
```
mzbac, line 65: @awni I noticed this when I tried to qlora mixtral using the guanaco dataset. The memory spiked and crashed the training, after I changed the iterate_batches to only return a specified max seq length, the training worked as expected.

llms/mlx_lm/tuner/trainer.py
```diff
@@ -175,6 +171,7 @@ def train(
                 tokenizer=tokenizer,
                 batch_size=args.batch_size,
                 num_batches=args.val_batches,
+                max_seq_length=args.max_seq_length,
```
awni, line 174: Did you add a new argument? I don't see it in the diff?

awni: Oh I see it in the trainer args now, nvm

llms/mlx_lm/tuner/trainer.py
awni: Cool. I would suggest we keep a warning below because some people may have really long paragraphs and the right way to deal with that is not to truncate them here but to split them in a preprocessing step.

So first:
1. Encode full length
2. Warn if length is too long ("longest sentence {max(lengths)} will be truncated to {max_seq_length}"
3. Truncate and make an array batch

mzbac: @awni I updated it based on the suggestion. My current Mac studio is running the training. I didn't do comprehensive testing, but from a quick test on my laptop it looks good.

awni (approved):
Thank you!!

PR #380, opened 2024-01-26
Use better time for measuring tokens/second 🚀 

ashishdatta:
I purpose to change all usage of `time.time()` to [`time.perf_counter()`](https://docs.python.org/3/library/time.html#time.perf_counter) in `utils.py` for MLX-LM

`time.perf_counter` is more preferable when measuring and benchmarking since it is monotonic and higher precision clock compared to `time.time()`

Rough benchmarks with StableLM-2-Zephyr 1.6B model with this change

with time.time()
Prompt: 44.321 tokens-per-sec
Generation: 113.937 tokens-per-sec

with time.perf_counter()
Prompt: 192.664 tokens-per-sec
Generation: 117.253 tokens-per-sec

~~+300% on tokens/s for prompt~~ <- My bad 
+3% for tokens/s for generate

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -149,7 +149,7 @@ def generate(
 
     prompt = mx.array(tokenizer.encode(prompt))
 
-    tic = time.time()
+    tic = time.perf_counter()
     tokens = []
     skip = 0
     REPLACEMENT_CHAR = "\ufffd"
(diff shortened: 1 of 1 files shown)


awni:
The timing change is almost certainly not due to using `time.perf_counter`. Our prompt time is just not that accurate because it doesn't do a warmup. So if you run it the first time it will be really slow (needs to load up a bunch of kernels, allocate memory, etc) and the second time it will be really fast 😄 .

Probably you did the timings in succession? If you switch the order (`perf_counter` first) then you will likely see that `time.time` gives a much nicer result.

awni:
Actually, I'm really curious about this. How did you do the benchmark? It is very strange that it is so much faster.

ashishdatta:
I edited the comment because you are right I forgot to factor in the warmup for prompt generation, Friday things. But the generation performance "gain"  is still true regardless of order.

awni (approved):
Thanks!!

PR #386, opened 2024-01-29
StableLM 3B and Stable-code-3b Model Support

ashishdatta:
This PR adds support for StableLM 3B and by extension stable-code-3b. 

I am using vocabulary size of the tokenizer to infer the usage of QKV bias.

usage example:

python -m mlx_lm.generate \
--model stabilityai/stable-code-3b \
--max-tokens 500 \
--prompt "Write a quick sort in C++" \
--colorize

### DIFF
--- a/llms/mlx_lm/models/stablelm_epoch.py
+++ b/llms/mlx_lm/models/stablelm_epoch.py
@@ -20,7 +20,6 @@ class ModelArgs(BaseModelArgs):
     intermediate_size: int
     norm_eps: float
     rope_theta: float
-    use_qkv_bias: bool
 
 
 class LayerNorm(nn.LayerNorm):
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/models/stablelm_epoch.py
```diff
@@ -45,19 +44,20 @@ def __init__(self, config: ModelArgs):
                 f"hidden_size must be divisible by num_heads (got `hidden_size`: {self.hidden_size}"
                 f" and `num_heads`: {self.num_heads})."
             )
+        self.use_qkv_bias = True if config.vocab_size > 50304 else False
```
awni, line 47: Is this available in the config?

ashishdatta: I added the same key to the HuggingFace config.json so now we don't need to use the hard coded value.

awni:
Maybe I miss something but it looks like the only change is the commented out change?

awni:
Is github acting up or did your change somehow disappear?

PR #387, opened 2024-01-29
Add grad checkpointing and PE in the transformer example

angeloskath:
This provides an example use of gradient checkpointing in the `transformer_lm` example and also adds positional encoding.


transformer_lm/jax_main.py
```diff
@@ -9,7 +9,7 @@
 import jax
 import jax.numpy as jnp
 import numpy as np
-from tree_utils import tree_flatten
+from mlx.utils import tree_flatten
```
awni, line 13: Nit: There is a comment below that lists TODOs, maybe we can remove that (or at least the posenc bit)

transformer_lm/main.py
```diff
@@ -67,7 +79,9 @@ def main(args):
     vocab, train, valid, test = datasets.load_dataset(args.dataset)
 
     # Initialize model:
-    model = TransformerLM(len(vocab), args.num_blocks, args.dim, args.num_heads)
+    model = TransformerLM(
+        len(vocab), args.num_blocks, args.dim, args.num_heads, args.checkpoint
+    )
```
awni, line 84: Should we also do a `del grads` below? Seems like it could be good to include that as a best practice.

awni (approved):
LGTM!! I think you were planning to move the other framework examples in this PR.. if so feel free to do so o/w we can land this after the checkpoint PR gets merged.

One more request, could you add a `requirements.txt` with the right version of `mlx`:

```
mlx>=0.1
```

Issue #356, opened 2024-01-23
[Feature Request] Add Stable LM 2 1.6B Lora or QLora example


madroidmaq:
Recently, Stability.AI released a 1.6B small model, which is of great significance for mobile deployment. It is expected that examples of corresponding fine-tuning (Lora&QLora) can be added.

When I try to convert the `stabilityai/stablelm-2-zephyr-1_6b` model, I encountered a problem with `ValueError: Model type stablelm_epoch not supported.`， the details are as follows

```
python convert.py --hf-path stabilityai/stablelm-2-zephyr-1_6b -q
[INFO] Loading
Fetching 5 files: 100%|███████████████████████████████████████████████████████████| 5/5 [00:00<00:00,  8.06it/s]
The repository for stabilityai/stablelm-2-zephyr-1_6b contains custom code which must be executed to correctly load the model. You can inspect the repository content at https://hf.co/stabilityai/stablelm-2-zephyr-1_6b.
You can avoid this prompt in future by passing the argument `trust_remote_code=True`.

Do you wish to run the custom code? [y/N] y
A new version of the following files was downloaded from https://huggingface.co/stabilityai/stablelm-2-zephyr-1_6b:
- configuration_stablelm_epoch.py
. Make sure to double-check they do not contain any added malicious code. To avoid downloading new versions of the code file, you can pin a revision.
The repository for stabilityai/stablelm-2-zephyr-1_6b contains custom code which must be executed to correctly load the model. You can inspect the repository content at https://hf.co/stabilityai/stablelm-2-zephyr-1_6b.
You can avoid this prompt in future by passing the argument `trust_remote_code=True`.

Do you wish to run the custom code? [y/N] y
tokenization_arcade100k.py: 100%|██████████████████████████████████████████| 9.89k/9.89k [00:00<00:00, 10.9MB/s]
A new version of the following files was downloaded from https://huggingface.co/stabilityai/stablelm-2-zephyr-1_6b:
- tokenization_arcade100k.py
. Make sure to double-check they do not contain any added malicious code. To avoid downloading new versions of the code file, you can pin a revision.
[INFO] Quantizing
ERROR:root:Model type stablelm_epoch not supported.
Traceback (most recent call last):
  File "/Users/madroid/Desktop/mlx-examples/lora/convert.py", line 97, in <module>
    weights, config = quantize(weights, config, args)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #388, opened 2024-01-30
LoRA: Remove unnecessary model type judgments

madroidmaq:
1. Supported models are already checked in the load_model function in utils, no need to repeat the check in lora
2. The checks in lora are not synchronized with those in utils

After deleting this part of the judgment, the Lora module will support the following model fine-tuning:
- stablelm_epoch
- qwen
- plamo

see [MODEL_MAPPING](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/utils.py#L19-L27) in mlx_lm utils.py .

Also solves issue #356

### DIFF
--- a/llms/mlx_lm/lora.py
+++ b/llms/mlx_lm/lora.py
@@ -12,9 +12,6 @@ from .tuner.lora import LoRALinear
 from .tuner.trainer import TrainingArgs, evaluate, train
 from .utils import generate, load
 
-SUPPORTED_MODELS = [llama.Model, mixtral.Model, phi2.Model]
-
-
 def build_parser():
     parser = argparse.ArgumentParser(description="LoRA or QLoRA finetuning.")
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


mzbac:
The reason the check is there is because the qwen has a different attention implementation, so the Lora layers default to applying to q_proj and v_proj won't work. I think the best way may be to change the logic to exclude the qwen until we have a better way for users to specify which layers they want to apply Lora.

madroidmaq:
@mzbac Thank you for your reminder.

What I've come up with so far is a new LORA_SUPPORTED_MODELS list in utils.py, where you can add models for LoRA that are already supported in LORA_SUPPORTED_MODELS.

When others want to support new model reasoning, they can see the configuration with LoRA in the nearest place. The reverse is also true. Please let me know if there are other improvements that can be made.

mzbac:
That makes sense. Thanks for fixing this. Hopefully, we can have support for custom LoRa layers very soon.

awni (approved):
Looks great, thanks!

PR #391, opened 2024-01-30
Passing parameterized batching function to trainer

chimezie:
The mlx_lm.tuner module is very _re-usable_ and makes building other training approaches on top of it pretty straightforward.  However, being able to specify an **iterate_batches** function in the same way a loss function can be provided would make it much easier to build other training methods (completion only training, for example) on top of the current framework.

This PR adds the ability to pass a user-specified **loss** and **iterate_batches** function to the mlx_lm.tuner.trainer.train function


awni:
Wdyt @mzbac? Looks pretty ok to me..

mzbac:
Looks good to me, I also found there is a needs that user may want to just batch in sequence instead of random batch. 
Thanks @chimezie for adding this 🚀

awni (approved):
Nice and simple, thanks for adding this!

PR #392, opened 2024-01-30
fix token count computation -- produced wrong tps measurements

davidkoski:
PR #342 inadvertently counted the number of characters in the resulting string as the number of tokens.  This inflated the tokens per second count by roughly a factor of 3 (when producing 100 tokens).

Before:

```
python -m mlx_lm.generate --model ~/Mistral-7B-v0.1-hf-4bit-mlx -m 20
==========
Prompt: hello
…

I’ve had my 1366 Samsung 24″ monitor for
==========
Prompt: 15.634 tokens-per-sec
Generation: 60.364 tokens-per-sec
```

After:

```
python -m mlx_lm.generate --model ~/Mistral-7B-v0.1-hf-4bit-mlx -m 20
==========
Prompt: hello
…

I’ve had my 1366 Samsung 24″ monitor for
==========
Prompt: 16.123 tokens-per-sec
Generation: 27.702 tokens-per-sec
```

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -171,21 +171,22 @@ def generate(
                 print(s[skip:], end="", flush=True)
                 skip = len(s)
 
-    tokens = tokenizer.decode(tokens).replace(REPLACEMENT_CHAR, "")
+    token_count = len(tokens)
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
Thanks for the fix!

PR #393, opened 2024-01-30
Work with tuple shapes

awni:
MLX switches to use tuple for shape in Python in https://github.com/ml-explore/mlx/pull/591

I ran every example and I only found 2 which had issues: stable_diffusion and whisper. Also I really appreciated how easy our examples are to run in the process 😄 .


angeloskath (approved):
👍

PR #396, opened 2024-01-31
fix(mlx-lm): apply lora layer doesn't update the lora weights

mzbac:
@awni, I found this while working on the moe model stuff. I noticed that the lora weight doesn't apply to the gate and the weight distribution is almost identical to uniform. So, I dug a bit into the mlx-lm lora code and found out that we only replace the linear layer but never update the lora weight from adapters.


awni:
So this means `fuse` was broken right? I should do a patch release probably

awni (approved):
Thanks for debugging and fixing this!

mzbac:
> So this means `fuse` was broken right? I should do a patch release probably

Yeah, I believe it breaks the current fuse in mlx-lm. However, I don't think many people are using mlx-lm lora yet. so not sure if we need an urgent patch release or not.

danilopeixoto:
Thanks! This helped me.

PR #398, opened 2024-01-31
BUG FIX: Decoding results in garbled text when multiple tokens represent a single character (e.g., Chinese).

gklab:


lora/lora.py
```diff
@@ -292,8 +292,9 @@ def generate(model, prompt, tokenizer, args):
 
         tokens.append(token.item())
         s = tokenizer.decode(tokens)
-        print(s[skip:], end="", flush=True)
-        skip = len(s)
+        if len(s) - skip > 1:
+            print(s[skip:-1], end="", flush=True)
+            skip = len(s) - 1
```
awni, line 297: Is this in the same spirit as https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/utils.py#L173-L175 ?

awni: Or do we have this same problem in `mlx-lm`? If so would be good to add that fix there also!!

gklab: > /llms/mlx_lm/utils.py#L173-L175

Yes, it serves the same purpose as this approach.

awni (approved):
Looks great, thanks!

PR #399, opened 2024-01-31
Add Repetitive penalty to LLM inference - mlx-lm

### LINKED ISSUE #385: Repetition penalty for LLM inference in MLX LM
Implement `repetition_penalty` parameter for LLM inference with `mlx_lm.generate`.

**Paper**

[CTRL: A Conditional Transformer Language Model for Controllable Generation](https://arxiv.org/abs/1909.05858)

**Implementation example**

[huggingface/transformers: logits_process.py](https://github.com/huggingface/transformers/blob/f1cc6157210abd9ea953da097fb704a1dd778638/src/transformers/generation/logits_process.py#L288)

**Possible MLX implementation**

```python
# Generate step

# Apply repetition penalty
score = logits[:, input_ids]
score = mx.where(score < 0, score * repetition_penalty, score / repetition_penalty)
logits[:, input_ids] = score

# Apply temperature sampling
```

vishal-14069:
## Proposed changes

This PR adds Repetitive Penalty functionality to `mlx_lm/generate`

Repetition penalty is used during inference to reduce repetitive generations from the language model. 

Hugging Face implementation: [huggingface/transformers: logits_process.py](https://github.com/huggingface/transformers/blob/f1cc6157210abd9ea953da097fb704a1dd778638/src/transformers/generation/logits_process.py#L288)
Paper: https://arxiv.org/abs/1909.05858

Closes #385 
## Checklist

Put an `x` in the boxes that apply.

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the necessary documentation (if needed)


llms/mlx_lm/utils.py
```diff
@@ -82,10 +82,34 @@ def get_model_path(path_or_hf_repo: str) -> Path:
...
+        logits[:, indices] = selected_logits
+    return logits
+
+
 def generate_step(
     prompt: mx.array,
     model: nn.Module,
-    temp: float = 0.0,
+    temp: 0.0,
+    generated_tokens: Any,
```
mzbac, line 111: Maybe we can make the generated_tokens optional so we don't break the API contract and maintain backward compatibility.

llms/mlx_lm/utils.py
```diff
@@ -123,7 +156,8 @@ def generate(
     model: nn.Module,
     tokenizer: PreTrainedTokenizer,
     prompt: str,
-    temp: float = 0.0,
+    temp: float,
```
mzbac, line 159: Keep the `temp` and `repetition_penalty` optional to maintain backward compatibility.

llms/mlx_lm/utils.py
```diff
@@ -154,13 +189,16 @@ def generate(
     skip = 0
     REPLACEMENT_CHAR = "\ufffd"
 
-    for (token, prob), n in zip(generate_step(prompt, model, temp), range(max_tokens)):
-        if token == tokenizer.eos_token_id:
+    for (token, prob), n in zip(
```
mzbac, line 192: Personally, I feel that we should accept a parameter that allows the user to decide the number of tokens to apply for repetition_penalty. We should also maintain a separate list to use as repetition_penalty context instead of using the existing tokens list. This could cause issues when applying repetition_penalty to very long generated token lists.

vishal-14069: I actually thought about this, what are your thoughts about using a queue instead of a list to maintain the number of tokens to apply for repetition_penalty?

We can initialize the queue with a maxlen of number of tokens to apply for repetition penalty. The tokens in the queue would be automatically popping from the front as new tokens are added to the queue. Kind of like a fixed sliding window over the tokens.

Possible implementation:

```python
def generate_step(prompt: mx.array, model: nn.Module, temp: float, repetition_penalty: float = 1.0,
    repetition_context: Optional[Any] = None # previously generated_tokens
) -> Generator[Tuple[mx.array, mx.array], None, None]:
    
    # def sample
    
    y = prompt
    cache = None

    while True:
        logits, cache = model(y[:, None], cache=cache)
        logits = logits.squeeze(1)

        if repetition_context is not None:
            logits = apply_penalty(logits, repetition_context, repetition_penalty)

        y, prob = sample(logits)
        yield y, prob
        
def generate(model: nn.Module, tokenizer: PreTrainedTokenizer, prompt: str, temp: Optional[float], 
repetition_penalty: Optional[float], repetition_context_size: Optional[int], 
max_tokens: int = 100, verbose: bool = False, formatter: Callable = None,
) -> str:
 
    prompt_tokens = mx.array(tokenizer.encode(prompt))
    repetition_context = deque(maxlen=repetition_context_size)
    repetition_context.extend(tokenizer.encode(prompt))

    for (token, prob), n in zip(
        generate_step(prompt=prompt_tokens, model= model, 
                      temp=temp, repetition_penalty=repetition_penalty, 
                      repetition_context=repetition_context), range(max_tokens)):
                      
        if token[0] == tokenizer.eos_token_id:
            break
        if n == 0:
            prompt_time = time.perf_counter() - tic
            tic = time.perf_counter()
        tokens.append(token[0].item())

        if verbose:
           .....

        repetition_context.append(token[0].item())  # Update repetition context after each token

    tokens = tokenizer.decode(tokens).replace(REPLACEMENT_CHAR, "")
    
```

mzbac: As long as it is simple and easy to understand, I personally don't have a preference either way. However, it would be great to maintain the same coding style that mlx-lm currently uses.

vishal-14069:
Hi @mzbac I've implemented all the suggestions that you gave:

- Made `repetition_penalty` and `repetition_context` optional  in  `mlx_lm/utils/generate_step` for backward compatibility 

- Made `temp`, `repetition_penalty` and `repetition_context_size` optional in `mlx_lm/utils/generate` for backward compatibility 

- Implemented the `repetition_context_size` where the user can choose the number of tokens the repetition penalty has to be applied to, and I have also setup a separate list for repetitive penalty context 

Please review it and let me know, if there are any further changes to make. Thank you.

llms/mlx_lm/utils.py
```diff
@@ -150,20 +189,27 @@ def generate(
...
+        generate_step(
+            prompt_tokens, model, temp, repetition_penalty, repetition_context
+        ),
+        range(max_tokens),
+    ):
+        if token[0] == tokenizer.eos_token_id:
```
mzbac, line 207: What is the reason we need to change to use `token[0]` instead of `token`?

mzbac:
> Hi @mzbac I've implemented all the suggestions that you gave:
> 
> * Made `repetition_penalty` and `repetition_context` optional  in  `mlx_lm/utils/generate_step` for backward compatibility
> * Made `temp`, `repetition_penalty` and `repetition_context_size` optional in `mlx_lm/utils/generate` for backward compatibility
> * Implemented the `repetition_context_size` where the user can choose the number of tokens the repetition penalty has to be applied to, and I have also setup a separate list for repetitive penalty context
> 
> Please review it and let me know, if there are any further changes to make. Thank you.

Looks good. I will pull down the changes and run some tests on my local machine. Just a quick note, I noticed that you made changes to the code format in your PR. Have you installed the pre-commit hook and used it to format your code before committing? FYI, https://github.com/ml-explore/mlx-examples/blob/main/CONTRIBUTING.md#pull-requests

vishal-14069:
Sounds good. Yes, I have installed the pre-commit hook. Also ran a pre-commit check before committing, both the checks passed.

llms/mlx_lm/utils.py
```diff
@@ -139,6 +176,8 @@ def generate(
        tokenizer (PreTrainedTokenizer): The tokenizer.
        prompt (str): The string prompt.
        temp (float): The temperature for sampling (default 0).
+       repetition_penalty (Optional[float]): The penalty factor for repeating tokens.
```
mzbac, line 179: Issue: The new parameters need to be moved to the end of the function parameters. Otherwise, it will break the generate function which generate and receive max_token as repetition_penalty causing an error.

llms/mlx_lm/utils.py
```diff
@@ -85,10 +83,34 @@ def get_model_path(path_or_hf_repo: str) -> Path:
...
+        logits[:, indices] = selected_logits
+    return logits
+
+
 def generate_step(
     prompt: mx.array,
     model: nn.Module,
-    temp: float = 0.0,
+    temp: 0.0,
+    repetition_penalty: Optional[float],
```
mzbac, line 112: Issue: Assign default values for `repetition_penalty` and `repetition_context` as well, in case some users directly imports generate_step and uses it, so that we don't break their code.

llms/mlx_lm/utils.py
```diff
@@ -126,7 +161,9 @@ def generate(
     model: nn.Module,
     tokenizer: PreTrainedTokenizer,
     prompt: str,
-    temp: float = 0.0,
+    temp: Optional[float] = 0.0,
+    repetition_penalty: Optional[float] = 1.0,
+    repetition_context_size: Optional[int] = 5,
```
mzbac, line 166: Nit: Set the default repetition_context_size to 0 in order to avoid unnecessarily saving the token context if the user doesn't use `repetition_penalty`.

llms/mlx_lm/utils.py
```diff
@@ -185,7 +237,7 @@ def generate(
             print("No tokens generated for this prompt")
             return
         prompt_tps = prompt.size / prompt_time
```
mzbac, line 239: Somehow I am getting an error on `prompt_tps = prompt.size / prompt_time`. I suspect it is because you changed the `prompt` to `prompt_tokens`, so it throws the error "str" doesn't have a size property.

llms/mlx_lm/utils.py
```diff
@@ -150,20 +189,27 @@ def generate(
...
+        ),
+        range(max_tokens),
+    ):
+        if token[0] == tokenizer.eos_token_id:
             break
         if n == 0:
             prompt_time = time.perf_counter() - tic
             tic = time.perf_counter()
-        tokens.append(token.item())
+        tokens.append(token[0].item())
```
mzbac, line 212: Issue: Changing `token` to `token[0]` seems unnecessary, and using `token[0]` as a magic number would be bad for readability. Could you please revert the changes?

mzbac:
> Sounds good. Yes, I have installed the pre-commit hook. Also ran a pre-commit check before committing, both the checks passed.

Overall, it looks good. However, I found a few issues during my local test. Would you mind taking a look to see if those issues make sense to you?

llms/mlx_lm/utils.py
```diff
@@ -113,11 +138,21 @@ def sample(logits: mx.array) -> Tuple[mx.array, float]:
...
+        )
+
     y = prompt
     cache = None
+
     while True:
-        logits, cache = model(y[None], cache=cache)
-        logits = logits[:, -1, :]
+        logits, cache = model(y[:, None], cache=cache)
+        logits = logits.squeeze(1)
```
mzbac, line 151: Issue: Revert `logits = logits.squeeze(1)` back to `logits = logits[:, -1, :]` because we only return the last token from the generate step instead of the whole sequence.

llms/mlx_lm/utils.py
```diff
@@ -113,11 +138,21 @@ def sample(logits: mx.array) -> Tuple[mx.array, float]:
...
+            f"repetition_penalty must be a non-negative float, got {repetition_penalty}"
+        )
+
     y = prompt
     cache = None
+
     while True:
-        logits, cache = model(y[None], cache=cache)
-        logits = logits[:, -1, :]
+        logits, cache = model(y[:, None], cache=cache)
```
mzbac, line 150: Issue: The batch dimensions should be the first dimensions. Please revert this back to `y[None]` and revert `logits = logits.squeeze(1)` to `logits = logits[:, -1, :]`, as we need to retrieve the last token from the sequence dimension instead of removing sequence dimensions.

llms/mlx_lm/utils.py
```diff
@@ -113,11 +138,21 @@ def sample(logits: mx.array) -> Tuple[mx.array, float]:
         prob = softmax_logits[0, token]
         return token, prob
 
+    if repetition_penalty < 0 or not isinstance(repetition_penalty, float):
```
mzbac, line 141: Nit: I feel the check is a bit too restrictive. Maybe only `repetition_penalty` not being None is enough.

llms/mlx_lm/utils.py
```diff
@@ -150,20 +189,27 @@ def generate(
         print("=" * 10)
         print("Prompt:", prompt)
 
-    prompt = mx.array(tokenizer.encode(prompt))
+    prompt_tokens = mx.array(tokenizer.encode(prompt))
+    repetition_context = []
+    repetition_context.extend(tokenizer.encode(prompt))
```
mzbac, line 194:
```suggestion
    repetition_context.extend([token.item() for token in prompt_tokens])
```

llms/mlx_lm/utils.py
```diff
@@ -174,6 +220,12 @@ def generate(
                 print(s[skip:], end="", flush=True)
                 skip = len(s)
 
+    repetition_context.append(
+        token[0].item()
+    )  # Update repetition context after each token
+    repetition_context = repetition_context[
+        -repetition_context_size:
+    ]  # Maintain the specified context size
```
mzbac, line 228: Issue: The update repetition context should be inside the loop.
```suggestion
      repetition_context.append(
          token[0].item()
      )  # Update repetition context after each token
      repetition_context = repetition_context[
          -repetition_context_size:
      ]  # Maintain the specified context size
```

vishal-14069:
Hi @mzbac , I apologize for my oversight, especially about the `prompt.size` issue and the `repetition_context` list being outside the for loop, these changed when I merged after resolving conflicts from my latest pull.

llms/mlx_lm/utils.py
vishal-14069: `repetition_penalty` is supposed to be greater than 1.0 right? With `repetition_penalty` not being None we will end up accepting values below 1.0, that includes negative float/int

llms/mlx_lm/utils.py
vishal-14069: In the `generate_step` function, I was squeezing the first dimension of the logits array, and I ended up getting a  array with topk tokens after `mx.argmax`, and `token==tokenizer.eos_token_id` would fail because `token` consisted of topk tokens. I thought `token[0]` would extract the actual value from the y tensor in `generate_step`.

Following up with your suggestion, I changed `logins.squeeze(1)` to `logins[ : , -1 , : ]` that allowed me to get the last tokens from the sequence dimension, solving this problem. Thanks for that :)

vishal-14069:
Hey @mzbac, I have made the necessary changes. Please review the PR. Thank you.

llms/mlx_lm/utils.py
vishal-14069: I have removed one of the checks in the latest PR.

llms/mlx_lm/utils.py
```diff
@@ -150,14 +186,21 @@ def generate(
         print("=" * 10)
         print("Prompt:", prompt)
 
-    prompt = mx.array(tokenizer.encode(prompt))
+    prompt_tokens = mx.array(tokenizer.encode(prompt))
+    repetition_context = []
+    repetition_context.extend([token.item() for token in prompt_tokens])
```
mzbac, line 191: nit: don't create a repetition context if the user doesn't apply the repetition penalty.

llms/mlx_lm/utils.py
```diff
@@ -174,22 +217,29 @@ def generate(
...
+        repetition_context.append(
+            token.item()
+        )  # Update repetition context after each token
+        repetition_context = repetition_context[
+            -repetition_context_size:
+        ]  # Maintain the specified context size
+
     token_count = len(tokens)
-    token_string = tokenizer.decode(tokens).replace(REPLACEMENT_CHAR, "")
+    tokens = tokenizer.decode(tokens).replace(REPLACEMENT_CHAR, "")
```
mzbac, line 228: issue: `token_string` is a better name because it represents the decoded tokens, not longer tokens but rather a string.

mzbac:
@vishal-14069 looks good, but I have two minor concerns. Maybe ask @awni to review it and share his thoughts if he have time.

vishal-14069:
@mzbac I have added logic to form the `repetition_context` list only when `repetition_penalty>1.0`, I tested it out by using a couple of print statements, it works as expected. I have also reverted `tokens` to `token_string`.

llms/mlx_lm/utils.py
```diff
@@ -184,8 +238,8 @@ def generate(
         if token_count == 0:
             print("No tokens generated for this prompt")
             return
-        prompt_tps = prompt.size / prompt_time
-        gen_tps = (token_count - 1) / gen_time
+        prompt_tps = prompt_tokens.size / prompt_time
+        gen_tps = (len(token_count) - 1) / gen_time
```
mzbac, line 242: Nit: `token_count` is already a number, there's no need to apply `len(token_count)` again.

mzbac (approved):
I pulled down the code and tested it on my local machine, and it seems to be working as expected.

vishal-14069:
Hi @awni, if possible can you please take a look at this PR? Thank you:)

awni:
Yes for sure, sorry for the delay! Looks like a nice addition!

llms/mlx_lm/utils.py
```diff
@@ -132,10 +166,12 @@ def generate(
     model: nn.Module,
     tokenizer: PreTrainedTokenizer,
     prompt: str,
-    temp: float = 0.0,
+    temp: Optional[float] = 0.0,
```
awni, line 169: `Optional` implies the value can be `None` which is not the case for this.

llms/mlx_lm/utils.py
```diff
@@ -132,10 +166,12 @@ def generate(
...
     max_tokens: int = 100,
     verbose: bool = False,
     formatter: Callable = None,
+    repetition_penalty: Optional[float] = 1.0,
+    repetition_context_size: Optional[int] = 0,
```
awni, line 174: Same for these two?

llms/mlx_lm/utils.py
```diff
@@ -91,10 +91,34 @@ def get_model_path(path_or_hf_repo: str) -> Path:
     return model_path
 
 
+def apply_penalty(logits: mx.array, generated_tokens: Any, penalty: float):
```
awni, line 94: Nit: could you name this `apply_repitition_penalty`? it's a bit vague as is.

llms/mlx_lm/utils.py
```diff
@@ -91,10 +91,34 @@ def get_model_path(path_or_hf_repo: str) -> Path:
...
+    """
+    Apply repetition penalty to specific logits based on the given context.
+
+    Args:
+        logits (mx.array): The logits produced by the language model.
+        generated_tokens (any): A list of N previous tokens.
+        penalty (float): The repetition penalty factor to be applied.
+
+    Returns:
+        None: The function modifies the input logits in place.
```
awni, line 104: This is not correct.

llms/mlx_lm/utils.py
```diff
@@ -156,14 +194,23 @@ def generate(
         print("=" * 10)
         print("Prompt:", prompt)
 
-    prompt = mx.array(tokenizer.encode(prompt))
+    prompt_tokens = mx.array(tokenizer.encode(prompt))
+
+    repetition_context = []
+    if repetition_penalty > 1.0:
+        repetition_context.extend([token.item() for token in prompt_tokens])
```
awni, line 201:
```suggestion
    prompt = tokenizer.encode(prompt)

    repetition_context = []
    if repetition_penalty > 1.0:
        repetition_context = prompt
        
    prompt = mx.array(prompt)
```

awni, line 201:
```suggestion
    repetition_context = []
    if repetition_penalty > 1.0:
        repetition_context = prompt_tokens.tolist()
```

llms/mlx_lm/utils.py
```diff
@@ -180,6 +227,15 @@ def generate(
...
                 skip = len(s)
 
+        if repetition_penalty > 1.0:
+            repetition_context.append(
+                token.item()
+            )  # Update repetition context after each token
+            print(repetition_context)
+            repetition_context = repetition_context[
+                -repetition_context_size:
+            ]  # Maintain the specified context size
```
awni, line 237: Shouldn't we be doing this above as well when we first make the repition_context?

llms/mlx_lm/utils.py
```diff
@@ -156,14 +194,23 @@ def generate(
...
     tic = time.perf_counter()
     tokens = []
     skip = 0
     REPLACEMENT_CHAR = "\ufffd"
 
-    for (token, prob), n in zip(generate_step(prompt, model, temp), range(max_tokens)):
+    for (token, prob), n in zip(
+        generate_step(
+            prompt_tokens, model, temp, repetition_penalty, repetition_context
+        ),
```
awni, line 211: I think I am missing something but you are passing the context at the get go to the `generate_step` which makes a generator.

Then you update the context below as a slice of the old context which will make a new list. I do not think that will be used by the generate step function...

mzbac: Hmm, maybe that's the reason why I saw the repetition penalty being applied heavily to the first few tokens. I thought that as the context got longer, the penalty wouldn't apply as much.

vishal-14069: hmmm, yes the context wasn't used by the `generate_step` function. I made these changes in `generate_step` function so that the context gets updated within the generator loop. 

```python

# generate_step

y, prob = sample(logits)
repetition_context.append(y.item())
if len(repetition_context) > repetition_context_size:
    repetition_context= repetition_context[-repetition_context_size:]
yield y, prob

In the `generate` function we initialize `repetition_context` with the prompt tokens and call the generator_step function through a loop as usual. 

After making the above changes, I was running a few tests and I saw that context size of 0 and context size >=10 had significant impact on the generated tokens, kinda obvious I guess.

I can setup a PR with these changes for review, let me know.

**Comment by vishal-14069** [comment]
Hi @awni @mzbac, here are the changes I have made in this PR

* I have made sure that `repetition_context` is used by the `generate_step` function, and fixed the repetition penalty logic as well.

* I did test out the `generate` and `generate_step` function with and without repetition penalty and it seems to work fine. I can share the code as well as the output if needed.

* I have chosen to keep the function arguments as optional to have backward compatibility and in return I have added error handling for None values.

* Renamed the apply penalty function to `apply_repetition_penalty` 

Please review it and let me know if anything needs to be changed. Thank you.

**Comment by danilopeixoto** [comment]
> I'm curious, have you found this to be useful in any cases? If so I would love to see the an example of it working.

@awni @vishal-14069 This is what I initially had in mind:

**Demo**
[Implementing the Repetition Penalty Parameter for Mixtral Model on Apple Silicon using MLX](https://www.linkedin.com/posts/danilopeixoto8_applesilicon-inference-llm-activity-7161786918177910784-aWx_/)

**Comment by mzbac** [comment]
> I'm curious, have you found this to be useful in any cases? If so I would love to see the an example of it working.

@awni, sometimes when the model generates long text, it starts to repeat certain words. For example, while I am implementing the mlx-lm/server, I let the model generate a maximum of 4080 tokens. It gets stuck on something like:
```
USER: What is the best way to use a blender?
ASSISTANT: The best to use blender, you add the combination of the ingredients, the needed, the, and the butter.

USER: What is the best to make of the butter, the sugar, the, the sugar, the, the sugar the, the, the, the, the, the, the, the, the, the, the, the, the, the, the, the, the, the, the, the, the, the, the, the
```

**Comment by awni** [comment]
Yea I've definitely seen that as well @mzbac. Does this repetition penalty suppress those cases? That would be a good sanity check for it.

**Comment by mzbac** [comment]
> Yea I've definitely seen that as well @mzbac. Does this repetition penalty suppress those cases? That would be a good sanity check for it.

Alright, I will try to do some sanity check tomorrow.

**Comment by vishal-14069** [comment]
Hi @awni @mzbac I did some tests with a similar repetitive prompt. This is test from the latest PR. Here is output:

```python

prompt2 = "what is AI AI AI AI AI AI AI AI AI AI AI AI ?"

# Generate response without repetition penalty
response_without_penalty = generate(
    model=model,
    tokenizer=tokenizer,
    prompt=prompt2,
    temp=0,  
    max_tokens=50,
)

# Generate response with repetition penalty
response_with_penalty = generate(
    model=model,
    tokenizer=tokenizer,
    prompt=prompt2,
    temp=0,  
    max_tokens=50,
    repetition_penalty=1.3,
    repetition_context_size=20  # Applying repetition penalty
)

print("Response without repetition penalty:", response_without_penalty)
print("Response with repetition penalty:", response_with_penalty)
```

Output:

```markdown

Response without repetition penalty:  AI AI AI AI AI AI AI AI AI, AI AI AI AI AI AI AI AI AI AI AI AI AI AI AI AI AI AI AI AI AI AI AI AI AI, AI, AI, AI, AI, AI, AI, AI,

Response with repetition penalty: 
AI, Artificial Intelligence and Machine Learning are all terms that have been used interchangeably. However they each mean different things, so it is important to understand the difference between them before you start using these terms.
```

mzbac:
@awni I have done quick test on my local with phi2:
1. without repetitive penalty:
```
python -m mlx_lm.generate --model mlx-community/phi-2-hf-4bit-mlx --max-tokens 4080 --prompt "once up a time"
```
Output:
```
==========
Prompt: once up a time
-lapse video of the ephemeral flashes.
"This study is the first to provide a direct link between mercury, selenium, and lightning," said study co-author William L. Taylor, a professor of geology at the University of Tennessee.
The team found that lightning kills off the mercury in the air and deposits it into the air, soil, and water. The mercury in the air is then absorbed by the plants in the area. The high concentration of mercury in the soil, water, and plants can be toxic to wildlife and humans.
The researchers also found that lightning can be used as an indicator of the presence of mercury in the air. The team found that lightning was more likely to occur in areas with a high concentration of mercury in the air.
"This study is a great example of how scientific research can be used to help protect our environment and health," said Taylor.
The team's findings have been published in the journal Environmental Science & Technology.
The team's findings have been published in the journal Environmental Science & Technology. The study is the first to provide a direct link between lightning and mercury deposition. The team found that lightning kills off the mercury in the air and deposits it into the air, soil, and water. The high concentration of mercury in the soil, water, and plants can be toxic to wildlife and humans.
The team's findings have been published in the journal Environmental Science & Technology. The study is the first to provide a direct link between lightning and mercury deposition. The team found that lightning kills off the mercury in the air and deposits it into the air, soil, and water. The high concentration of mercury in the soil, water, and plants can be toxic to wildlife and humans.
The team's findings have been published in the journal Environmental Science & Technology. The study is the first to provide a direct link between lightning and mercury deposition.^CTraceback (most re
```
2. with rrepetition_context_size 20 and repetition_penalty 1.2:
```
python -m mlx_lm.generate --model mlx-community/phi-2-hf-4bit-mlx --max-tokens 4080 --prompt "once up a time"
```
Output:
```
==========
Prompt: once up a time
-lapse video of the ephemeral flashes.
"This study is just an initial step, but it suggests that high-speed cameras can be used to capture the first moments of a flash starting and then tracking it as it evolves," said Steven Ting, assistant professor in the Department of Physics at the University at Buffalo.
Much remains to be discovered about these enigmatic events, said Michael Tremaine of the University at Buffalo and lead author on a paper describing the research in Physical Review Letters.
"We still don't understand how these flashes are triggered, or what causes them to happen," Tremaine said.
The study was motivated by the desire of astronomers in Chile and South Africa to get a better handle on what happens when an optical burst occurs.
"Since these flashes can be observed in the ultraviolet, visible and near-infrared parts of the electromagnetic spectrum," Tremaine said. "We wanted to see if it was possible to capture this brief time window."
The researchers used a high-speed camera and an infrared filter in front of the lens to capture flashes that occur on astronomical objects known as blazars. These are very energetic sources in the distant universe, and they emit a large spectrum of radiation that includes ultraviolet light.
The camera captured images at up to 1,000 frames per second in the visible and near-infrared part of the electromagnetic spectrum. Tremaine said that a typical high-speed camera can capture these types of events at up to 100 frames per second.
"This is like trying to get a video recording of the first flash that happens when you blow out a birthday candle," Tremaine said. "The video can show the entire process, but it is extremely difficult to analyze."
By using a high-speed camera combined with an infrared filter, the researchers were able to record flashes in ultraviolet light that take place on astronomical objects known as blazars. Blazar outbursts have been observed to produce a "frighteningly large" number of optical flashes, Tremaine said.
The team found that the brightness and duration of these events are very different from one another, suggesting a correlation between flash size and flash rate.
"This suggests that the number of flashes produced by an optical burst is controlled, or at least has a limit," Tremaine said. "It also means that the flash rate can be used to estimate how many optical flashes are produced."
The team was also able to determine that the rate at which an optical flash occurs is controlled by a very brief, intense burst of ultraviolet light.
"It appears that the rate at which optical flashes appear is controlled by a very brief, intense burst of ultraviolet light," Tremaine said. "This suggests that the rate at which optical flashes appear may be controlled by a very brief, intense burst of ultraviolet light.
"This could have important implications for the future development of optical flash systems," he said, "which are used in a variety of applications."
Tremaine and his colleagues plan to continue their research^CTraceback (most recent call last):
```

llms/mlx_lm/utils.py
```diff
@@ -150,20 +200,37 @@ def generate(
...
+       repetition_penalty (float, optional): The penalty factor for repeating tokens.
+       repetition_context_size (int, optional): The number of tokens to consider for repetition penalty.
     """
 
     if verbose:
         print("=" * 10)
         print("Prompt:", prompt)
 
-    prompt = mx.array(tokenizer.encode(prompt))
+    prompt_tokens = mx.array(tokenizer.encode(prompt))
```
awni, line 211: Nit: you don't need to change the name of the prompt here do you? Could just keep it as `prompt` as before unless I missed something.

awni:
@vishal-14069 this looks much better, thanks for the updates! And thank you and @mzbac for sharing the results.

I left a couple more comments. Could you check? After that I think we can merge this.

llms/mlx_lm/utils.py
mzbac: it kind of makes sense to change it since the prompt type changed be an array of token ids instead of a string. I'm not sure if I suggested these changes, but I don't have a strong opinion on this..

awni: Fair point, let's keep it then.

llms/mlx_lm/utils.py
```diff
@@ -91,10 +91,35 @@ def get_model_path(path_or_hf_repo: str) -> Path:
...
+    return logits
+
+
 def generate_step(
     prompt: mx.array,
     model: nn.Module,
-    temp: float = 0.0,
+    temp: 0.0,
+    repetition_penalty: Optional[float] = None,
+    repetition_context: Optional[List[int]] = None,
```
awni, line 121: (I thought I left this comment earlier, but it didn't save....)

Can repetition context simply be constructed inside the `generate_step` function rather passing it as a parameter? It looks like it is only used inside this function and you have all you need to construct it? Do we gain anything by passing it from outside the function?

mzbac: Maybe it was to apply the penalty for prompt, but the generate_step has the prompt tokens that can also be done inside the generate_step.

vishal-14069: Yes, I have removed `repetition_context` parameter,  and constructed it inside the `generate_step` function.

I have one suggestion, with the current approach we are applying the context size on the prompt as well. Maybe we apply repetition penalty for every single token in the prompt, and use context size only on the generated tokens?

mzbac: @vishal-14069 from what I can see in the transformers source code, it seems like the prompt included for it, but I cannot be 100% sure. Maybe you can double-check to see if it needs to be applied to the prompt.

https://github.com/huggingface/transformers/blob/main/src/transformers/generation/utils.py#L2645
https://github.com/huggingface/transformers/blob/main/src/transformers/generation/logits_process.py#L288

llms/mlx_lm/utils.py
vishal-14069: Yes, the repetition penalty needs to be applied on the prompt as well. In my last comment what I was trying to say is that, in the latest approach we are using `repetition_context_size` on the prompt tokens.  

```python
repetition_context = []
     if repetition_penalty and repetition_penalty > 1.0:
         repetition_context = prompt_tokens.tolist()
         repetition_context = repetition_context[-repetition_context_size:] # this part 
```

I am suggesting that we use repetition penalty on the prompt tokens, but not the `repetition_context_size` on the prompt tokens, rather we apply repetition penalty on all the prompt tokens irrespective of the `repetition_context_size`  which will create a window on the prompt. 

Maybe we can use `repetition_context_size` only on the generated tokens. 

Suggested change:

```python
repetition_context = []
     if repetition_penalty and repetition_penalty > 1.0:
         repetition_context = prompt_tokens.tolist()
```

I will setup a PR, it will be easier for review.

vishal-14069:
Hi @mzbac @awni the latest PR contains these changes:

- The `repetition_context` is removed as an argument from `generate_step` function rather it is constructed inside `generate_step`

- I have also removed the usage of `repetition_context_size` on the prompt tokens, I've used it only on the generated tokens. Let me know if this change works.

Thank you.

llms/mlx_lm/utils.py
vishal-14069: Yes, the repetition penalty needs to be applied on the prompt as well. In my last comment what I was trying to say is that, in the latest approach we are using `repetition_context_size` on the prompt tokens.  

```python
repetition_context = []
     if repetition_penalty and repetition_penalty > 1.0:
         repetition_context = prompt_tokens.tolist()
         repetition_context = repetition_context[-repetition_context_size:] # this part 
```

I am suggesting that we use repetition penalty on the prompt tokens, but not the `repetition_context_size` on the prompt tokens, rather we apply repetition penalty on all the prompt tokens irrespective of the `repetition_context_size`  which will create a window on the prompt. 

Maybe we can use `repetition_context_size` only on the generated tokens. 

Suggested change:

```python
repetition_context = []
     if repetition_penalty and repetition_penalty > 1.0:
         repetition_context = prompt_tokens.tolist()
```

I can setup a PR, it will be easier for review.

llms/mlx_lm/utils.py
```diff
@@ -119,12 +146,33 @@ def sample(logits: mx.array) -> Tuple[mx.array, float]:
...
+        repetition_penalty < 0 or not isinstance(repetition_penalty, float)
+    ):
+        raise ValueError(
+            f"repetition_penalty must be a non-negative float, got {repetition_penalty}"
+        )
+
     y = prompt
     cache = None
+
+    repetition_context = prompt.tolist()
```
mzbac, line 159: @vishal-14069 I am a bit concerned if the prompt is very large. For instance, having too many tokens might lead to problems with text generation. Perhaps it is still necessary to apply a size limit to it.

vishal-14069: Hmm alright, yeah that makes sense. I will revert it back.

llms/mlx_lm/utils.py
```diff
@@ -91,10 +91,34 @@ def get_model_path(path_or_hf_repo: str) -> Path:
     return model_path
 
 
+def apply_repetition_penalty(logits: mx.array, generated_tokens: Any, penalty: float):
+    """
+    Apply repetition penalty to specific logits based on the given context.
```
awni, line 96: Maybe worth putting a reference to the paper in the PR here?

llms/mlx_lm/utils.py
```diff
@@ -91,10 +91,34 @@ def get_model_path(path_or_hf_repo: str) -> Path:
...
+    return logits
+
+
 def generate_step(
     prompt: mx.array,
     model: nn.Module,
-    temp: float = 0.0,
+    temp: 0.0,
+    repetition_penalty: Optional[float] = None,
+    repetition_context_size: Optional[int] = None,
```
awni, line 121: Is `None` a reasonable default for the `repetition_context_size`? IIRC it will use the full sequence if this is `None`? Maybe we should set it at a more reasonable default? What do you typically use when you run this?

awni:
@vishal-14069 I have a couple more comments, sorry for belaboring this. Would you mind checking? O/w I think the implementation looks great and I expect it will be quite useful!!

llms/mlx_lm/utils.py
vishal-14069: Yes if `repetition_context_size` is `none` it will be applied for the entire sequence. I feel that it's a bit tricky to set a default repetition context size as it is subjective. Since we are setting the `max_tokens` to be 100 as default in `generate` I think we can set a default `repetition_context_size` with respect to `max_tokens` being 100? Probably 20 or 25?

@mzbac @awni  any thoughts on this ?

awni: Anything in the 10-20 range seems reasonable to me. I would lean towards less penalty as a default rather than higher so using a smaller default repetition context size seems more in that direction.

vishal-14069:
Hi @awni, here are the changes made in the latest PR:

- I have updated `apply_repetition_penalty` function's docstring with the CTRL paper link

- I have set the `repetition_context_size` to be 20 in `generate_step`, as I have found that a repetition context size of 20 to be reasonable and provides good results, while not having too much impact on the generated tokens.

awni (approved):
Thanks!!

kamranjon:
Is this exposed through the command line interface? I didn't see it in the help document. It would be nice to surface this because it's one of the recommended parameters to set for the Gemma 3 models.

PR #405, opened 2024-02-03
add segformer example implementation

Jimexist:
so far this is blocked on:
- https://github.com/ml-explore/mlx/issues/100


awni:
Awesome! Would be great if you can comment on that PR re groups and dilation so we can get a sense of priority

awni:
Given the age of this PR and that we have SAM already, let's close it. Would be happy to point to an external repo with an implementation if you pick it back up!

Issue #407, opened 2024-02-04
max_seq_length


ivanfioravanti:
max_seq_length is available in TrainingArgs of tuner. It would be great being able to pass it as parameter to lora.py to change it during training.

ivanfioravanti:
Merged!

ivanfioravanti:
@awni would you like me to update the old lora sample based on wikisql?

awni:
No that's not necessary.. I think it's ok for that example to stay without new features, mostly there for instructive purposes. The mlx-lm LoRA is the one that people should use if they are looking for something with more bells and whistles.

PR #408, opened 2024-02-04
Add max sequence length argument in lora.py

ivanfioravanti:
For the issue #407 

A new argument "--max_seq_length" has been added to the command-line parser and passed as a parameter to the main function of the lora.py script. This allows users to specify and control the maximum sequence length during training.


awni (approved):
Nice!!

PR #413, opened 2024-02-05
chore(mlx-lm): add model weight index in save_weights

mzbac:
Due to the various different model formats, such as hf/pytorch/custom, it would be beneficial if we had a model weight metadata index. This way, we wouldn't have to download the model weight and check it locally. This approach also aligns with most of the Hugging Face model repository.

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -388,7 +388,8 @@ def save_weights(save_path: Union[str, Path], weights: Dict[str, Any]) -> None:
         else "model.safetensors"
     )
 
-    index_data = {"metadata": {"total_size": 0}, "weight_map": {}}
+    total_size = sum(v.nbytes for v in weights.values())
+    index_data = {"metadata": {"total_size": total_size}, "weight_map": {}}
 
     for i, shard in enumerate(shards):
         shard_name = shard_file_format.format(i + 1, shards_count)
@@ -398,8 +399,6 @@ def save_weights(save_path: Union[str, Path], weights: Dict[str, Any]) -> None:
 
         for weight_name in shard.keys():
             index_data["weight_map"][weight_name] = shard_name
-
-        index_data["metadata"]["total_size"] += shard_path.stat().st_size
 
     index_data["weight_map"] = {
         k: index_data["weight_map"][k] for k in sorted(index_data["weight_map"])
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -12,7 +12,6 @@ from huggingface_hub import snapshot_download
 from transformers import AutoConfig, AutoTokenizer, PreTrainedTokenizer
 
 # Local imports
-
 from .models import llama, mixtral, olmo, phi2, plamo, qwen, qwen2, stablelm_epoch
 from .tuner.utils import apply_lora_layers
 
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -12,7 +12,7 @@ from huggingface_hub import snapshot_download
 from transformers import AutoConfig, AutoTokenizer, PreTrainedTokenizer
 
 # Local imports
-from .models import llama, mixtral, phi2, plamo, qwen, stablelm_epoch, qwen2
+from .models import llama, mixtral, phi2, plamo, qwen, qwen2, stablelm_epoch
 from .tuner.utils import apply_lora_layers
 
 # Constants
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/utils.py
```diff
@@ -24,10 +24,14 @@
...
-    "qwen2": qwen2
+    "qwen2": qwen2,
 }
 LORA_SUPPORTED_MODELS = [
-    llama.Model, mixtral.Model, phi2.Model, stablelm_epoch.Model, qwen2.Model
+    llama.Model,
+    mixtral.Model,
+    phi2.Model,
+    stablelm_epoch.Model,
+    qwen2.Model,
```
awni, line 34: ❤️

llms/mlx_lm/utils.py
```diff
@@ -384,6 +388,26 @@ def save_weights(save_path: Union[str, Path], weights: Dict[str, Any]) -> None:
...
+        mx.save_safetensors(str(shard_path), shard)
+
+        for weight_name in shard.keys():
+            index_data["weight_map"][weight_name] = shard_name
+
+        index_data["metadata"]["total_size"] += shard_path.stat().st_size
+
+    sorted_weight_map = {
+        k: index_data["weight_map"][k] for k in sorted(index_data["weight_map"])
+    }
```
awni, line 406:
```suggestion
    index_data["weight_map"] = {
        k: index_data["weight_map"][k] for k in sorted(index_data["weight_map"])
    }
```
```suggestion
    index_data["weight_map"] = {
        k: index_data["weight_map"][k] for k in sorted(index_data["weight_map"])
    }
```

llms/mlx_lm/utils.py
```diff
@@ -384,6 +388,26 @@ def save_weights(save_path: Union[str, Path], weights: Dict[str, Any]) -> None:
...
+
+    with open(save_path / "model.safetensors.index.json", "w") as f:
+        json.dump(
+            {"metadata": index_data["metadata"], "weight_map": sorted_weight_map},
```
awni, line 410:
```suggestion
            index_data,
```

llms/mlx_lm/utils.py
```diff
@@ -384,6 +388,26 @@ def save_weights(save_path: Union[str, Path], weights: Dict[str, Any]) -> None:
...
         shard_name = shard_file_format.format(i + 1, shards_count)
-        mx.save_safetensors(str(save_path / shard_name), shard)
+        shard_path = save_path / shard_name
+
+        mx.save_safetensors(str(shard_path), shard)
+
+        for weight_name in shard.keys():
+            index_data["weight_map"][weight_name] = shard_name
+
+        index_data["metadata"]["total_size"] += shard_path.stat().st_size
```
awni, line 402: The total size is supposed to be the size in bytes right? (As compared to say the # of params or something)?

awni (approved):
Looks great! I think this will be a really nice addition, thanks for adding it. I left minor comments, let me know when you are ready to merge.

llms/mlx_lm/utils.py
mzbac: I just implemented the naive way to put the file size in bytes therre. If you prefer, I can double-check the transformers implementation to see if it should be "pramas size" or "file size".

awni: Maybe @pcuenca knows. @pcuenca we are wondering if in safetensors index json, the `total_size` is bytes or # params or some other thing. Do you happen to know? (E.g. here https://huggingface.co/mistralai/Mistral-7B-v0.1/blob/main/model.safetensors.index.json#L3)

mzbac: Looks like it's the parameter size. I will update the implementation accordingly.
https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L379

awni: Thanks!!

mzbac:
@awni, I have updated the total_size to match the parameter size. This PR should be ready to merge now. Let me know if anything else needs to be changed.

pcuenca (approved):
This is a great addition, thanks a lot for adding it!

At some point we could split the shards and build the index at the same time. I think there's a helper function being worked on in the HF side that we could potentially leverage in the future.

mzbac:
> This is a great addition, thanks a lot for adding it!
> 
> At some point we could split the shards and build the index at the same time. I think there's a helper function being worked on in the HF side that we could potentially leverage in the future.

Yeah, I was worried that if the make_shard function returns an additional index, it might break existing users who use the make_shard utility function. Otherwise, following the same approach as transformers would be better.

pcuenca:
By the way, this will unlock the "Safetensors" section in the model card, see for example https://huggingface.co/codellama/CodeLlama-13b-Instruct-hf:

Screenshot 2024-02-06 at 10 44 58

cc @Wauplin in case we are missing anything for that to happen.

pcuenca:
> I was worried that if the make_shard function returns an additional index, it might break

Same! @Wauplin, there was an effort to provide a library-agnostic way to build shards iirc?

awni (approved):
Beautiful, thanks!!

Wauplin:
> Same! @Wauplin, there was an effort to provide a library-agnostic way to build shards iirc?

Yes indeed! PR is in progress here: https://github.com/huggingface/huggingface_hub/pull/1938. Most of the logic comes from `transformers`. I'll let you know once it's merged and released!

PR #414, opened 2024-02-05
LoRA: add training callbacks

madroidmaq:
In order to better monitor the details of the training process, we do some custom logic processing, and expose the training details to the caller in the form of a callback. For example, with this callback I can upload the training data to wandb (or other similar platforms).

For example, this PR #364.

If there are other better ways to do this, please feel free to suggest them.

### DIFF
--- a/llms/mlx_lm/lora.py
+++ b/llms/mlx_lm/lora.py
@@ -7,7 +7,6 @@ import mlx.optimizers as optim
 import numpy as np
 from mlx.utils import tree_flatten
 
-from .tuner.lora import LoRALinear
 from .tuner.trainer import TrainingArgs, evaluate, train
 from .tuner.utils import linear_to_lora_layers
 from .utils import generate, load
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -119,6 +119,17 @@ def evaluate(
     return np.sum(all_losses) / ntokens
 
 
+class TrainingCallback:
+
+    def on_train_loss_report(self, steps: int, loss: float, it_sec: float, tokens_sec: float):
+        """Called to report training loss at specified intervals."""
+        pass
+
+    def on_val_loss_report(self, steps: int, loss: float, val_time: float):
+        """Called to report validation loss at specified intervals or the beginning."""
+        pass
+
+
 def train(
     model,
     tokenizer,
@@ -127,8 +138,11 @@ def train(
     val_dataset,
     args: TrainingArgs = TrainingArgs(),
     loss: callable = default_loss,
-    iterate_batches: callable = iterate_batches
+    iterate_batches: callable = iterate_batches,
+    training_callback=None,
 ):
+    print(f"Starting training..., iters: {args.iters}")
+
     # Create checkpoints directory if it does not exist
     if not os.path.exists("checkpoints"):
         os.makedirs("checkpoints")
@@ -138,7 +152,6 @@ def train(
 
     losses = []
     n_tokens = 0
-    print("Starting training..., iters:", args.iters)
     # Main training loop
     start = time.perf_counter()
     for it, batch in zip(
@@ -168,11 +181,17 @@ def train(
             train_loss = np.mean(losses)
 
             stop = time.perf_counter()
+            it_sec = args.steps_per_report / (stop - start)
+            tokens_sec = float(n_tokens) / (stop - start)
             print(
                 f"Iter {it + 1}: Train loss {train_loss:.3f}, "
-                f"It/sec {args.steps_per_report / (stop - start):.3f}, "
-                f"Tokens/sec {float(n_tokens) / (stop - start):.3f}"
+                f"It/sec {it_sec:.3f}, "
+                f"Tokens/sec {tokens_sec:.3f}"
             )
+
+            if training_callback is not None:
+                training_callback.on_train_loss_report(it + 1, train_loss, it_sec, tokens_sec)
+
             losses = []
             n_tokens = 0
             start = time.perf_counter()
@@ -190,11 +209,15 @@ def train(
                 max_seq_length=args.max_seq_length,
                 iterate_batches=iterate_batches
             )
+            val_time = time.perf_counter() - stop
             print(
                 f"Iter {it + 1}: "
                 f"Val loss {val_loss:.3f}, "
-                f"Val took {(time.perf_counter() - stop):.3f}s"
+                f"Val took {val_time:.3f}s"
             )
+
+            if training_callback is not None:
+                training_callback.on_val_loss_report(it + 1, val_loss, val_time)
 
             start = time.perf_counter()
 
@@ -205,6 +228,7 @@ def train(
             print(
                 f"Iter {it + 1}: Saved adapter weights to {os.path.join(checkpoint_adapter_file)}."
             )
+
     # save final adapter weights
     save_adapter(model=model, adapter_file=args.adapter_file)
     print(f"Saved final adapter weights to {os.path.join(args.adapter_file)}.")
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -121,7 +121,7 @@ def evaluate(
 
 class TrainingCallback:
 
-    def on_train_loss_report(self, steps: int, loss: float, it_sec: float, tokens_sec: float):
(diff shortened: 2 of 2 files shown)


awni:
I wonder if we can simplify this a bit:
- Keep the original printing as is
- Provide a single argument logging_callback which is called in a couple places. I don't think it needs to be called everywhere right?

chimezie:
I have had to do something similar to this (passing callbacks to **iterate_batches** i.e., #391).  For the case of downstream logging, I pass reported_train_loss_data and validation_loss_data lists to the function.  And they are updated within the function in this way:

```python
        # Report training loss if needed
        if (it + 1) % args.steps_per_report == 0:
            [.. snip ..]
            if reported_train_loss_data is not None:
                reported_train_loss_data.append((it, train_loss, iters_per_sec, num_tokens_per_sec))
```
and

```python
        # Report validation loss if needed
        if it == 0 or (it + 1) % args.steps_per_eval == 0:
            [..snip..]
            if validation_loss_data is not None:
                validation_loss_data.append((it, val_loss, val_run_time))
```

I think that is all that is needed for upstream applications to send the logging information to a CSV file, wandb, etc..

madroidmaq:
> I wonder if we can simplify this a bit:
> 
> * Keep the original printing as is
> * Provide a single argument logging_callback which is called in a couple places. I don't think it needs to be called everywhere right?

@awni I've done some logic simplification as you suggested, please REVIEW some of the changes in this section to see if they meet your expectations.

llms/mlx_lm/tuner/trainer.py
```diff
@@ -183,23 +211,28 @@ def train(
                 num_batches=args.val_batches,
                 max_seq_length=args.max_seq_length,
             )
-            print(
-                f"Iter {it + 1}: "
-                f"Val loss {val_loss:.3f}, "
-                f"Val took {(time.perf_counter() - stop):.3f}s"
-            )
+            took = time.perf_counter() - stop
```
awni, line 214:
```suggestion
            val_time = time.perf_counter() - stop
```

llms/mlx_lm/tuner/trainer.py
```diff
@@ -183,23 +211,28 @@ def train(
...
+
+            print(f"Iter {it + 1}: Val loss {loss:.3f}, Val took {took:.3f}s")
+            if training_callback is not None:
+                training_callback.on_val_loss_report(it + 1, val_loss, took)
```
awni, line 218:
```suggestion
            print(f"Iter {it + 1}: Val loss {loss:.3f}, Val took {val_time:.3f}s")
            if training_callback is not None:
                training_callback.on_val_loss_report(it + 1, val_loss, val_time)
```

llms/mlx_lm/lora.py
```diff
@@ -177,15 +177,15 @@ def load_dataset(args):
...
         l.self_attn.q_proj = LoRALinear.from_linear(l.self_attn.q_proj)
         l.self_attn.v_proj = LoRALinear.from_linear(l.self_attn.v_proj)
         if hasattr(l, "block_sparse_moe"):
             l.block_sparse_moe.gate = LoRALinear.from_linear(l.block_sparse_moe.gate)
 
-    p = sum(v.size for _, v in tree_flatten(model.parameters())) / 10**6
+    p = sum(v.size for _, v in tree_flatten(model.parameters())) / 10 ** 6
     print(f"Total parameters {p:.3f}M")
-    p = sum(v.size for _, v in tree_flatten(model.trainable_parameters())) / 10**6
+    p = sum(v.size for _, v in tree_flatten(model.trainable_parameters())) / 10 ** 6
```
awni, line 188: Could you run the pre-commit hooks? I think the formatter liked it more the way it was.

llms/mlx_lm/tuner/trainer.py
```diff
@@ -118,6 +118,28 @@ def evaluate(
...
+        """Called to report validation loss at specified intervals or the beginning."""
+        pass
+
+    def on_checkpoint_save(self, steps: int, file: str):
+        """Called to save model checkpoints at specified intervals."""
+        pass
+
+    def on_train_end(self, file: str):
+        """Called at the end of training."""
+        pass
```
awni, line 140: I'm wondering if we need all of this or if it could be simplified to just a single function? Like for the example you gave logging to wanddb, just having a little callback which saves train and validation loss would work right? Or did you need the callbacks in other places as well?

madroidmaq: There is no strong necessity. You can also know the start and end by calling the train function itself. I removed this part in the new submission.

But if the callback contains these two functions, more complete logic can be executed in the callback. At the same time, there is no need to put some initialization code logic together with the logic of calling the train function. Of course, this flexibility will increase the complexity of the code.

llms/mlx_lm/tuner/trainer.py
```diff
@@ -118,6 +118,21 @@ def evaluate(
...
+        """Called to report training loss at specified intervals."""
+        pass
+
+    def on_val_loss_report(self, steps: int, loss: float, val_time: float):
+        """Called to report validation loss at specified intervals or the beginning."""
+        pass
+
+    def on_checkpoint_save(self, steps: int, file: str):
+        """Called to save model checkpoints at specified intervals."""
+        pass
```
awni, line 133: Are you using this one? If not, maybe remove as well.

awni:
Thanks for simplifying! Would you mind rebasing and resolving conflicts? Should be good to go o/w.

madroidmaq:
> Thanks for simplifying! Would you mind rebasing and resolving conflicts? Should be good to go o/w.

@awni I've merged the latest code and resolved the conflict.

In addition to this, I'd like to add the number of tokens that have been trained to the log printout and callback, so that I can visualize how many tokens have been used throughout the entire LoRA training process, which is more often referred to as the number of tokens used by Microsoft in total than the number of iterations that have been trained. My initial additions to the new formula are roughly as follows:

```
Starting training..., iters: 12800
Iter 1: Val loss 3.092, Val took 2.807s
Iter 10: Train loss 2.905, It/sec 6.574, Tokens/sec 687.604, Trained Tokens 1046
Iter 20: Train loss 2.206, It/sec 5.338, Tokens/sec 564.809, Trained Tokens 2104
Iter 30: Train loss 1.563, It/sec 5.374, Tokens/sec 562.138, Trained Tokens 3150
Iter 40: Train loss 1.185, It/sec 5.781, Tokens/sec 604.068, Trained Tokens 4195
Iter 50: Train loss 1.024, It/sec 5.925, Tokens/sec 618.541, Trained Tokens 5239
Iter 60: Train loss 0.937, It/sec 5.961, Tokens/sec 608.658, Trained Tokens 6260
Iter 70: Train loss 0.823, It/sec 5.940, Tokens/sec 594.631, Trained Tokens 7261
Iter 80: Train loss 0.803, It/sec 5.933, Tokens/sec 598.066, Trained Tokens 8269
Iter 90: Train loss 0.779, It/sec 6.023, Tokens/sec 613.189, Trained Tokens 9287
Iter 100: Train loss 0.811, It/sec 5.937, Tokens/sec 622.208, Trained Tokens 10335
```

If you think this part of the information page deserves to be added, I can submit this part of the logic adjustment.

madroidmaq:
> I think it's a nice idea to add the number of tokens, feel free to add that on to this PR!

Its DONE

llms/mlx_lm/tuner/trainer.py
```diff
@@ -127,18 +138,24 @@ def train(
...
+    training_callback=None,
 ):
+    print(f"Starting training..., iters: {args.iters}")
+
     # Create checkpoints directory if it does not exist
     if not os.path.exists("checkpoints"):
         os.makedirs("checkpoints")
 
+    if training_callback is not None:
+        training_callback.on_train_begin(args.iters)
```
awni, line 151: This looks like a bug?

awni:
This all looks great!! Could you fix the issue I pointed out? Then I will merge it, thank you!

llms/mlx_lm/tuner/trainer.py
madroidmaq: Yes, this was an bug I introduced when rebasing the main branch and it's fixed now.

madroidmaq:
> This all looks great!! Could you fix the issue I pointed out? Then I will merge it, thank you!

Yes, this was an bug I introduced when rebasing the main branch and it's fixed now.

awni (approved):
🚀

ivanfioravanti:
WOW! Love this one! I was looking for a good way to pass data to wandb! Thanks @madroidmaq 🙏

PR #415, opened 2024-02-06
Olmo in MLX LM

awni:
Simple PR just adds an Olmo model class to run the new Allen AI model.

Using the safetensors version, model works:

```shell
python -m mlx_lm.generate --model saurabh-shah/olmo-7b-safetensors --prompt "Write a quick sort in C++" 
Fetching 14 files: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████| 14/14 [00:00<00:00, 36815.21it/s]
==========
.
Write a quick sort in C++.

## Requirements

- You must use a C++ compiler to complete this assignment.
- You may not use any additional libraries.
- You must use C++11.
- Your sort must be a stable sort (i.e., it returns the same order for the same inputs
  across runs).
- Your sort must be in-place.

## Submission

Submit your code as a
==========
Prompt: 9.931 tokens-per-sec
Generation: 20.045 tokens-per-sec
```

### DIFF
--- a/llms/mlx_lm/models/olmo.py
+++ b/llms/mlx_lm/models/olmo.py
@@ -1,0 +1,159 @@
+from dataclasses import dataclass
+from typing import Dict, Optional, Tuple, Union
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .base import BaseModelArgs
+
+try:
+    import hf_olmo
+except ImportError:
+    print("To run olmo install ai2-olmo: pip install ai2-olmo")
+    exit(1)
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    d_model: int
+    n_layers: int
+    mlp_hidden_size: int
+    n_heads: int
+    vocab_size: int
+    embedding_size: int
+    rope_theta: float = 10000
+    rope_traditional: bool = False
+    model_type: str = None
+
+
+class LayerNorm(nn.LayerNorm):
+    def __call__(self, x: mx.array) -> mx.array:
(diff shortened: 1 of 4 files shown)


mzbac:
That's something I planned to do tonight :p

awni:
Let me know if you have any comments! Also we should do some 4 bit versions, any help there much appreciated!

llms/mlx_lm/models/olmo.py
```diff
@@ -0,0 +1,159 @@
+from dataclasses import dataclass
+from typing import Dict, Optional, Tuple, Union
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .base import BaseModelArgs
+
+try:
+    import hf_olmo
```
mzbac, line 10: Just curious, what is the reason we need to import hf_olmo? It doesn't seem to be used.

mzbac:
Once the PR is merged, I can help quantify the olmo series models to mlx 4-bit tonight.

llms/mlx_lm/models/olmo.py
awni: Well yes, I added that because the tokenizer needs it (I am not actually sure if it needs it but it tries to import it and it won't load otherwise). And I didn't want to add this as a general dependency..which seemed a bit overkill.

mzbac: makes sense. Thanks for clarifying it.

angeloskath (approved):
👌

PR #418, opened 2024-02-06
fix-mistral-download-link

thelightbleu:
fixing broken link for mistral-7B-v0.1 model

Original change from mistral.ai:

https://github.com/mistralai/mistral-src/commit/8598cf582091a596671be31990448e0620017851


awni (approved):
Thanks!!

PR #421, opened 2024-02-07
Adding StoppingCriteria to generate in utils

sderygithub:
A stopgap solution I had implemented locally; I defer to maintainers if this is something they want to integrate further. Let me know!

### DIFF
--- a/llms/mlx_lm/stopping_criteria.py
+++ b/llms/mlx_lm/stopping_criteria.py
@@ -1,0 +1,31 @@
+from abc import ABC
+
+class StoppingCriteria(ABC):
+    """Abstract base class for all stopping criteria that can be applied during generation."""
+
+    def __call__(self, token_string: str, **kwargs) -> bool:
+        raise NotImplementedError("StoppingCriteria needs to be subclassed")
+
+class StopSequenceCriteria(StoppingCriteria):
+    """Stopping criteria that compares the generated string for presence of a
+    sequence of tokens."""
+
+    def __init__(self, stop_sequence, prompt):
+        self.stop_sequence = stop_sequence
+        self.prompt = prompt
+
+    def __call__(self, tokens_string: str, **kwargs):
+        # Remove prompt to prevent wanting to stop on newline but the prompt contains some
+        generated_text = tokens_string.replace(self.prompt,'')
+        # Check if the target sequence appears in the generated text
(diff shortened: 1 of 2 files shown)


awni:
Personally I think a better solution for this kind of thing is to use `generate_step` and do the user-specific logic in a user-sepcific `generate` function.  As you can see our `generate` function is already on the overly complex side and adding another option has a mutiplicative effect on complexity IMO.

Is there anything we could do to make `generate_step` more user friendly as a standalone API?

sderygithub:
I think the generate_step function is fine as it is; to my mind it depends on how much you want to converge for parity with HF (or even OpenAI who also has similar user-facing "stopping sequence"). Not a big deal either way (and I agree with your concern for complexity), users can easily implement wha they need :)

awni:
> to my mind it depends on how much you want to converge for parity with HF (or even OpenAI who also has similar user-facing "stopping sequence"). 

Well to be honest there is no plan there. I think it would be nice to have the top few most useful features but not the tail features in this repo. That way it's a good compromise between easily extensible and also pretty useful standalone.

If there are some common accepted standards then that's a good case for including them here.  I wasn't aware that passing in a sequence of tokens as a stopping criteria was common, do you happen to know if it's a frequently used feature?

sderygithub:
Make sense to me. I don't know how popular this criteria is. My feelings tell me it's nowhere near "MaxToken" :)

HF [doesn't seem](https://github.com/huggingface/transformers/blob/main/src/transformers/generation/stopping_criteria.py
) to have implemented the StopSequence once. 
OpenAI has it on their [FAQ page](https://help.openai.com/en/articles/5072263-how-do-i-use-stop-sequences)

🤷 

IIWY I'd raise an issue and let others +1 the problem organically; I don't see a particular reason why this needs to be implemented now :)

mzbac:
Stop sequence is a kind of application-level concern. To implement it correctly, you may need to create your own custom generate function to intercept the generation steps. However, building one is quite easy if you understand how mlx works. I think providing an example on how to implement it would be more suitable for this case. I have already built one in the PR for adding OpenAI API server. Hopefully, that can give people a taste of how they can implement their own stop sequence functionality.

awni:
Closing for now! We can revisit once we have a better sense of priority / integration strategy.

PR #427, opened 2024-02-09
LoRA: Improve validation error for LoRA layer count exceeding model layer

madroidmaq:
This commit enhances the error handling when the specified LoRA layer count exceeds the total number of layers in the model. It clarifies the error message to provide actionable feedback for users, guiding them to adjust their input parameters accordingly.

### DIFF
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -128,8 +128,8 @@ def train(
     loss: callable = default_loss,
 ):
     # Create checkpoints directory if it does not exist
-    if not os.path.exists('checkpoints'):
-        os.makedirs('checkpoints')
+    if not os.path.exists("checkpoints"):
+        os.makedirs("checkpoints")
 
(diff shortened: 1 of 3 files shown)


awni:
Great call! Would you mind rebasing this on main? Probably should to move this validation to https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/tuner/utils.py#L10 and refactor slightly to handle models with different namings. E.g. you could make it a function and call it at the top of each branch

madroidmaq:
> Great call! Would you mind rebasing this on main? Probably should to move this validation to https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/tuner/utils.py#L10 and refactor slightly to handle models with different namings. E.g. you could make it a function and call it at the top of each branch

@awni I've re-tooled it based on the latest branching of the main, so take a look and see if there's anything you need to tweak further.

awni (approved):
Thank you!!

Issue #430, opened 2024-02-10
Saved checkpoints are overwritten by the final model / following checkpoints


yukiarimo:
How to make checkpoints on every 25 steps save each to a directory like `/checkpoint/iter_num_adapters.npz` instead of overwriting everything into `adapters.npz`

ivanfioravanti:
@awni  I'll fix this one. There is also a bug that save adapters only after eval counter is reached.

yukiarimo:
Hey, I see that you've changed something. But it is still not working. I'm worried about this particular file that I was using: https://github.com/ml-explore/mlx-examples/blob/main/lora/lora.py

It's the same!

awni:
Ah the update was to MLX LM, which you should use for more feature rich fine-tuning: [docs here](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/LORA.md).

The basic Lora example is not going to change much from here on out in the interest of keeping it simple / educational / hackable.

yukiarimo:
Also, will the save fusing method and GGUFing work for that approach?

awni:
Although you are right, the help is broken :(

```
python -m mlx_lm.lora --help
Traceback (most recent call last):
  File "/Users/awnihannun/miniconda3/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/awnihannun/miniconda3/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/awnihannun/repos/mlx-examples/llms/mlx_lm/lora.py", line 6, in <module>
    import mlx.optimizers as optim
  File "/Users/awnihannun/repos/mlx/python/mlx/optimizers/__init__.py", line 3, in <module>
    from mlx.optimizers.optimizers import *
  File "/Users/awnihannun/repos/mlx/python/mlx/optimizers/optimizers.py", line 10, in <module>
    class Optimizer:
  File "/Users/awnihannun/repos/mlx/python/mlx/optimizers/optimizers.py", line 123, in Optimizer
    def _maybe_schedule(self, name: str, param : Union[Callable, Any]):
NameError: name 'Any' is not defined
```

awni:
Basically: `pip install mlx-lm`

Then:

```shell
python -m mlx_lm.lora --help            
usage: lora.py [-h] [--model MODEL] [--max-tokens MAX_TOKENS] [--temp TEMP]
[...]

PR #431, opened 2024-02-11
Add checkpoints directory for adapter weights

ivanfioravanti:
The code was modified to create a checkpoints directory if it doesn't exist yet. Adapter weights are now saved to this checkpoints directory during the training iterations. Corrected indentation of Save adapter weights code because it was part of "if eval".

To implement checkpoints asked in #430 and fix a bug in the save adapters file.


ivanfioravanti:
@awni tested locally during fine tuning, ready for your review.

awni:
Thanks for adding this @ivanfioravanti and good catch with that indentation bug!

awni (approved):

PR #433, opened 2024-02-12
Clean up loss function extraction

Harrolee:
loss_fn(), originally loss(), was extracted from the Transformer class. References to the original loss function were not updated.


Harrolee:
resolves issue [434](https://github.com/ml-explore/mlx-examples/issues/434)

awni (approved):
Thank you!

PR #443, opened 2024-02-15
Change gqa to use repeat instead of concatenate

angeloskath:
Simplify repeat and provide ~1.5 tps for quantized mistral 7B.


awni (approved):
Simpler and faster, very nice!

I aso see the perils here of having a different implementation file for each model

PR #446, opened 2024-02-16
fix: check LoRA layers number error

madroidmaq:
The parameter was passed incorrectly when checking the number of LoRA layers.

Related commit：#427


awni (approved):
oops, thanks for the fix!

PR #461, opened 2024-02-19
LlaVA in MLX

nkasmanoff:
Hi, please see the following for what I’ve made so far for converting a Llava checkpoint, (https://huggingface.co/llava-hf/llava-1.5-7b-hf) into an MLX implementation. 

I chose this model because it was easy to load its accompanying implementation on HF, and compare the architecture / model weights side by side. 

There’s a few other choices I’ve made to make this easier on the first pass, such as combining all of the safe-tensors shards into a single PyTorch file, that way I could lean on some prior MLX examples implementations. Obviously this could be a bit restrictive as the LlaVA models get bigger and this file size > RAM, but that is one of the several issues I wanted to point out in this draft I can hopefully get some help in fixing.

There are several TODOs in the repo showing these issues. I think most urgently, something I am confused by is how to reconcile the supposed # of attention heads (32) with what I’m saving from the downloaded weights. 

I tried to outline my work in the notebook ‘Local Llava’ (can delete in the official PR) which is where I’d suggest anyone able to help get started from. 

The other major TODO I also have left is figuring out how to do the forward pass & model.generate that takes into account text and image inputs, but figure that can wait until confirmation the model is actually loaded correctly.

Improved documentation, README, and tests to come along afterwards.

Thank you in advance to anyone who can help!

Feel free to close if another implementation comes and looks further along :-)

P.S: Another nice to have for this would be the ability to use other multi modal variants that accept different base models such as Phi and Mistral.


mzbac:
Nice initiative. I am also attempting to port the code to MLX, but I have not made any progress yet. I am happy to help in any way that I can, just clarify the issue you mentioned regarding reconciling the supposed number of attention heads (32). Could you provide more details on that?

nkasmanoff:
@mzbac Thank you! Appreciate any help, or if this gives you a way to start a few steps ahead on your version of the port. Will explain my observation below:

According to the model config, the base LLM is [Vicuna 7b 1.5](https://huggingface.co/lmsys/vicuna-7b-v1.5
), a fine-tuned version of Llama 2. 

From that information, I used the config file of that model to fill out what the base LLM of LlaVa should be: 
Pasted Graphic

However, when I do so, I get the following error for loading in the language model weights:

Screenshot 2024-02-20 at 8 00 28 AM

What this tells me is that the pre-trained weights have a shape of 4096 x 4096, whereas our implementation of Llama2 uses the following:

Pasted Graphic 1

Where 4096 (dim size) * 32 (num heads) = 131072, which leads me to believe there is something wrong either with how I saved the pre-trained weights, or some mismatch in how Torch saved the attention heads versus here. 

If I were to update the model config, set num_heads = 1, this size mismatch won't happen. It's possible this is then actually correct, but I haven't gotten anywhere in the forward pass & inference section of the port which would quickly prove what's working or not.

nkasmanoff:
@mzbac that makes a ton of sense, thank you for the point it looks like the weights now load in properly! A ton of this code will inevitably need to be cleaned up before merging but next up in my mind is getting the processor and inference working, I'll share any updates I have on that when I can

nkasmanoff:
@mzbac tracking here, looks like there's a decent chunk of tasks remaining. Not sure when I'll be able to get to them all, so for anyone reading this and curious (I think) these are the remaining steps we'd need to do:

For the forward pass:
- Update the clip vision encoder output, it should use a specific hidden state specified in the config
- Figure out how to merge vision embeddings into the text tokens (see https://github.com/huggingface/transformers/blob/4f09d0fd888dbf2660313f9715992822acfb99ce/src/transformers/models/llava/modeling_llava.py#L279)


Once done, this should then be able to be converted to actual generation and not just a single ouput, i.e. port over model.generate. Haven't looked this code over yet so I can't say what sub-tasks remain.


And then there's the general housekeeping. Deleting unused classes, finding speedups, etc. 

I'd imagine it's also valuable to port over the text and image processor like was done for CLIP. That's something else that'd be great to add in.

mzbac:
Yeah, I am currently working on making the clip model compatible with HF format and updating the model to support output hidden states. I think I can complete that by the end of this week. If you're okay with it, you can leave that task to me.
FYI: https://github.com/ml-explore/mlx-examples/pull/472

nkasmanoff:
@mzbac I just looked at your CLIP PR and looks great! Once that is merged I can sync my fork with main, and update the forward pass for llava. I still need to do some more research on how the full forward pass is implemented to make sure nothing else like this "surprises" me, that may take some time, so if you (or anyone reading) also feels comfortable trying your hand at this too it's no problem.

mzbac:
Yeah, I started to look at Llava's forward pass and got some understanding of how it works. However, I will set up some tests to ensure there are no surprises in our implementation.

mzbac:
@nkasmanoff, while I was trying to set up the test, I noticed that I had to refactor quite a bit of code in order to be able to run the test and load the model. After writing two tests for image features and merging image features with input ids into input embeddings, most of the LLava model was completed (the rest is just a normal forward pass for llama LLM). I ended up finishing the llava model forward pass. Here is the complete code (https://github.com/mzbac/mlx-examples/tree/llava/llava), let me know if you would like to sync it back to your branch. 
Note: The crucial part of the Llava model is merging image features with input IDs, which I had to implement using NumPy due to MLX not supporting boolean indexing. However, this operation only occurs once before generation, so I hope the performance won't be too bad.

awni:
I'm really excited about the progress on this implementation!

> Note: The crucial part of the Llava model is merging image features with input IDs, which I had to implement using NumPy due to MLX not supporting boolean indexing. 

Could you point me to the code for that? Maybe there is a workaround using MLX still.

mzbac:
> I'm really excited about the progress on this implementation!
> 
> > Note: The crucial part of the Llava model is merging image features with input IDs, which I had to implement using NumPy due to MLX not supporting boolean indexing.
> 
> Could you point me to the code for that? Maybe there is a workaround using MLX still.

Yeah, here is the merged image feature and input embedding part: https://github.com/mzbac/mlx-examples/blob/llava/llava/llava.py#L94-L131.

nkasmanoff:
> @nkasmanoff, while I was trying to set up the test, I noticed that I had to refactor quite a bit of code in order to be able to run the test and load the model. After writing two tests for image features and merging image features with input ids into input embeddings, most of the LLava model was completed (the rest is just a normal forward pass for llama LLM). I ended up finishing the llava model forward pass. Here is the complete code (https://github.com/mzbac/mlx-examples/tree/llava/llava), let me know if you would like to sync it back to your branch. Note: The crucial part of the Llava model is merging image features with input IDs, which I had to implement using NumPy due to MLX not supporting boolean indexing. However, this operation only occurs once before generation, so I hope the performance won't be too bad.

@mzbac looks very good to me! I haven't made any major changes on my end, so I think syncing your changes would be ideal. Happy to test those changes on my end too.

mzbac:
Here is an ASCII diagram showing how Llava works at a high level, hope it can help understand the implementation easier.
```sh
Initial Setup:
+-------------------+         +---------------+
| Text Sequence     |         | Raw Images    |
| [T1, <IMG>, T2,   |         | [Image1,      |
|  T3, <IMG>, T4]   |         |  Image2]      |
+-------------------+         +---------------+

Step 1: Convert Text and <IMG> Tokens to Embeddings
+---------------------------------------------------------+
| Text and <IMG> Token Embedding Model                    |
|                                                         |
| [T1, <IMG>, T2, T3, <IMG>, T4]                          |
|           |                                             |
|           V                                             |
| [T1_emb, IMG_emb, T2_emb, T3_emb, IMG_emb, T4_emb]      |
+---------------------------------------------------------+

Step 2: Convert Images to Feature Patches Using Vision Model
+------------------------------------------------------------+
| Vision Model                                               |
|                                                            |
| Image1 ---> [I1_1, I1_2, I1_3]                             |
| Image2 ---> [I2_1, I2_2, I2_3]                             |
+------------------------------------------------------------+

Step 3: Convert Image Patches to Embeddings
+------------------------------------------------------------+
| Image Patch Embedding Conversion                           |
|                                                            |
| [I1_1, I1_2, I1_3] ---> [I1_1_embed, I1_2_embed, I1_3_embed]|
| [I2_1, I2_2, I2_3] ---> [I2_1_embed, I2_2_embed, I2_3_embed]|
+------------------------------------------------------------+

Step 4: Replace IMG_emb with Image Patch Embeddings in Sequence
+------------------------------------------------------------+
| Updated Sequence Embeddings                                |
|                                                            |
| [T1_emb, I1_1_embed, I1_2_embed, I1_3_embed, T2_emb,       |
|  T3_emb, I2_1_embed, I2_2_embed, I2_3_embed, T4_emb]       |
+------------------------------------------------------------+

Step 5: Feed the Updated Sequence into the LLM
+------------------------------------------------------------+
|                   Large Language Model                     |
|                                                            |
| Input: [T1_emb, I1_1_embed, I1_2_embed, I1_3_embed,        |
|         T2_emb, T3_emb, I2_1_embed, I2_2_embed, I2_3_embed,|
|         T4_emb]                                            |
|                                                            |
|                        |                                   |
|                        V                                   |
|                Output: Model Predictions                   |
+------------------------------------------------------------+

```

nkasmanoff:
@awni quick observation:

Since GeLU is now supported in MLX, I'm working to use the library example rather than the activation defined here. However, there's a minor discrepancy between the two, for the fast approximation.

In this (& CLIP example) the approximation is 

`x * mx.sigmoid(1.702 * x)`, but [here](https://github.com/ml-explore/mlx/blob/e6418781abd4232fcb3613cc300a025ccf5f405b/python/mlx/nn/layers/activations.py#L192) it's `x * mx.sigmoid(1.773 * x)` 

Off of some test examples it seems like 1.702 gives better performance, and searching online also seems to back up 1.702. I'd still like to use the native implementation rather than my own, but I think we should update that code too if it's okay.

llava/llava.py
```diff
@@ -69,22 +69,27 @@ def get_input_embeddings(
...
         inputs_embeds = self.language_model.model.embed_tokens(input_ids)
+        # get the ouptut hidden states from the vision model
         _, _, hidden_states = self.vision_tower(
             pixel_values.transpose(0, 2, 3, 1), output_hidden_states=True
         )
+        # select the hidden states from the desired layer
         selected_image_feature = hidden_states[self.vision_feature_layer]
 
         if self.vision_feature_select_strategy == "default":
+            # default strategy is to select all the hidden states except the first one (CLS token?)
```
mzbac, line 82: The first one is the embedding -> https://github.com/ml-explore/mlx-examples/blob/main/clip/model.py#L280

awni:
> Off of some test examples it seems like 1.702 gives better performance, and searching online also seems to back up 1.702. I'd still like to use the native implementation rather than my own, but I think we should update that code too if it's okay.

I don't know why we wouldn't have the standard approximation there. Would be great if you send a PR to fix it. Would be helpful if you include some reference that implements it with 1.702.

nkasmanoff:
@mzbac 

Updated the PR name as it's a bit further along and I think closer to merging. 

This was a very useful link. Mind if I paste it and add as reference in the repo? Would love to give you credit for this (among the many other contributions you made) https://github.com/ml-explore/mlx-examples/pull/461#issuecomment-1962807987

I'm also considering adding an additional file, a lightweight LoRA example, which fine tunes the language model portion of Llava on this dataset: https://huggingface.co/datasets/HuggingFaceM4/LLaVA_Wild_Modif. Think this is needed for the first merge?

awni:
I'm following along here, very excited about this PR. Let us know when it's readyish for review.

mzbac:
@nkasmanoff, thank you for the update! I appreciate the link being used as a reference and am excited about the progress on the PR. Regarding the LoRA example, I think we may need to address merging image features input ids using numpy instead of mlx. I'm not sure if we can fine-tune LoRA with the current np implementation. Maybe we could get some help from @awni.

nkasmanoff:
@awni I think it is! I saw @mzbac as the reviewer hence the tag, but please take a look too :-)

llava/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Setup:
+
+Install the dependencies:
+
+```bash
+pip install -r requirements.txt
+```
+
+## Run
```
mzbac, line 13: Would we mention that the can use the generate.py script as the cli as well? That may make it easier for them to test things out without having to code the script.

llava/language.py
```diff
@@ -0,0 +1,238 @@
...
+            )
+        self.model = Llama(config)
+        self.lm_head = nn.Linear(
+            config.hidden_size, config.vocab_size, bias=False)
+
+    def __call__(
+        self,
+        inputs: mx.array,
+        cache=None,
+        inputs_embeds=None,
```
mzbac, line 224: Most of the implementation is copied from mlx-lm's llama, with only updates made to the forward pass to allow for directly passing inputs_embeds for the initial prompt evaluation.

llava/llava.py
```diff
@@ -0,0 +1,196 @@
...
+            non_image_indices
+        ]
+
+        image_to_overwrite = np.all(final_embedding == 0, axis=-1)
+        reshaped_image_features = image_features.reshape(-1, embed_dim)
+        final_embedding[image_to_overwrite, :] = reshaped_image_features[
+            : np.sum(image_to_overwrite)
+        ]
+
+        return mx.array(final_embedding)
```
mzbac, line 150: @awni, this is the part where we have to use NumPy as a workaround because MLX doesn't support boolean indexing. Maybe you could give us some pointers on how we can implement this using MLX.

mzbac: 1, you are right. I skipped the error checking for mismatched image tokens and image batches to simplify the implementation, but we could definitely add it. FYI: https://github.com/huggingface/transformers/blob/main/src/transformers/models/llava/modeling_llava.py#L333-L337
2. My understanding is that as long as the total number of images matches the total image tokens in a batch, it should work with batches. Given the final embedding is batched, however, I have not done any batch testing.

awni: Regarding 2, makes sense! Does the rest of the code work for batched text? E.g. the generation part properly handles different text examples? If it's just using our MLX LM llama implementation (which it looks like) then the answer is probably no. I don't know if that simplifies much though.

mzbac: Correct me if I am wrong, but I thought the Llama example should be able to handle batched text generation. The only difference is that the initial prompt will get the embedding with text and image from the code above and directly go to LLM's encode layer without going through the embedding/norm layer. Once that forward pass is completed, we get the next token logits and key-value cache, then the rest of process is just a standard forward pass for LLM.

mzbac:
```
    def _merge_input_ids_with_image_features(
        self, image_features, inputs_embeds, input_ids
    ):
        image_token_index = self.config.image_token_index
        num_images, num_image_patches, embed_dim = image_features.shape

        # Positions of <image> tokens in `input_ids` and assume batch size is 1
        image_positions = [
            idx
            for idx, token_id in enumerate(input_ids[0])
            if token_id == image_token_index
        ]

        if len(image_positions) != num_images:
            raise ValueError(
                f"The input provided to the model are wrong. The number of image tokens is {len(image_positions)} while"
                f" the number of image given to the model is {num_images}."
            )

        text_segments = []
        start_idx = 0

        # Split the input embeddings into text segments based on the positions of <image> tokens
        for position in image_positions:
            text_segments.append(inputs_embeds[:, start_idx:position])
            start_idx = position + 1

        # Add the remaining text embeddings after the last <image> token
        if start_idx < inputs_embeds.shape[1]:
            text_segments.append(inputs_embeds[:, start_idx:])

        # Reshape image feature from (num_images, num_image_patches, embed_dim) to (num_images*num_image_patches*, embed_dim)
        image_embeddings = image_features.reshape(-1, image_features.shape[-1])

        final_embeddings = []
        for i, text_segment in enumerate(text_segments):
            final_embeddings.append(text_segment[0])
            if i < len(image_positions):
                # Add a slice of image embeddings corresponding to the current position.
                # This effectively replaces one <image> token with its associated image embeddings.
                final_embeddings.append(
                    image_embeddings[i : i + num_image_patches + 1]
                ) 

        # This creates a final embeding in shape (1, num_image_patches + num_text_tokens, embed_dim) representing the merged sequence of text and image embeddings.  
        final_embeddings = mx.concatenate(final_embeddings, axis=0).reshape(
            1, -1, embed_dim
        )

        return final_embeddings
```

awni: @mzbac are you going to add the snippet you shared?

mzbac: @awni, I don't have permission to push the changes. Maybe either you or @nkasmanoff could update it.

awni: 😬 sorry. Yea happy to, or @nkasmanoff can give you temporary permissions to push to his fork.

nkasmanoff: @mzbac just gave you permission, let me know if that worked!

mzbac: @nkasmanoff @awni I have pushed the changes. Please note that the test.py will fail due to the fast_gelu to native gelu change, but if you build the latest mlx locally, the test will pass.

awni: Hmm, we should put the version in the requirement in that case. Should work with `0.5` and up.

awni:
Also it looks like this is almost ready to go. I will review a bit more tonight, but from what I looked at, it looked quite nice already!

nkasmanoff:
> Also it looks like this is almost ready to go. I will review a bit more tonight, but from what I looked at, it looked quite nice already!

@awni this is great to hear!

It depends if you think this is essential to this PR, but some ideas for additional features which are missing now that may be useful later on.

- Adding a MLX native processor + tokenizer, rather than lean on transformers.
- Make possible to use other variants of LlaVA, or VLMs.
- Fine tuning (discussed previously) demo.


My opinion is we don't need any for first example, but is there anything else to consider?

@mzbac

awni:
Thanks for the list @nkasmanoff,  also think we can leave most of it as follow up depending on which direction this example goes.

awni (approved):
🚀 

Really great addition, thank you @mzbac and @nkasmanoff !

nxphi47:
@nkasmanoff @mzbac Great works! Do you plan to support [Llava-1.6-hf](https://huggingface.co/llava-hf/llava-v1.6-mistral-7b-hf) too ? 
It does require this change from transformers: https://github.com/huggingface/transformers/pull/29012

mzbac:
> @nkasmanoff @mzbac Great works! Do you plan to support [Llava-1.6-hf](https://huggingface.co/llava-hf/llava-v1.6-mistral-7b-hf) too ? It does require this change from transformers: [huggingface/transformers#29012](https://github.com/huggingface/transformers/pull/29012)

@nxphi47 would like to add support for llava 1.6, however, could you create an issue for that? Comments on the closed PR may get lost.

jrp2014:
1.6 seems to need an image_newline model parameter. It doesn’t seem to be a straight drop in replacement for 1.5

jrp2014:
Is v1.6 support abandoned, or are some of the newer mlx features helpful to getting it up and running?

awni:
I think you can run it in [MLX VLM](https://github.com/Blaizzy/mlx-vlm), CC @Blaizzy

Blaizzy:
Thanks @awni!

[Llava 1.6 (Next)](https://github.com/Blaizzy/mlx-vlm/issues/39) is coming to MLX VLM.

Alongside a [trainer](https://github.com/Blaizzy/mlx-vlm/pull/41/files).

PR #468, opened 2024-02-20
Bug fix in lora.py

SHSE:
It fails on loading the dataset because `load_and_check` doesn't return anything.


awni (approved):
Oops, thanks for the fix!!

Issue #465, opened 2024-02-20
unable to run mixtral


jbrodriguez:
i'm getting the following error

```
~/tmp/mlx-examples/llms/mixtral main* 5m
❯ python3 mixtral.py --model-path mlx_model
[INFO] Loading model from disk.
[INFO] Starting generation...
In the beginning the Universe was created.Traceback (most recent call last):
  File "<redacted>/tmp/mlx-examples/llms/mixtral/mixtral.py", line 324, in <module>
    for token, _ in zip(generate(prompt, model, args.temp), range(args.max_tokens)):
  File "<redacted>/tmp/mlx-examples/llms/mixtral/mixtral.py", line 275, in generate
    logits, cache = model(prompt[None])
                    ^^^^^^^^^^^^^^^^^^^
  File "<redacted>/tmp/mlx-examples/llms/mixtral/mixtral.py", line 213, in __call__
    h, cache[e] = layer(h, mask, cache[e])
                  ^^^^^^^^^^^^^^^^^^^^^^^^
  File "<redacted>/tmp/mlx-examples/llms/mixtral/mixtral.py", line 177, in __call__
    r, cache = self.attention(self.attention_norm(x), mask, cache)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<redacted>/tmp/mlx-examples/llms/mixtral/mixtral.py", line 106, in __call__
    queries = self.rope(queries)
              ^^^^^^^^^^^^^^^^^^
  File "<redacted>/tmp/mlx-examples/llms/mixtral/mixtral.py", line 51, in __call__
    costheta, sintheta = RoPE.create_cos_sin_theta(
                         ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: type object 'RoPE' has no attribute 'create_cos_sin_theta'
```

these are the contents of my mlx_model folder

```
❯ l mlx_model/
.rw-r--r--  354 <redacted> staff 20 Feb 12:06 -N config.json
.rw-r--r-- 493k <redacted> staff 20 Feb 12:01 -N tokenizer.model
.rw-r--r-- 3.6G <redacted> staff 20 Feb 12:01 -N weights.0.npz
.rw-r--r-- 3.6G <redacted> staff 20 Feb 12:02 -N weights.1.npz
.rw-r--r-- 3.6G <redacted> staff 20 Feb 12:02 -N weights.2.npz
[...]

PR #469, opened 2024-02-20
Update RoPE initialization in mixtral.py

ivanfioravanti:
Fix for issue #465  
The RoPE initialization in the mixtral.py file has been updated. Previously, RoPE was being instantiated directly, but now it has been switched to be instantiated through the nn module.

### DIFF
--- a/llms/mixtral/mixtral.py
+++ b/llms/mixtral/mixtral.py
@@ -40,26 +40,6 @@ class RMSNorm(nn.Module):
         return self.weight * output
 
 
-class RoPE(nn.RoPE):
-    def __init__(self, dims: int, traditional: bool = False):
-        super().__init__(dims, traditional)
-
-    def __call__(self, x, offset: int = 0):
-        shape = x.shape
-        x = mx.reshape(x, (-1, shape[-2], shape[-1]))
-        N = x.shape[1] + offset
-        costheta, sintheta = RoPE.create_cos_sin_theta(
-            N, self.dims, offset=offset, base=1000000, dtype=x.dtype
-        )
-
-        rope = (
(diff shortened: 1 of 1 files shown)


llms/mixtral/mixtral.py
```diff
@@ -76,7 +56,7 @@ def __init__(self, args: ModelArgs):
         self.wk = nn.Linear(args.dim, args.n_kv_heads * args.head_dim, bias=False)
         self.wv = nn.Linear(args.dim, args.n_kv_heads * args.head_dim, bias=False)
         self.wo = nn.Linear(args.n_heads * args.head_dim, args.dim, bias=False)
-        self.rope = RoPE(args.head_dim, traditional=True)
+        self.rope = nn.RoPE(args.head_dim, traditional=True)
```
angeloskath, line 59: In order for this to match the above we should probably pass the base parameter to RoPE. ie `nn.RoPE(args.head_dim, base=1000000, traditional=True)`.

angeloskath (changes requested):
Thanks so much for doing that! Did you by any chance test that the generations still make sense?

Left a small comment for a forgotten base argument.

llms/mixtral/mixtral.py
ivanfioravanti: It was working well even with base Mistral value 🤔

angeloskath: Hm interesting. Let me go search which one should be there and then I 'll merge this. Thanks!

ivanfioravanti: 1000000 is the right one, but please use #470 from @jbrodriguez

angeloskath:
Thanks @ivanfioravanti for the fix and for giving it up to @jbrodriguez :-)

PR #482, opened 2024-02-23
LoRA: Extract the run function for easy use in scripts file

madroidmaq:
Extracting the logic from `if __name__ == "__main__":` into a separate `run` function makes it easy to call the logic in the `lora.py` file from your own python scripts.

Below is a simple code example of my training scenario:

```py
import argparse
from mlx_lm import lora
from mlx_lm.tuner.trainer import TrainingCallback

def finetuning():
    # build args
    parser = lora.build_parser()
    args = parser.parse_args(args=[])
    args.model = "Qwen/Qwen1.5-1.8B-Chat"
    args.train = True

    # training
    wandb_callback = WandbCallback()
    lora.run(args, training_callback = wandb_callback)

if __name__ == "__main__":
   finetuning()
```

In addition to the command line, it's finally possible to fine-tune it by writing code.

### DIFF
--- a/llms/mlx_lm/lora.py
+++ b/llms/mlx_lm/lora.py
@@ -7,7 +7,7 @@ import mlx.optimizers as optim
 import numpy as np
 from mlx.utils import tree_flatten
 
-from .tuner.trainer import TrainingArgs, evaluate, train
+from .tuner.trainer import TrainingArgs, TrainingCallback, evaluate, train
 from .tuner.utils import linear_to_lora_layers
 from .utils import generate, load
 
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/lora.py
```diff
@@ -160,10 +160,7 @@ def load_dataset(args):
     return train, valid, test
 
 
-if __name__ == "__main__":
-    parser = build_parser()
-    args = parser.parse_args()
-
+def run_lora(args, training_callback: TrainingCallback = None):
```
awni, line 163: Ok naming this `run` instead of `run_lora`? I think it will be slightly more flexible if we ever have multiple tuning strategies, and it's already clear it's for lora since it will be `lora.run`.

awni (commented):
Nice! Change looks good.  Just a minor comment then we can merge!

madroidmaq:
> Nice! Change looks good. Just a minor comment then we can merge!

@awni  It's DONE.

awni (approved):

PR #486, opened 2024-02-25
Add top-p sampling for text generation

peterjc123:
Implementation adapted from [here](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/server.py#L65-L87)


awni:
Thanks for the contribution!

@mzbac what do you think, should we include this here? It might make sense to have the HTTP server call this function since the two generates are almost identical?

mzbac:
Yeah, it makes sense to move the top_k into the generate_step and update the server script to use the generate_step directly. The reason I didn't use it when creating the server script was because the `repetition_penalty` was pending, trying to avoid conflict. I will update the server script once this is merged.

awni (approved):
Thank you!

Issue #478, opened 2024-02-22
mlx_lm.server  OpenAI compatible endpoints


namp:
Currently mlx_lm.server only implements a single endpoint: /v1/chat/completions 

It is essential to add more OpenAI compatible endpoints similarly to the implementation of the llama_cpp.server:

https://github.com/abetlen/llama-cpp-python/tree/main/llama_cpp/server

mzbac:
Just out of curiosity, what is the other endpoint you are more interested in getting support from mlx-lm?

namp:
v1/completions

Thanks!

PR #489, opened 2024-02-26
chore: add /v1/completions for server.py

mzbac:
Partially resolve https://github.com/ml-explore/mlx-examples/issues/478.

### DIFF
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -4,7 +4,7 @@ import time
 import uuid
 from collections import namedtuple
 from http.server import BaseHTTPRequestHandler, HTTPServer
-from typing import List, Optional, Tuple
+from typing import Callable, List, Optional
 
 import mlx.core as mx
 import mlx.nn as nn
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/server.py
```diff
@@ -46,7 +46,7 @@ def stopping_criteria(
             and how many tokens should be trimmed from the end if it has (`trim_length`).
     """
     if tokens and tokens[-1] == eos_token_id:
-        return StopCondition(stop_met=True, trim_length=0)
+        return StopCondition(stop_met=True, trim_length=1)
```
mzbac, line 49: Avoid sending the EOS token back to the client. Otherwise, we will have to manually enable skipping special tokens in decoding (https://huggingface.co/docs/transformers/main_classes/tokenizer#transformers.PreTrainedTokenizer.decode.skip_special_tokens)

awni (approved):
Looks great, thanks for the addition!

I'm not entirely sure what the purpose of the completions is over the chat endpoint (it seems like OpenAI is deprecetating the raw completions?).

mzbac:
> Looks great, thanks for the addition!
> 
> I'm not entirely sure what the purpose of the completions is over the chat endpoint (it seems like OpenAI is deprecetating the raw completions?).

Sometimes, users may prefer to use a raw prompt that they can handle the formatting by themselves, which makes sense to use the completions endpoint over chat completions. This is mostly for custom SFT models, I guess.

Issue #488, opened 2024-02-26
mlx_lm.lora need an option to set --eos-token


southkorea2013:
Hi,

Please help to add an option "-eos-token" to run a LORA fine-tuned Qwen llm with mlx_lm.lora. 

Thanks,
Nan

southkorea2013:
If I used the fused model with mlx_lm.generate, I found it cannot answer questions with the training datasets, but it works with mlx_lm.lora with the adapters.npz. Thanks.

mzbac:
> If I used the fused model with mlx_lm.generate, I found it cannot answer questions with the training datasets, but it works with mlx_lm.lora with the adapters.npz. Thanks. Are there any options I need to modify for running the mlx_lm.fuse ?

Yeah, a fused model will experience performance degradation due to the quant-dequant operation. Therefore, making the generate command to be able to load the base model and adapters would be a better solution.

southkorea2013:
> > If I used the fused model with mlx_lm.generate, I found it cannot answer questions with the training datasets, but it works with mlx_lm.lora with the adapters.npz. Thanks. Are there any options I need to modify for running the mlx_lm.fuse ?
> 
> Yeah, a fused model will experience performance degradation due to the quant-dequant operation. Therefore, making the generate command to be able to load the base model and adapters would be a better solution.

Thanks.

Solido:
Hi!
Is there a benchmark to evaluate the perplexity degradation for quantization?

PR #494, opened 2024-02-27
chore(mlx-lm): add adapter support in generate.py

mzbac:
close https://github.com/ml-explore/mlx-examples/issues/488

### DIFF
--- a/llms/mlx_lm/LORA.md
+++ b/llms/mlx_lm/LORA.md
@@ -74,17 +74,7 @@ python -m mlx_lm.lora \
 
 ### Generate
 
-1. Using the lora module for basic generation:
-
-```
-python -m mlx_lm.lora \
-    --model <path_to_model> \
-    --adapter-file <path_to_adapters.npz> \
-    --max-tokens 50 \
-    --prompt "<your_model_prompt>"
-```
-
-2. For a feature-rich generation experience, use mlx_lm.generate:
+For generation use mlx_lm.generate:
 
 ```shell
 python -m mlx_lm.generate \
--- a/llms/mlx_lm/LORA.md
+++ b/llms/mlx_lm/LORA.md
@@ -72,6 +72,27 @@ python -m mlx_lm.lora \
     --test
 ```
 
+### Generate
+
+1. Using the lora module for basic generation:
+
+```
+python -m mlx_lm.lora \
+    --model <path_to_model> \
+    --adapter-file <path_to_adapters.npz> \
+    --max-tokens 50 \
+    --prompt "<your_model_prompt>"
+```
+
+2. For a feature-rich generation experience, use mlx_lm.generate:
+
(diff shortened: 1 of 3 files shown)


llms/mlx_lm/LORA.md
```diff
@@ -72,6 +72,27 @@ python -m mlx_lm.lora \
...
+### Generate
+
+1. Using the lora module for basic generation:
+
+```
+python -m mlx_lm.lora \
+    --model <path_to_model> \
+    --adapter-file <path_to_adapters.npz> \
+    --max-tokens 50 \
+    --prompt "<your_model_prompt>"
```
awni, line 84: What do you think about removing `generate` from the original lora script? We could just have it use `generate`. Would simplify things a bit

mzbac: Personally, I feel that using generate.py with an adapter is more natural. hence, it might not be a bad idea to remove generate from the lora script, but this could potentially cause issues for those who are used to fine-tuning testing with lora's generate. Perhaps we could consider starting to raise warnings in lora's generate and removing it in a later version?

awni: I don't think many people use the lora generate it so I'm ok just deleting it to simplify the process. I'm also ok with warning for now if you feel otherwise.

awni:
@mzbac I'm waiting for you on this right? Lmk when it's ready to merge.

mzbac:
> @mzbac I'm waiting for you on this right? Lmk when it's ready to merge.

Yeah, I will update it tonight and let you know when it's ready for merging.

mzbac:
@awni I have removed the generation support in lora.py, and it will now raise a "not implemented" error to prompt users to use mlx_lm.generate if they attempt to generate from lora.py. Let me know if there is anything else that needs to be changed.

llms/mlx_lm/LORA.md
```diff
@@ -72,6 +72,17 @@ python -m mlx_lm.lora \
     --test
 ```
 
+### Generate
+
+For generation use mlx_lm.generate:
```
awni, line 77:
```suggestion
For generation use `mlx_lm.generate`:
```

llms/mlx_lm/lora.py
```diff
@@ -234,15 +234,8 @@ def run(args, training_callback: TrainingCallback = None):
...
-        model.eval()
-        generate(
-            model=model,
-            tokenizer=tokenizer,
-            temp=args.temp,
-            max_tokens=args.max_tokens,
-            prompt=args.prompt,
-            verbose=True,
+        raise NotImplementedError(
+            "Please use mlx_lm.generate with trained adapter for generation."
```
awni, line 238: ❤️

awni (approved):
Awesome, much needed, thank you!

PR #495, opened 2024-02-27
Feature/mistral few shot

dhirlea:
Added few-shot prompt example for Mistral models


awni:
Thanks for the contribution. Our goal with that mistral example  is to keep it very simple and minimal so we aren't planning to merge additional "features" to it for the most part.

Issue #490, opened 2024-02-26
[mlx-lm] Add format option to saveweights 


awni:
E.g something like below so saved models can be used in Transformers library:

```
metadata={"format": "pt"})
```

See original issue in MLX: https://github.com/ml-explore/mlx/issues/743#issuecomment-1965427589

PR #496, opened 2024-02-27
Add metadata when saving safetensors

alexweberk:
Add metadata format="pt" for safetensors so that model's are accessible to `transformers` users as well.

Tested locally for loading via transformers on a M3 Max MacBook Pro.

https://github.com/ml-explore/mlx-examples/issues/490

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -498,7 +498,7 @@ def save_weights(
         shard_name = shard_file_format.format(i + 1, shards_count)
         shard_path = save_path / shard_name
 
-        mx.save_safetensors(str(shard_path), shard)
+        mx.save_safetensors(str(shard_path), shard, metadata={"format": "pt"})
 
         for weight_name in shard.keys():
             index_data["weight_map"][weight_name] = shard_name
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -498,7 +498,7 @@ def save_weights(
         shard_name = shard_file_format.format(i + 1, shards_count)
         shard_path = save_path / shard_name
 
-        mx.save_safetensors(str(shard_path), shard, metadata={"format": "pt"})
+        mx.save_safetensors(str(shard_path), shard, metadata={"format": "mlx"})
 
         for weight_name in shard.keys():
             index_data["weight_map"][weight_name] = shard_name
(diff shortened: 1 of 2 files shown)


llms/mlx_lm/utils.py
```diff
@@ -498,7 +498,7 @@ def save_weights(
         shard_name = shard_file_format.format(i + 1, shards_count)
         shard_path = save_path / shard_name
 
-        mx.save_safetensors(str(shard_path), shard)
+        mx.save_safetensors(str(shard_path), shard, metadata={"format": "pt"})
```
awni, line 501: I'm wondering if this should be an option rather than the default. On the one hand MLX doesn't really need that field so it's kind of unused otherwise. On the other hand it seems a bit strange to default the format to PyTorch whenever saving MLX models. Also we may want to use that field in the future..

Do you all know where does the format gets read in Transformers? Maybe we could ask them to add a MLX field and default the format to MLX instead?

alexweberk: Yeah, what you said actually makes a lot of sense. From the error I was getting, it looks like if the Transformers library added 'mlx' to the following line, and if MLX default saves safetensors files with format="mlx" this can be achieved.

https://github.com/huggingface/transformers/blob/63caa370e6c618dbe7d3fd4cbf545cc32eca1a15/src/transformers/modeling_utils.py#L507

awni: Would you be open to making an issue or PR for that and link it? I would definitely comment / +1 it.

alexweberk: Yeah, sure thing.

alexweberk: @awni I added an issue/PR on Transformers repo: 
- https://github.com/huggingface/transformers/issues/29333
- https://github.com/huggingface/transformers/pull/29335

I also made a commit to save the format in "mlx" instead per above discussion.

awni: Awesome, thanks!

awni (approved):
I'm happy to merge this. It's kind of the right format anyway.

awni:
@alexweberk could you check the format. Looks like you did not run the pre-commit hooks so the format test failed.

alexweberk:
@awni Apologies, you're right. I ran the pre-commit hooks now so hope it's fine now.

PR #498, opened 2024-02-27
Prevent llms/mlx_lm from serving the local directory as a webserver

Y4hL:
By default BaseHTTPRequestHandler serves the local directory as a website using the do_HEAD and do_GET methods. Should definitely not be the case. Although there are definitely bigger issues at hand if somebody runs these examples in production before looking at them, I'd still suggest having a security policy

Threw in some type hint fixes as a treat :)

### DIFF
--- a/ACKNOWLEDGMENTS.md
+++ b/ACKNOWLEDGMENTS.md
@@ -12,3 +12,4 @@ MLX Examples was developed with contributions from the following individuals:
 - Shunta Saito: Added support for PLaMo models.
 - Gabrijel Boduljak: Implemented `CLIP`.
 - Markus Enzweiler: Added the `cvae` examples.
+- Rasmus Kinnunen: Fixed a security hole in the `llms/mlx_lm` example
--- a/llms/mlx_lm/SERVER.md
+++ b/llms/mlx_lm/SERVER.md
@@ -4,7 +4,9 @@ You use `mlx-lm` to make an HTTP API for generating text with any supported
 model. The HTTP API is intended to be similar to the [OpenAI chat
 API](https://platform.openai.com/docs/api-reference).
 
-NOTE: mlx-lm is not recommended for production. It only implements basic security checks.
+> [!NOTE]  
+> The MLX LM server is not recommended for production as it only implements
+> basic security checks.
 
 Start the server with: 
 
--- a/llms/mlx_lm/SERVER.md
+++ b/llms/mlx_lm/SERVER.md
@@ -4,6 +4,8 @@ You use `mlx-lm` to make an HTTP API for generating text with any supported
 model. The HTTP API is intended to be similar to the [OpenAI chat
 API](https://platform.openai.com/docs/api-reference).
 
+NOTE: mlx-lm is not recommended for production. It only implements basic security checks.
+
 Start the server with: 
 
 ```shell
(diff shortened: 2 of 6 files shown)


mzbac:
mlx-lm server is definitely not intended for production use. You are right, we should add some warnings when launching the server.

mzbac:
@awni, do you think we can prioritize this PR since it's security-related and could potentially have a negative impact if people use the server in a production environment?

awni:
Yea I will review it shortly.

awni:
I will probably add a line to the README as well to be super clear about that.

Y4hL:
We could hardcode binding the server only to the local loopback, this way it forces people to take a closer look before using it for anything other than on-device testing.

mzbac:
> We could hardcode binding the server only to the local loopback, this way it forces people to take a closer look before using it for anything other than on-device testing.

The issue with hardcoded local loopback is that sometimes users may want to test using a different machine. For example, I usually run the server on my Mac Ultra but do testing on my MBP.

Y4hL:
Let's leave it at just the warning then

Y4hL:
@mzbac I'm thinking of refactoring the server a little. On the [OpenAI API Reference](https://platform.openai.com/docs/api-reference/completions/create) "text_completion" is marked as legacy. Do you think it would be suitable to remove it, for the sake of keeping the example as simple as possible?

mzbac:
The text completion endpoint was brought up from https://github.com/ml-explore/mlx-examples/issues/478. Sometimes users may want to use it with a custom prompt format instead of using the chat template or default role mapping, especially when they have some custom SFT fine-tuned models.

Y4hL:
> The text completion endpoint was brought up from #478. Sometimes users may want to use it with a custom prompt format instead of using the chat template or default role mapping, especially when they have some custom SFT fine-tuned models.

Okay I'll keep it in, thanks!

awni (approved):
Thanks for the fix, that's great!

M-I:
The curl example doesn't work. I tried reverting to before this commit and it started working.
I'm not sure what's the issue, the curl example is suppose to make a POST request, but in the traceback there was:

```
File "/Users/byboh/oss/mlx-examples/llms/mlx_lm/server.py", line 165, in __init__
    delattr(self, "do_HEAD")
```

Out of curiosity, what kind of request do you need to do to have access to the filesystem? With the "unsafe" revision, trying a GET to http://localhost:8080/README.md returns a 501 error and a POST a 404.

Is it just on my machine?

mzbac:
> The curl example doesn't work. I tried reverting to before this commit and it started working. I'm not sure what's the issue, the curl example is suppose to make a POST request, but in the traceback there was:
> 
> ```
> File "/Users/byboh/oss/mlx-examples/llms/mlx_lm/server.py", line 165, in __init__
>     delattr(self, "do_HEAD")
> ```
> 
> Out of curiosity, what kind of request do you need to do to have access to the filesystem? With the "unsafe" revision, trying a GET to http://localhost:8080/README.md returns a 501 error and a POST a 404.
> 
> Is it just on my machine?

I believe there was a mistake in mixing up the BaseHTTPRequestHandler and SimpleHTTPRequestHandler, so the old version does not actually expose the local folder via `GET`. The server.py is currently broken due to these changes. However, the server.py is not designed for production usage, so just be cautious when using it in production.

PR #501, opened 2024-02-28
Refactoring of mlx_lm example

Y4hL:
Looking for improvements in the mlx_lm example to simplify the code and better mimic the OpenAI API

### DIFF
--- a/ACKNOWLEDGMENTS.md
+++ b/ACKNOWLEDGMENTS.md
@@ -12,4 +12,3 @@ MLX Examples was developed with contributions from the following individuals:
 - Shunta Saito: Added support for PLaMo models.
 - Gabrijel Boduljak: Implemented `CLIP`.
 - Markus Enzweiler: Added the `cvae` examples.
-- Rasmus Kinnunen: Fixed a security hole in the `llms/mlx_lm` example
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -351,9 +351,7 @@ class APIHandler(BaseHTTPRequestHandler):
         stop_words = body.get("stop", [])
         stop_words = [stop_words] if isinstance(stop_words, str) else stop_words
         stop_id_sequences = [
-            _tokenizer.encode(stop_word, return_tensors="np", add_special_tokens=False)[
-                0
-            ]
+            TOKENIZER.encode(stop_word, return_tensors="np", add_special_tokens=False)[0]
             for stop_word in stop_words
         ]
         eos_token_id = _tokenizer.eos_token_id
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -255,7 +255,12 @@ class APIHandler(BaseHTTPRequestHandler):
                 break
 
         text = _tokenizer.decode(tokens)
-        return response_creator(response_id, requested_model, prompt, tokens, text)
+        response = response_creator(response_id, requested_model, prompt, tokens, text)
+        try:
+            self.wfile.write(f"{response}\n\n".encode())
+            self.wfile.flush()
+        except Exception as error:
+            print(error)
 
     def handle_stream(
         self,
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -356,40 +356,28 @@ class APIHandler(BaseHTTPRequestHandler):
         ]
         eos_token_id = _tokenizer.eos_token_id
         max_tokens = body.get("max_tokens", 100)
-        stream = body.get("stream", False)
         requested_model = body.get("model", "default_model")
         temperature = body.get("temperature", 1.0)
         top_p = body.get("top_p", 1.0)
         repetition_penalty = body.get("repetition_penalty", 1.0)
         repetition_context_size = body.get("repetition_context_size", 20)
-        if not stream:
-            return self.generate_response(
-                prompt,
-                chat_id,
-                requested_model,
-                stop_id_sequences,
-                eos_token_id,
-                max_tokens,
-                temperature,
-                top_p,
-                repetition_penalty,
-                repetition_context_size,
-                create_chat_response,
-            )
-        else:
-            self.handle_stream(
-                prompt,
-                chat_id,
-                requested_model,
-                stop_id_sequences,
-                eos_token_id,
-                max_tokens,
-                temperature,
-                top_p,
-                repetition_penalty,
-                repetition_context_size,
-                create_chat_chunk_response,
-            )
+
+        # Determine response type
+        stream = body.get("stream", False)
+        method = self.handle_stream if stream else self.generate_response
+        method(
+            prompt,
+            chat_id,
+            requested_model,
+            stop_id_sequences,
+            eos_token_id,
+            max_tokens,
+            temperature,
+            top_p,
+            repetition_penalty,
+            repetition_context_size,
+            create_chat_response,
+        )
 
     def handle_completions(self, body: dict):
         completion_id = f"cmpl-{uuid.uuid4()}"
@@ -406,40 +394,28 @@ class APIHandler(BaseHTTPRequestHandler):
         ]
         eos_token_id = _tokenizer.eos_token_id
         max_tokens = body.get("max_tokens", 100)
-        stream = body.get("stream", False)
         requested_model = body.get("model", "default_model")
         temperature = body.get("temperature", 1.0)
         top_p = body.get("top_p", 1.0)
         repetition_penalty = body.get("repetition_penalty", 1.0)
         repetition_context_size = body.get("repetition_context_size", 20)
-        if not stream:
-            return self.generate_response(
-                prompt,
-                completion_id,
-                requested_model,
-                stop_id_sequences,
-                eos_token_id,
-                max_tokens,
-                temperature,
-                top_p,
-                repetition_penalty,
-                repetition_context_size,
-                create_completion_response,
-            )
-        else:
-            self.handle_stream(
-                prompt,
-                completion_id,
-                requested_model,
-                stop_id_sequences,
-                eos_token_id,
-                max_tokens,
-                temperature,
-                top_p,
-                repetition_penalty,
-                repetition_context_size,
-                create_completion_chunk_response,
-            )
+
+        # Determine response type
+        stream = body.get("stream", False)
+        method = self.handle_stream if stream else self.generate_response
+        method(
+            prompt,
+            completion_id,
+            requested_model,
+            stop_id_sequences,
+            eos_token_id,
+            max_tokens,
+            temperature,
+            top_p,
+            repetition_penalty,
+            repetition_context_size,
+            create_completion_response,
+        )
 
 
 def run(host: str, port: int, server_class=HTTPServer, handler_class=APIHandler):
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -234,7 +234,7 @@ class APIHandler(BaseHTTPRequestHandler):
         top_p: float,
         repetition_penalty: Optional[float],
         repetition_context_size: Optional[int],
-        response_creator: Callable[[str, str, mx.array, List[int], str], dict],
+        response_creator: Callable[[str, str, mx.array, list[int], str]],
     ):
         tokens = []
         for (token, _), _ in zip(
@@ -271,7 +271,7 @@ class APIHandler(BaseHTTPRequestHandler):
         top_p: float,
         repetition_penalty: Optional[float],
         repetition_context_size: Optional[int],
-        response_creator: Callable[[str, str, str], dict],
+        response_creator: Callable[[str, str, str]],
     ):
         self.send_response(200)
         self.send_header("Content-type", "text/event-stream")
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -5,9 +5,8 @@ import json
 import time
 import uuid
 import warnings
-from collections import namedtuple
 from http.server import BaseHTTPRequestHandler, HTTPServer
-from typing import Callable, List, Optional
+from typing import Callable, List, NamedTuple, Optional
 
 import mlx.core as mx
 import mlx.nn as nn
@@ -26,7 +25,9 @@ def load_model(model_path: str, adapter_file: Optional[str] = None):
     _model, _tokenizer = load(model_path, adapter_file=adapter_file)
 
 
-StopCondition = namedtuple("StopCondition", ["stop_met", "trim_length"])
+class StopCondition(NamedTuple):
+    stop_met: bool
+    trim_length: int
 
 
 def stopping_criteria(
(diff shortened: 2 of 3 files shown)


ACKNOWLEDGMENTS.md
```diff
@@ -12,4 +12,3 @@ MLX Examples was developed with contributions from the following individuals:
 - Shunta Saito: Added support for PLaMo models.
 - Gabrijel Boduljak: Implemented `CLIP`.
 - Markus Enzweiler: Added the `cvae` examples.
-- Rasmus Kinnunen: Fixed a security hole in the `llms/mlx_lm` example
```
Y4hL, line 15: Turns out I was only using about half of my braincells yesterday and got BaseHTTPRequestHandler and SimpleHTTPRequestHandler mixed up, BaseHTTPRequestHandler doesn’t actually serve the local directory by default… awkward 🫠 Will be working overtime to make up for it

llms/mlx_lm/server.py
```diff
@@ -234,7 +248,12 @@ def generate_response(
                 break
 
         text = _tokenizer.decode(tokens)
-        return response_creator(response_id, requested_model, prompt, tokens, text)
+        response = response_creator(response_id, requested_model, prompt, tokens, text)
+        try:
+            self.wfile.write(f"{response}\n\n".encode())
```
Y4hL, line 253: @mzbac Responses the the handle_stream method send the json as a "data:" header. Is that how OpenAI does it? If so we'll need to add it here too. The OpenAI API reference doesn't mention if json is part of the body or data header, at least I couldn't find anything

mzbac: It's necessary. I would suggest that when you work on the server, it's better to set up integration with some clients so you can test out your changes. For example, set up chat-ui with mlx-lm.server https://github.com/mzbac/mlx-chat-ui

Y4hL: So far I've been working to try to match the OpenAI API Reference, but I guess their documentation does have its limits. I do find it strange though that json data would be contained in a http header.

mzbac: What do you mean by "in header"? The stream response is being sent via SSE. https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events

Y4hL: Ah that link helps, ty

Y4hL:
Should be ready for review.

Changes summed up:

- Parse more of the parameters shared between the completion types before calling completion type specific methods.

- Store more of the request data inside of the object instead of passing it around functions

- More reliably imitate the OpenAI API. The system fingerprint and created timestamp would change between every single token completion, while OpenAI's API keeps it static for that request.

- Added documentation for all important methods.

llms/mlx_lm/server.py
```diff
@@ -80,244 +81,263 @@ def convert_chat(messages: any, role_mapping: Optional[dict] = None):
...
-
-
 class APIHandler(BaseHTTPRequestHandler):
 
-    def __init__(self, *args, **kwargs):
-        # Prevent exposing local directory by deleting HEAD and GET methods
-        delattr(self, "do_HEAD")
-        delattr(self, "do_GET")
+    # System fingerprint is the same for all instances of the class
+    system_fingerprint = f"fp_{uuid.uuid4()}"
```
mzbac, line 87: Is the BaseHTTPRequestHandler instantiated for each request?

llms/mlx_lm/server.py
```diff
@@ -80,244 +81,263 @@ def convert_chat(messages: any, role_mapping: Optional[dict] = None):
...
 
+    def __init__(self, *args, **kwargs):
+        """
+        Create static request specific metadata
+        """
+        self.created = int(time.time())
         super().__init__(*args, **kwargs)
 
-    def _set_headers(self, status_code=200):
+    def _set_completion_headers(self, status_code: int = 200):
```
mzbac, line 96: Since you are using it for both completion and 404 responses, maybe it would be better to just name it as _set_headers?

llms/mlx_lm/server.py
```diff
@@ -80,244 +81,263 @@ def convert_chat(messages: any, role_mapping: Optional[dict] = None):
...
             if stop_condition.stop_met:
                 if stop_condition.trim_length:
                     tokens = tokens[: -stop_condition.trim_length]
                 break
 
         text = _tokenizer.decode(tokens)
-        return response_creator(response_id, requested_model, prompt, tokens, text)
+        response = self.generate_response(text, "stop", len(prompt), len(tokens))
+
+        self.wfile.write(f"{json.dumps(response)}\n\n".encode())
```
mzbac, line 251: SSE events without the `data:` prefix will not work with most openapi integrated clients.

mzbac:
Looks like a nice refactoring. Just curious, have you done any testing on your local with a client integration?

llms/mlx_lm/utils.py
```diff
@@ -355,14 +355,14 @@ def load_model(model_path: Path, lazy: bool = False) -> nn.Module:
...
-    adapter_file: str = None,
+    adapter_file: Optional[str] = None,
     lazy: bool = False,
 ) -> Tuple[nn.Module, PreTrainedTokenizer]:
     """
     Load the model and tokenizer from a given path or a huggingface repository.
 
     Args:
-        model_path (Path): The path or the huggingface repository to load the model from.
+        path_or_hf_repo (Path): The path or the huggingface repository to load the model from.
```
awni, line 365: 👍

llms/mlx_lm/utils.py
```diff
@@ -355,14 +355,14 @@ def load_model(model_path: Path, lazy: bool = False) -> nn.Module:
 def load(
     path_or_hf_repo: str,
     tokenizer_config={},
-    adapter_file: str = None,
+    adapter_file: Optional[str] = None,
```
awni, line 358: 👍

llms/mlx_lm/utils.py
```diff
@@ -203,7 +203,7 @@ def generate(
     temp: float = 0.0,
     max_tokens: int = 100,
     verbose: bool = False,
-    formatter: Callable = None,
+    formatter: Optional[Callable] = None,
```
awni, line 206: 👍

awni:
Thank you for the updates it does indeed look quite nice. 

Good question from @mzbac, were you able to run any local testing?

We still sadly have 0 tests for this package. Something that I am planning to remedy soon.

awni:
Also, looks like the formatting failed. Could you run the pre-commit hooks?

llms/mlx_lm/server.py
Y4hL: This should keep the fingerprint the same for the entirety of the program for all requests, however worst case scenario it only stays the same for that specific request, which is still better than changing it for each token as it was before

llms/mlx_lm/server.py
Y4hL: I named it like this because it's clear that there are different headers for streams and completions. Maybe we just add a generic _add_header for cases where something goes wrong

llms/mlx_lm/server.py
Y4hL: I left the packets as they were in the previous implementation. This one should be for the non-stream responses and I thought they probably don't use SSE since it's just a single packet, so I didn't change anything.

mzbac: For the non-stream response, we should simply return the response as JSON. In that case, should we remove the additional line breaks?
```
self.wfile.write(json.dumps(response).encode("utf-8")
```

llms/mlx_lm/server.py
mzbac: ok, in that case maybe just keep the imp as it is, don't  have to overcomplicate it.

llms/mlx_lm/server.py
Y4hL: I don't know how the clients are implemented, but if it's a normal http request, we need to also include a Content-Length header if we leave out newlines. Do you have an example where this kind of thing is implemented? Either client or server

mzbac: > I don't know how the clients are implemented, but if it's a normal http request, we need to also include a Content-Length header if we leave out newlines. Do you have an example where this kind of thing is implemented? Either client or server

Maybe I am missing something here, but regardless of what kind of client we are using, we should not add additional unnecessary line breaks in the response.

Y4hL: The client needs a way to know when to stop reading data from the socket, so we either need to add a content-length header, or see the implementation for a client we want compatility with

mzbac: It shouldn't be the case. That's handled by the HTTP protocol, not the application layer. I quickly tried it on my local machine without trailing line breaks and it worked fine.

Y4hL: The newlines should be removed yes, however we do need to also send a content-length header, not sure why it works without either.

[HTTP Response specification by Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Messages#body_2)
[Stack overflow answer on including newlines in a HTTP body](https://stackoverflow.com/a/30677511/22499229)

mzbac: I have tested with `curl` and `chrome`, both seem to work fine without the content-length. I assume that most modern browsers/tools should be able to handle this properly. Maybe just remove the additional line break, and if the issue occurs later, we can address it then.

Y4hL: I just sent in a patch that implements the Content-Length header, since it wasn't too big of a change

awni:
@Y4hL is the change summary you mention above still accurate? I'm hoping to dive into this PR soon, and just want to get oriented

Y4hL:
@awni Yes it is, although the changes have become more than I anticipated, they're still towards the same goal

Y4hL:
I'm currently looking into fixing #524, but I can't get to it right away. I'll create a separate PR for it after this gets merged, else the changes will just keep piling on.

llms/mlx_lm/server.py
```diff
@@ -80,344 +75,311 @@ def convert_chat(messages: any, role_mapping: Optional[dict] = None):
...
-                except Exception as e:
-                    print(e)
-                    break
+
+            # Continue generating tokens until buffer is as large as the longest stop_id_sequence
+            if len(stop_sequence_buffer) < max_stop_id_sequence_len:
+                continue
+
+            # Continue until unicode character is fully generated
+            if "\ufffd" in TOKENIZER.decode(token):
```
mzbac, line 307: Would using the variable as `REPLACEMENT_CHAR` make it more readable than `"\ufffd"`?

llms/mlx_lm/server.py
```diff
@@ -5,43 +5,38 @@
...
+class StopCondition(NamedTuple):
+    stop_met: bool
+    trim_length: int
 
 
 def stopping_criteria(
     tokens: List[int],
-    stop_id_sequences: List[np.ndarray],
-    eos_token_id: int,
+    stop_id_sequences: List[List[int]],
```
mzbac, line 29: Isn't the stop_id_sequences a `List[np.ndarray]`? what's the reason you want to change it to `List[List[int]]` ?

llms/mlx_lm/server.py
```diff
@@ -80,344 +75,311 @@ def convert_chat(messages: any, role_mapping: Optional[dict] = None):
...
+        prompt_token_count: Optional[int] = None,
+        completion_token_count: Optional[int] = None,
+    ) -> dict:
+        """
+        Generate a single response packet based on response type (stream or not),
+            completion type and parameters
+
+        Args:
+            text (str): Text generated by model
+            finish_reason (Union[str, None]):
```
mzbac, line 164:
```suggestion
            finish_reason (Union[Literal["length", "stop"], None]):
```

llms/mlx_lm/server.py
Y4hL: The character only gets used in one place, and in my opinion the comment above it makes it clear enough what it does

llms/mlx_lm/server.py
Y4hL: We used to call tokenizer decode and encode with 
```python
return_tensors="np"
```
However I did some testing, and not converting it to np made it 20% faster. The np.array_equal also checks List[int] against a np.ndarray, which is just as fast as checking 2 python lists for equality.

The performance difference is almost negligible, but we get to drop the numpy import entirely with this change, so we do gain something out of it.

llms/mlx_lm/server.py
mzbac: Not really, readers need to understand that it's a REPLACEMENT_CHAR and then realize that some Unicode characters require more than one character to be decoded correctly. Without this information, it will be very difficult to understand why we need to skip that decoding.

llms/mlx_lm/server.py
Y4hL: Yeah, good catch, thank you

llms/mlx_lm/server.py
Y4hL: I generated random length random character combinations and called tokenizer.encode once with `return_tensors="np"` and once without
```python
import random
import string
import timeit

import numpy as np
import mlx.core as mx

from utils import load

MODEL, TOKENIZER = load("mlx-community/quantized-gemma-2b")

# Generate data to encode
words = [''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(random.randint(1, 10))) for _ in range(100)]

comparison = [141, 1321, 231, 31]

def test_tokenizer():
    tokens = TOKENIZER.encode(random.choice(words), add_special_tokens=False)
    if tokens == comparison:
        return



def test_np_tokenizer():
    tokens = TOKENIZER.encode(random.choice(words), return_tensors="np", add_special_tokens=False)
    if np.array_equal(tokens, comparison):
        pass


print("With np:    ", timeit.timeit(test_np_tokenizer, number=1000000))
print("Without np: ", timeit.timeit(test_tokenizer, number=1000000))
```
My results were:
```
With np:     12.122312665997015
Without np:  9.860362332998193
```

mzbac:
I have pulled down the PR, did a quick test with chat-ui integration, and it seems to have worked as expected. However, I didn't test the stop sequence given we have a bug for the current implementation anyway.

llms/mlx_lm/server.py
Y4hL: How that I think about it, shouldn't we then also remove it from `tokens` and the sequence buffer?

mzbac: We cannot remove it. The REPLACEMENT_CHAR was caused by partial decoding (only the first part of Unicode was generated at the time, and we will need the second token to be able to decode). so, if the next token comes, the tokenizer should be able to properly decode with two tokens.

Y4hL: Ah, so it isn't technically needed for decoding of unicode, but the tokenizer is implemented in a way where it looks out for them?

mzbac: Yeah, I'm not an expert on tokenizers, but my understanding is that it has something to do with how tokenizers decode Unicode.

Y4hL: Is this enough?
```python
# "\ufffd" is used to indicate to the tokenizer, that subsequent characters
# should be combined into a single unicode character
if "\ufffd" in TOKENIZER.decode(token):
    continue
```

awni (approved):
Looks great!! Thanks @Y4hL for the contributions and @mzbac for the review!

awni:
Looks like the formatter failed. Could you run the pre-commit hooks please?

Issue #500, opened 2024-02-28
Starcoder2 Support


lazarust:
It would be nice if we could support `starcoder2`. 

Expected Behavior:

`python -m mlx_lm.generate --model bigcode/starcoder2-3b --prompt "Write a quick sort in C++"`

awni:
Happy to accept a contribution for that!

lazarust:
@awni I'll take a stab at it and see what I can do! Could you point me to a good starting point?

awni:
Of course! 

So it's relatively simple, you add the model file to this directory: https://github.com/ml-explore/mlx-examples/tree/main/llms/mlx_lm/models 

You can see other examples in there, I recommend starting from one of the models that is pretty similar to starcoder2

Make sure the name of the model file is the same as the `model_type` in the config (`starcoder2`) (https://huggingface.co/bigcode/starcoder2-7b/blob/main/config.json#L17)

If you want to add LoRA support you can edit this [section of the code](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/tuner/utils.py#L27-L60) as needed

awni:
>  Just a heads-up the model is a FIM type

What's FIM? And does the HF tokenizer template handle that or do we need to do something special?

lazarust:
@awni Could you also assign me to this issue? I can't seem to do it

Vaibhavs10:
> What's FIM? And does the HF tokenizer template handle that or do we need to do something special?

FIM - Fill in the Middle. Nothing special. Just that it is not chat-tuned.

https://huggingface.co/bigcode/starcoder2-15b#running-the-model-on-cpugpumulti-gpu

sugatoray:
FIM: Fill in the Middle [arxiv-paper](https://arxiv.org/abs/2207.14255)

PR #502, opened 2024-02-28
Add Starcoder 2

Muhtasham:
Adding new [code models](https://x.com/BigCodeProject/status/1762842312005026258?s=20) that dropped

merged to [Transformers](
https://github.com/huggingface/transformers/pull/29228/files#diff-2a85c68b1238c345068cca421d21974fe8e478d94eb3931cd18fa364391122e6)


awni:
This looks nice. Is it working? 

There is an open issue btw #500, not sure if @lazarust made any progress? Maybe this is a good PR to collaborate on, though it looks like it is basically done?

lazarust:
@awni I haven't had time to make progress yet, but this solution looks like it's the right direction

Muhtasham:
> Is it working? 

@awni not yet hf is down 😞 but when trying to generate locally despite files being there and import working getting

`ValueError: Model type starcoder2 not supported. Error: No module named 'mlx_lm.models.starcoder2'` 

but when I do in python shell I do not get that error but hf down error, is this behaviour intended

@lazarust happy to collaborate, this is still draft based on original mistral implementation I think we need some more changes from https://github.com/huggingface/transformers/pull/29120

awni:
> but when I do in python shell I do not get that error but hf down error, is this behaviour intended

No it's not. Make sure you do not have another MLX LM installed (`pip uninstall mlx-lm`). Also make sure you have an editable install `pip install -e .`

Muhtasham:
qq @awni  is gelu_pytorch_tanh equivalnet in mlx is 

```python
    def __call__(self, x) -> mx.array:
        return self.w2(nn.gelu(self.w1(x)) * self.w3(x))
```

angeloskath:
@Muhtasham the tanh approximation in PyTorch is not strictly the same as `nn.gelu` but it is close to several significant digits so it should not cause any issues. See the plot at https://github.com/ml-explore/mlx/pull/744#issuecomment-1965616223 for the difference between PyTorch's Tanh approximation and `nn.gelu` .

llms/README.md
```diff
@@ -46,7 +46,7 @@ You can convert models in the Python API with:
 ```python
 from mlx_lm import convert
 
-upload_repo = "mistralai/Mistral-7B-Instruct-v0.1"
+upload_repo = "mlx-community/My-Mistral-7B-v0.1-4bit"
```
sugatoray, line 49: Thank you for adding the model. 
- Why is this switching to `mlx-community/My-Mistral-7B-v0.1-4bit`? 
- Is this necessary for StarCoder2 model implementation?

Muhtasham: Hi this is just typo from elsewhere

sugatoray: Ok. In that case, please switch it back to what it was earlier.

awni: @sugatoray I think what @Muhtasham meant is that the previous version is a typo (which it is). It doesn't make sense to point to the mistralai repo as an example upload name. We should have something more suggestive of a custom repo. I would use the same thing we use later in the `README.md`:

```
mlx-community/my-4bit-mistral
```

As for including it in the PR, I think it is fine (or you can send another if you prefer).

awni:
How is this going? Were you able to generate any sensible code yet?

Muhtasham:
Having some,

`raise ValueError(f"Received parameters not in model: {extras}.")`

 I went through model config nothing seems extra  any hints @awni ?

command

```bash
python -m mlx_lm.convert \
    --hf-path bigcode/starcoder2-3b \
    -q \
    --upload-repo mlx-community/starcoder2-3b-4bit
```

llms/mlx_lm/models/starcoder2.py
```diff
@@ -0,0 +1,183 @@
...
+
+        return self.output(self.norm(h)), cache
+
+
+class Model(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+        self.model_type = args.model_type
+        self.model = Starcoder2Model(args)
+
```
mzbac, line 171: I think the lm_head is missing here.

llms/mlx_lm/models/starcoder2.py
```diff
@@ -0,0 +1,183 @@
...
+            scores += mask
+        scores = mx.softmax(scores.astype(mx.float32), axis=-1).astype(scores.dtype)
+        output = (scores @ values).transpose(0, 2, 1, 3).reshape(B, L, -1)
+        return self.wo(output), (keys, values)
+
+
+class MLP(nn.Module):
+    def __init__(self, dim, hidden_dim):
+        super().__init__()
+        self.w1 = nn.Linear(dim, hidden_dim, bias=False)
```
mzbac, line 99: The name of the MLP's linear layer does not match. FYI :https://github.com/huggingface/transformers/blob/main/src/transformers/models/starcoder2/modeling_starcoder2.py#L147

mzbac:
I suggest either referring to the transformers implementation if you are familiar with the codebase, or loading the model weights and checking the weight names which will give you a hint of the model structure.

https://github.com/huggingface/transformers/blob/main/src/transformers/models/starcoder2/modeling_starcoder2.py#L1070

llms/mlx_lm/models/starcoder2.py
```diff
@@ -0,0 +1,178 @@
...
+        dim = args.hidden_size
+        self.n_heads = n_heads = args.num_attention_heads
+        self.n_kv_heads = n_kv_heads = args.num_key_value_heads
+
+        self.repeats = self.n_heads // self.n_kv_heads
+
+        head_dim = args.hidden_size // args.num_attention_heads
+        self.scale = head_dim**-0.5
+
+        self.wq = nn.Linear(dim, n_heads * head_dim, bias=False)
```
mzbac, line 54: Starcoder2 uses the standard HF attention layer name, so you may need to update the q,k,v,o layer name here as well.
https://github.com/huggingface/transformers/blob/main/src/transformers/models/starcoder2/modeling_starcoder2.py#L211

llms/mlx_lm/models/starcoder2.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class Starcoder2Model(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+        self.args = args
+        self.vocab_size = args.vocab_size
+        self.num_hidden_layers = args.num_hidden_layers
+        assert self.vocab_size > 0
+        self.tok_embeddings = nn.Embedding(args.vocab_size, args.hidden_size)
```
mzbac, line 138: The embedding layer name needs to be updated -> https://github.com/huggingface/transformers/blob/main/src/transformers/models/starcoder2/modeling_starcoder2.py#L903

llms/mlx_lm/models/starcoder2.py
Muhtasham: thanks but Mistral also uses them https://github.com/huggingface/transformers/blob/f1b1379f37c6b9626bb1c795d89be4c0a606f957/src/transformers/models/mistral/modeling_mistral.py#L227

and in MLX implentation of Mistral have same name
https://github.com/ml-explore/mlx-examples/blob/3acc1ec84e838d1070b04ab77b76e31176955a8d/llms/mistral/mistral.py#L55C13-L55C14

or are you referring to sth else

armbues:
A couple more things from what I can see:
- Replace RMSNorm with LayerNorm
- Use transposed token embeddings (like in Gemma) as there are no output weights

llms/mlx_lm/models/starcoder2.py
```diff
@@ -0,0 +1,182 @@
...
+    def __call__(
+        self,
+        x: mx.array,
+        mask: Optional[mx.array] = None,
+        cache: Optional[Tuple[mx.array, mx.array]] = None,
+    ) -> mx.array:
+        r, cache = self.self_attn(self.input_layernorm(x), mask, cache)
+        h = x + r
+        r = self.mlp(self.post_attention_layernorm(h))
+        out = h + r
```
lazarust, line 128: I'm a little confused by what this is doing... could someone explain like I'm 5?

Muhtasham:
thanks for all the help!
 
Resolved all the layer errors but now getting 

`ValueError: The checkpoint you are trying to load has model type `starcoder2` but Transformers does not recognize this architecture. This could be because of an issue with the checkpoint, or because your version of Transformers is out of date.`

I have trasfomers: 4.38.2 @awni  any leads here

llms/mlx_lm/models/starcoder2.py
awni: There are some incredible resources to learn about how transformers work. Much better than what I could do in a GitHub comment. Check out e.g. https://jalammar.github.io/illustrated-transformer/

armbues:
Quick addition: apparently the 15B starcoder2 has output weights, so there needs to be a check when initializing the model.

llms/mlx_lm/models/starcoder2.py
lazarust: Thanks @awni!!

awni:
@armbues @Muhtasham [use this flag in the config](https://huggingface.co/bigcode/starcoder2-15b/blob/main/config.json#L24) to determine if embeddings should be shared with outputs.

llms/mlx_lm/models/starcoder2.py
Muhtasham: Thank you very much for all the help

llms/mlx_lm/requirements.txt
```diff
@@ -1,5 +1,5 @@
 mlx>=0.4
 numpy
-transformers>=4.38.0
+git+https://github.com/huggingface/transformers.git
```
awni, line 3: This is a pre-release? If starcoder requires this, let's wait until it hits pypi and put the version instead.

Muhtasham: Correct, sure

awni:
I tried the 3B model, it looks almost right but maybe missing something as it goes off the rails at the end.

Also it ran without crashing with my install of Transformers library (4.38.0). I reverted the requirement.. maybe you had an old version?

```
 python -m mlx_lm.generate --model bigcode/starcoder2-3b --prompt "Write a quick sort in C++" --temp 0.0 --colorize --max-tokens 100
Fetching 6 files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████| 6/6 [00:00<00:00, 98304.00it/s]
==========

#include <iostream>
using namespace std;
void swap(int &a, int &b)
{
int temp = a;
a = b;

}
int main()
{

int a = 1;

int b = 2;
int c = 3;

int main()

int main()

int

int
int
int



int

int




==========
Prompt: 18.702 tokens-per-sec
Generation: 33.758 tokens-per-sec
```

Muhtasham:
@awni  I uploaded wights with that version of transformers is that problem? 

`mlx-community/starcoder2-3b-4bit`

awni:
> @awni I uploaded wights with pre-release version of transformers is that problem?

Thanks for uploading the weihgts.

It's not a problem at all. The transformer version shouldn't effect the model you uploaded to Hugging Face.

awni:
Are you getting similar output as me for the 3B model? It looks like it should be a bit better..maybe the prompt is off?

Muhtasham:
> Are you getting similar output as me for the 3B model? It looks like it should be a bit better..maybe the prompt is off?

Yes got same, since it is base model for code completion and fill in the middle objective seems expected

But tried this 

```python
python -m mlx_lm.generate --model mlx-community/starcoder2-3b-4bit --prompt "def print_hello_world():" --temp 0.0 --colorize --max-tokens 100

Fetching 7 files: 100%|████████████████████████████████████████████████████| 7/7 [00:00<00:00, 56461.78it/s]
==========
Prompt: def print_hello_world():

    print("hello world")
def print_hello_hello_world():
    print("hello world")
def print_hello_world():
    print("hello world")
def print_hello_world():
    print("hello world")
def print_hello_world():
    print("hello world")

def print_hello_world():
    print("hello world")
def print_hello_world():
    print("hello world")

def print_hello_
==========
Prompt: 2.273 tokens-per-sec
Generation: 22.712 tokens-per-sec
```

also qq @awni should we do anything extra for ` "hidden_act": "gelu_pytorch_tanh"`

awni:
> fill in the middle objective

What do you mean by that exactly?

> should we do anything extra for  "hidden_act": "gelu_pytorch_tanh"

Where did that come up?

Muhtasham:
> What do you mean by that exactly?

Model is trained to predict middle span given the surrounding text, effectively learning to infill text See here: https://arxiv.org/abs/2207.14255

>> should we do anything extra for  "hidden_act": "gelu_pytorch_tanh" 
> Where did that come up?

Saw [that](https://huggingface.co/bigcode/starcoder2-3b/blob/cc9b0d492c17a1c8467bf462a071abad279bcb04/config.json#L8) in config but no act in MLP layers of the model

awni:
> Model is trained to predict middle span given the surrounding text, effectively learning to infill text

Got it, thanks! So how is one supposed to use it for generation? Are there instruct / fine-tuned variants?

Muhtasham:
> Got it, thanks! So how is one supposed to use it for generation? Are there instruct / fine-tuned variants?

It will follow, only found this public fine-tune of it https://huggingface.co/winstxnhdw/finetune_starcoder2

I was planning to fine-tune via mlx, we can also merge this as is if you want I will double check implementation with authors from HF, this will make it easier for them to check

mzbac:
> > Model is trained to predict middle span given the surrounding text, effectively learning to infill text
> 
> Got it, thanks! So how is one supposed to use it for generation? Are there instruct / fine-tuned variants?

By looking at the tokenizer config, it specifies token to allow the model to fill in the missing code in the middle of the code snippet, something like this `<fim_prefix>${textAboveCursor}<fim_suffix>${textBelowCursor}<fim_middle>`.

Muhtasham:
Ok [found](https://twitter.com/xiangyue96/status/17636) two more fine-tuned models

https://huggingface.co/m-a-p/OpenCodeInterpreter-SC2-3B
https://huggingface.co/m-a-p/OpenCodeInterpreter-SC2-7B

Muhtasham:
@awni for fine-tunes is it better to have separate file?

as they have lm_head which base model doesn't have like here 

https://huggingface.co/m-a-p/OpenCodeInterpreter-SC2-7B/blob/cf493c6e9ab2b6bbf0e095622a54a8277d0d294f/model.safetensors.index.json#L6

and here 

https://huggingface.co/bigcode/starcoder2-15b/blob/d020785eeb4c7e9493938fdafc11c10af30ca020/model.safetensors.index.json#L6

awni:
> @awni for fine-tunes is it better to have separate file?

They should be the same file because they have the same `model_type`. However it is a bit unfortunate that they don't have the config relating to the tied embeddings.  We might have to patch that model to set it dynamically if you want to use those fine-tunes (alternatively you could send them a PR to set the config flag properly).

awni:
> By looking at the tokenizer config, it specifies token to allow the model to fill in the missing code in the middle of the code snippet, something like this prompt = <fim_prefix>${textAboveCursor}<fim_suffix>${textBelowCursor}<fim_middle>.

I'm not sure i follow how you generate just the middle portion.. I guess I should look at the paper for the details.

llms/mlx_lm/models/stablelm.py
```diff
@@ -128,7 +128,6 @@ class DecoderLayer(nn.Module):
     def __init__(self, config: ModelArgs):
         super().__init__()
         self.self_attn = Attention(config=config)
-        self.input_layernorm = LayerNorm(config.hidden_size, eps=config.layer_norm_eps)
```
Muhtasham, line 131: @awni this is duplicate right?

awni: Yea, good catch! Would be great if you remove!

mzbac:
> > By looking at the tokenizer config, it specifies token to allow the model to fill in the missing code in the middle of the code snippet, something like this prompt = <fim_prefix>${textAboveCursor}<fim_suffix>${textBelowCursor}<fim_middle>.
> 
> I'm not sure i follow how you generate just the middle portion.. I guess I should look at the paper for the details.

Yeah, basically it's just a different way to structure the prompt and allow the model to autocomplete for middle portion. A more practical example is similar to using GitHub Copilot, where the model will fill in the content at the cursor position.

awni:
Ok, I'm a little confused as to what to do with this. The model doesn't generate sensible outputs, so is the plan to land it so people can use LoRA / fine-tunes? Or do we need to fix the prompt? Or is there some other thing that I'm missing?

mzbac:
> Ok, I'm a little confused as to what to do with this. The model doesn't generate sensible outputs, so is the plan to land it so people can use LoRA / fine-tunes? Or do we need to fix the prompt? Or is there some other thing that I'm missing?

I will do some testing tonight. In terms of fine-tuning, my understanding is that the FIM model is just different in training data prompt format and not different from normal fine-tuning. However, I have not done any FIM fine-tuning so I may be wrong.

mzbac:
I did some quick tests and it looks like there are issues in the mlx implementation. I couldn't figure out exactly where, but when you use the transformers example as shown below, you can see that the model generates correct FIM code. However, this doesn't work in the mlx implementation.
```
# pip install git+https://github.com/huggingface/transformers.git # TODO: merge PR to main
from transformers import AutoModelForCausalLM, AutoTokenizer

checkpoint = "bigcode/starcoder2-3b"
device = "cuda" # for GPU usage or "cpu" for CPU usage

tokenizer = AutoTokenizer.from_pretrained(checkpoint)

model = AutoModelForCausalLM.from_pretrained(checkpoint).to(device)

tokenizer.eos_token = "<file_sep>"
inputs = tokenizer.encode("<fim_prefix>\ndef quicksort(arr):\n<fim_suffix>\n<fim_middle>\n", return_tensors="pt").to(device)
outputs = model.generate(input_ids=inputs,
    temperature=0, 
    max_new_tokens=100,
)
print(tokenizer.decode(outputs[0]))
```
output:
```
<fim_prefix>
def quicksort(arr):
<fim_suffix>
<fim_middle>
        if len(arr) <= 1:
                return arr
        else:
                pivot = arr[0]
                less = [i for i in arr[1:] if i <= pivot]
                greater = [i for i in arr[1:] if i > pivot]
                return quicksort(less) + [pivot] + quicksort(greater)

print(quicksort([10, 5, 2, 3]))<file_sep>
```

awni:
> However, this doesn't work in the mlx implementation

You tried it with the correct prompt right? Do we add the FIM prompt by default or is that something you have to do manually at the moment?

mzbac:
> > However, this doesn't work in the mlx implementation
> 
> You tried it with the correct prompt right? Do we add the FIM prompt by default or is that something you have to do manually at the moment?

I used `python -m mlx_lm.generate --model bigcode/starcoder2-3b --prompt "<fim_prefix>\ndef quicksort(arr):\n<fim_suffix>\n<fim_middle>\n" --temp 0.0  --eos-token "<file_sep>" --ignore-chat-template`

mzbac:
Based on @Blaizzy's PR (https://github.com/ml-explore/mlx-examples/pull/518), the instruction prompt failed because it was using traditional Rope. Changing to traditional = False should fix it. However, I have tried his implementation and it still doesn't work with FIM prompt. I suspect there may be something wrong with the Rope implementation or configuration.

awni:
Ok just to make sure I understand - the only difference between this PR and #518 is that the RoPE traditional is False?

awni:
If that's the case, let's simply update that here? It's a one line change. I think it makes more sense than starting on a separate PR?

awni:
Can confirm, seems to work now! Thanks!

awni (approved):
Thanks for the addition and contributions everyone!

Blaizzy:
> Based on @Blaizzy's PR (#518),
> Can confirm, seems to work now! Thanks!
> Thanks for the addition and contributions everyone!

Most welcome 😊!

mzbac:
> Ok just to make sure I understand - the only difference between this PR and #518 is that the RoPE traditional is False?

@awni, there are some minor model args that need to be cleaned up. For example, the use of `rms_norm_eps` should be changed to `norm_epsilon` to match Starcoder2's configuration. Also, it would be more efficient to use the repeat header instead of mapping concatenate. I'm not sure if @Blaizzy would like to make a following PR for it; otherwise, I don't mind creating a patch PR to clean it up.

Blaizzy:
> Ok just to make sure I understand - the only difference between this PR and #518 is that the RoPE traditional is False?

Didn't notice the last update on this PR early Saturday. 

I'm happy it works now 🚀!

Blaizzy:
@mzbac I can do that :)

Blaizzy:
> Also, it would be more efficient to use the repeat header instead of mapping concatenate.

Could you elaborate on this?

mzbac:
> > Also, it would be more efficient to use the repeat header instead of mapping concatenate.
> 
> Could you elaborate on this?

Yeah, in the current implementation here https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/starcoder2.py#L71-L76, it would be updated to something like : https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/llama.py#L83-L85.

Blaizzy:
I see, thanks for clarifying! Will do.

I thought so too because I used the llama's repeat.

If I may ask, why was it done differently?

mzbac:
Yeah, based on the previous PR, simply repeating is faster than concatenating. Just FYI: https://github.com/ml-explore/mlx-examples/pull/443

pai-sameen:
This feature came at the perfect time for me! 24 hours ago there was no support and now there is. Love open source! 🩶

pai-sameen:
So I'm trying to fine-tune StarCoder2 using QLoRA. 

````shell
$ pwd
>>> /.../mlx-examples/llms
````

When I issue the fine-tuning command, I get an `AttributeError`:
````shell
$ python -m mlx_lm.lora \                                           
    --model mlx-community/starcoder2-3b-4bit \
    --train \
    --data $(realpath ../lora/data) \
    --iters 10


>>> None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
Loading pretrained model
Fetching 7 files: 100%|███████████████████████████████████████████████████████████████████████| 7/7 [00:00<00:00, 61422.86it/s]
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/sameenislam/source/sameenislam/mlx-examples/llms/mlx_lm/lora.py", line 246, in <module>
    run(args)
  File "/Users/sameenislam/source/sameenislam/mlx-examples/llms/mlx_lm/lora.py", line 172, in run
    linear_to_lora_layers(model, args.lora_layers)
  File "/Users/sameenislam/source/sameenislam/mlx-examples/llms/mlx_lm/tuner/utils.py", line 27, in linear_to_lora_layers
    if model.model_type in [
       ^^^^^^^^^^^^^^^^
  File "/Users/sameenislam/anaconda3/lib/python3.11/site-packages/mlx/nn/layers/base.py", line 137, in __getattr__
    super(Module, self).__getattr__(key, val)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'super' object has no attribute '__getattr__'. Did you mean: '__setattr__'?
````

Has anyone encountered this?

mzbac:
> So I'm trying to fine-tune StarCoder2 using QLoRA.
> 
> ```shell
> $ pwd
> >>> /.../mlx-examples/llms
> ```
> 
> When I issue the fine-tuning command, I get an `AttributeError`:
> 
> ```shell
> $ python -m mlx_lm.lora \                                           
>     --model mlx-community/starcoder2-3b-4bit \
>     --train \
>     --data $(realpath ../lora/data) \
>     --iters 10
> 
> 
> >>> None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
> Loading pretrained model
> Fetching 7 files: 100%|███████████████████████████████████████████████████████████████████████| 7/7 [00:00<00:00, 61422.86it/s]
> Traceback (most recent call last):
>   File "<frozen runpy>", line 198, in _run_module_as_main
>   File "<frozen runpy>", line 88, in _run_code
>   File "/Users/sameenislam/source/sameenislam/mlx-examples/llms/mlx_lm/lora.py", line 246, in <module>
>     run(args)
>   File "/Users/sameenislam/source/sameenislam/mlx-examples/llms/mlx_lm/lora.py", line 172, in run
>     linear_to_lora_layers(model, args.lora_layers)
>   File "/Users/sameenislam/source/sameenislam/mlx-examples/llms/mlx_lm/tuner/utils.py", line 27, in linear_to_lora_layers
>     if model.model_type in [
>        ^^^^^^^^^^^^^^^^
>   File "/Users/sameenislam/anaconda3/lib/python3.11/site-packages/mlx/nn/layers/base.py", line 137, in __getattr__
>     super(Module, self).__getattr__(key, val)
>     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'super' object has no attribute '__getattr__'. Did you mean: '__setattr__'?
> ```
> 
> Has anyone encountered this?
> 
> Also including this to show that inference is working:
> 
> ```shell
> $ python -m mlx_lm.generate --model mlx-community/starcoder2-3b-4bit --prompt "hello"
> 
> >>> None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
> Fetching 7 files: 100%|██████████████████████████████████████████████████████████████████████| 7/7 [00:00<00:00, 145347.17it/s]
> ==========
> Prompt: hello
> _1(void) {
> 	printf("hello, world\n");
> }
> 
> void helloworld_print_double_1(double a) {
> 	printf("%f\n", a);
> }
> 
> double helloworld_square_1(double a) {
> 	return a * a;
> }
> 
> double helloworld_square_2(double a) {
> 	return a * a;
> }
> 
> double helloworld_square_
> ==========
> Prompt: 16.348 tokens-per-sec
> Generation: 40.031 tokens-per-sec
> ```

There is a missing model_type in the starcoder2. You can try adding it to your local code as shown in this [PR](https://github.com/ml-explore/mlx-examples/pull/522).

pai-sameen:
> There is a missing model_type in the starcoder2. You can try adding it to your local code as shown in this [PR](https://github.com/ml-explore/mlx-examples/pull/522).

Awesome, I've made the change locally from PR #522 and it's working like a charm!
````shell
$ python -m mlx_lm.lora \ 
    --model $(realpath mlx_model) \
    --train \
    --data $(realpath ../lora/data) \
    --iters 10

>>> None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
Loading pretrained model
Total parameters 602.983M
Trainable parameters 1.212M
Loading datasets
Training
Starting training..., iters: 10
Iter 1: Val loss 2.363, Val took 30.839s
Iter 10: Train loss 2.274, Learning Rate 1.000e-05, It/sec 0.490, Tokens/sec 196.036, Trained Tokens 3999
Saved final adapter weights to adapters.npz.
````

P.S. `$(realpath mlx_model)` is the model obtained with:
````shell
$ python -m mlx_lm.convert \                                        
    --hf-path bigcode/starcoder2-3b \
    -q
````

PR #505, opened 2024-02-29
LoRA: adapter file Support path information

madroidmaq:
Enable the passed in args.adapter_file to contain path information, such as passing in "build/adapters.npz". The storage path of the final file is as follows:

```
.
├── build
│   ├── adapters.npz
│   └── checkpoints
│       ├── 100_adapters.npz
│       └── 200_adapters.npz
```

### DIFF
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -146,7 +146,9 @@ def train(
     def check_checkpoints_path(adapter_file):
         checkpoints_path = "checkpoints"
         if os.path.dirname(adapter_file):
-            checkpoints_path = os.path.join(os.path.dirname(adapter_file), "checkpoints")
+            checkpoints_path = os.path.join(
+                os.path.dirname(adapter_file), "checkpoints"
+            )
 
         if not os.path.exists(checkpoints_path):
             os.makedirs(checkpoints_path)
@@ -250,7 +252,9 @@ def train(
 
         # Save adapter weights if needed
         if (it + 1) % args.steps_per_save == 0:
-            checkpoint_adapter_file = f"{adapter_path}/{it + 1}_{os.path.basename(args.adapter_file)}"
+            checkpoint_adapter_file = (
+                f"{adapter_path}/{it + 1}_{os.path.basename(args.adapter_file)}"
+            )
             save_adapter(model=model, adapter_file=checkpoint_adapter_file)
             print(
                 f"Iter {it + 1}: Saved adapter weights to {os.path.join(checkpoint_adapter_file)}."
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/tuner/trainer.py
```diff
@@ -143,9 +143,20 @@ def train(
...
+        checkpoints_path = "checkpoints"
+        if os.path.dirname(adapter_file):
+            checkpoints_path = os.path.join(
+                os.path.dirname(adapter_file), "checkpoints"
+            )
+
+        if not os.path.exists(checkpoints_path):
+            os.makedirs(checkpoints_path)
+
+        return checkpoints_path
```
awni, line 156: Do you mind switching this to `pathlib.Path` instead of `os`? We use that everywhere else for the most part and I think it would be good to be consistent.

madroidmaq: > Do you mind switching this to `pathlib.Path` instead of `os`? We use that everywhere else for the most part and I think it would be good to be consistent.

Of course, I prefer to use `pathlib.Path`, using `os` follows the previous call more.

llms/mlx_lm/tuner/trainer.py
```diff
@@ -143,17 +143,15 @@ def train(
...
-            )
+    def check_checkpoints_path(adapter_file) -> str:
+        checkpoints_path = Path("checkpoints")
+        if Path(adapter_file).parent:
+            checkpoints_path = Path(adapter_file).parent / "checkpoints"
 
-        if not os.path.exists(checkpoints_path):
-            os.makedirs(checkpoints_path)
+        if not checkpoints_path.exists():
+            checkpoints_path.mkdir(parents=True)
```
awni, line 152:
```suggestion
        checkpoints_path.mkdir(parents=True, exist_ok=True)
```

llms/mlx_lm/tuner/trainer.py
```diff
@@ -143,17 +143,15 @@ def train(
...
 
-        return checkpoints_path
+        return str(checkpoints_path)
 
     # Create checkpoints directory if it does not exist
     adapter_path = check_checkpoints_path(args.adapter_file)
```
awni, line 157: It might simplify things a bit if you 
1. Make `args.adapter_file` a `Path` here instead.
2. Make a second variable for the `checkpoint_path` (it's a bit more clear as well IMO)

awni (changes requested):
Looks like the format failed, could you run the pre-commit hooks?

After that, we can merge it!

madroidmaq:
> Looks like the format failed, could you run the pre-commit hooks?
> 
> After that, we can merge it!

It‘s DONE.

awni (approved):

PR #509, opened 2024-02-29
Remove gc

awni:
Requires https://github.com/ml-explore/mlx/pull/759


awni:
I timed converting Mixtral on the M1 mini seems to work just as well without gc now.

angeloskath (approved):
👌

PR #514, opened 2024-03-01
Update to StableLM code

ashishdatta:
As part of Transformers `4.38.0` https://github.com/huggingface/transformers/pull/28810

StableLM is now part of Transformers without the need for `trust_remote_code`. It is now `stablelm_epoch` -> `stablelm`

Also update config to match the latest.


awni:
Thanks! Would you mind making the change here as well: https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/tuner/utils.py#L32

That should cover everything.

ashishdatta:
> Thanks! Would you mind making the change here as well: https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/tuner/utils.py#L32
> 
> That should cover everything.

done in eb40948

awni (approved):
🚀

PR #516, opened 2024-03-02
Stable diffusion XL

angeloskath:
Fixes the issues of #160 and adds stable diffusion XL turbo.

Specifically, this adds ancestral Euler sampler, adds extra information in the embedding necessary for SD-XL. Additionally this PR fixes fp16 generation and loads the weights using MLX since MLX now supports safetensor files.

I am adding the mandatory generated images below:
image_generations

On the M2 Ultra it takes quite a bit less than a second to fully generate an image. It takes ~5s to fully generate 16 images. In a following PR I will add compiling which will at the very least save a bit of memory.


stable_diffusion/stable_diffusion/model_io.py
```diff
@@ -1,7 +1,8 @@
-# Copyright © 2023 Apple Inc.
+# Copyright © 2023-2024 Apple Inc.
 
 import json
 from functools import partial
+from typing import Optional
```
awni, line 5: nit: I believe the `safetensor_open` import below is now unused

stable_diffusion/stable_diffusion/model_io.py
```diff
@@ -80,8 +97,9 @@ def map_unet_weights(key, value):
 
     if len(value.shape) == 4:
         value = value.transpose(0, 2, 3, 1)
+        value = value.reshape(-1).reshape(value.shape)
```
awni, line 100: Interesting.. the intention is to make the value row contiguous. Just curious did you notice it was faster?

Not suggesting a change here.. but I wonder if there is a cleaner / more general solution to this. Probably doesn't come up often except when using convolutions..

Muhtasham:
@angeloskath have you also tried quantized versions to fit in m2 base? If so can you please upload to hf

stable_diffusion/stable_diffusion/model_io.py
angeloskath: I didn't check but it must be as it will keep being copied into a temporary just before the matmul (when conv is done with a matmul) or conv. I hate this but I don't think there is any way around knowing things about the underlying implementation.

awni:
Updated the readme to be a little more up-to-date (feel free to revise as you see fit). @angeloskath maybe we should update the benchmarks in there as well?

stable_diffusion/image2image.py
```diff
@@ -26,6 +26,7 @@
     args = parser.parse_args()
 
     sd = StableDiffusion()
+    sd.ensure_models_are_loaded()
```
awni, line 29: Just wondering, why did you add that? Is it to get better timing information?

Also just a thought, but maybe it would be simpler to have the models subclass `nn.Module` and then you can just do a regular `mx.eval` on the params / state.

stable_diffusion/stable_diffusion/unet.py
```diff
@@ -106,10 +106,11 @@ def __init__(
     def __call__(self, x, encoder_x, attn_mask, encoder_attn_mask):
         # Save the input to add to the output
         input_x = x
+        dtype = x.dtype
 
         # Perform the input norm and projection
         B, H, W, C = x.shape
-        x = self.norm(x).reshape(B, -1, C)
+        x = self.norm(x.astype(mx.float32)).astype(dtype).reshape(B, -1, C)
```
awni, line 113: Note for LN in LLMS we downcast before the affine which works fine and avoids needing to cast the weights at each iteration.  May be a little optimization to take the default compiled layernorm from MLX LM and use it her.

awni (approved):
Looks amazing. Left a few minor comments... but otherwise g2g.

Regarding the benchmarks in the README.. they are a bit out of date. It might be good to either update (or simply remove since there is no easy way to reproduce them).

angeloskath:
Thanks for the additions @awni .

Added some small on the fly quantization support (unfortunately 4 bits doesn't really work for the unet). I tested it on the 8GB macmini and it can generate 4 images at a time with no-swapping during generation and some small swapping during decoding/upsampling. Total time is 21s for generation and another ~15 for reading the model in memory and quantizing ie <10s/image from pressing the enter key.

@Muhtasham I added quantization as an option as that is simpler for now (simply pass `-q` to `txt2image.py`). I also changed it to float16 by default which makes quite a bit of difference for small machines.

I will remove the benchmarks as they don't make much sense indeed and I will also update the layer norm to use the one in `nn`.

stable_diffusion/stable_diffusion/unet.py
angeloskath: Out of curiosity, is there a reason we are not moving the shapeless compile inside `mlx.nn` ?

awni: The only reason is that it doesn't do the up casting for the reduction.

awni: I was thinking about adding it there ... if you like that more feel free / I can add it, I would be on board.

stable_diffusion/image2image.py
angeloskath: We need to fix some things in `nn.Module` I think cause deleting layers is not that straightforward anymore. Probably somewhere along the changes for state etc something made this not possible. So I 'll leave it for later.

angeloskath:
So I added some more memory oriented optimizations. There is now 0-swapping when generating on the 8GB macmini with a peak memory usage of 7GB all spent on the decoder. There may be the need to tile the decoder to reduce that but I 'll leave that for another PR.

stable_diffusion/image2image.py
awni: Sounds good to leave for later... but I didn't fully get your comment. How does `eval`ing the model params relate to deleting the layers?

PR #518, opened 2024-03-02
PC: Add support for Starcoder2

Blaizzy:
Hey @awni and @Muhtasham,

I was super excited about integrating starcoder 2 into mlx. However, while working on the mlx-langchain integration, I realized that @Muhtasham beat me to the punch and submitted a PR. Kudos to you!

I took a look at the original draft PR a few days ago and noticed a few areas that could use some improvement (mlp and others). When I tested the code with the quantized models uploaded to the hub, it didn't quite work as expected. But not to worry, I've fixed it! ✨ ✅

This implementation is working well and matches the original model using transformers in full precision and quantized configurations.
```shell
python -m mlx_lm.generate --model mlx-community/starcoder2-3b-4bit --prompt "Write a quick sort in C++"  --colorize --max-tokens 200
```

Output:
```shell

```cpp
//https://www.hackerrank.com/challenges/quicksort3/problem

#include <bits/stdc++.h>

using namespace std;

vector<int> quickSort(vector<int> ar) {
    vector<int> output;
    int pivot = ar[0];
    for(int i = 1; i < ar.size(); i++){
        if(ar[i] < pivot)
            output.push_back(ar[i]);
    }
    output.push_back(pivot);
    for(int i = 1; i < ar.size(); i++){
        if(ar[i] > pivot)
            output.push_back(ar[i]);
    }
    return output;
}

int main(void) {
    vector <int>  ar;
    int num_of_elements;
    cin >> num
```


mzbac:
Nice work! Just curious, does this work with the FIM prompt?

llms/mlx_lm/models/starcoder2.py
```diff
@@ -0,0 +1,195 @@
...
+            keys = mx.concatenate([key_cache, keys], axis=2)
+            values = mx.concatenate([value_cache, values], axis=2)
+        else:
+            queries = self.rope(queries)
+            keys = self.rope(keys)
+
+        scores = (queries * self.scale) @ keys.transpose(0, 1, 3, 2)
+        if mask is not None:
+            scores += mask
+        scores = mx.softmax(scores, axis=-1).astype(values.dtype)
```
mzbac, line 93: Maybe we need to keep `scores.astype(mx.float32)` in case softmax overflows on float16 values.

awni:
@Blaizzy thanks for adding! I just merged #502 since it was basically done. I'm going to close this in favor of that. Appreciate the insights here which helped with #502!

Blaizzy:
> Nice work! Just curious, does this work with the FIM prompt?

Thanks! It should work without a problem. 

If not I can investigate.

Blaizzy:
> @Blaizzy thanks for adding! I just merged #502 since it was basically done. I'm going to close this in favor of that. Appreciate the insights here which helped with #502!

Mose welcome! 

It's my pleasure.

PR #519, opened 2024-03-03
chore(mlx-lm): fix broken server.py script

mzbac:
Address: https://github.com/ml-explore/mlx-examples/pull/498#issuecomment-1974960089, just in case there is a release for mlx-lm. We won't release the broken script.


awni (approved):
Thanks @mzbac I was planning to do a new dist soon!

PR #530, opened 2024-03-04
Update README.md

amcox886:
The default behaviour of where the convert.py saved files was wrong. It also was inconsistent with how the later script test.py is trying to use them (and assuming naming convention). 

I don't actually see a quick way to automate this since--as written--the  target directory is set directly by an argument. It would probably be best to rewrite it so that the argument is used as an override variable, but the default behaviour is to construct a file path based on set and unset arugments. This also is complex because "defaults" are assumed in the naming convention as well.

### DIFF
--- a/whisper/README.md
+++ b/whisper/README.md
@@ -39,8 +39,20 @@ To generate a 4-bit quantized model, use `-q`. For a full list of options:
 python convert.py --help
 ```
 
-By default, the conversion script will make the directory `mlx_models/tiny`
-and save the converted `weights.npz` and `config.json` there.
+By default, the conversion script will make the directory `mlx_models`
+and save the converted `weights.npz` and `config.json` there. Note that
+this does not automatically save different versions in this folder.
+However, the downloaded models are cached for future use. 
+
+Consider a scripted run (e.g. BASH, zsh) like the below example to stream-line filemaking:
+
+```
+model="tiny.en"
+python convert.py --torch-name-or-path ${model} --dtype float16 --mlx-path mlx_models/${model}_fp16
+python convert.py --torch-name-or-path ${model} --dtype float32 --mlx-path mlx_models/${model}_fp32
+python convert.py --torch-name-or-path ${model} -q --q_bits 4 --mlx-path mlx_models/${model}_quantized_4bits
+```
+Another solution would be to implement a loop that executes those commands for each of a list of Models. 
 
 ### Run
 


whisper/README.md
```diff
@@ -39,8 +39,20 @@ To generate a 4-bit quantized model, use `-q`. For a full list of options:
...
+
+Consider a scripted run (e.g. BASH, zsh) like the below example to stream-line filemaking:
+
+```
+model="tiny.en"
+python convert.py --torch-name-or-path ${model} --dtype float16 --mlx-path mlx_models/${model}_fp16
+python convert.py --torch-name-or-path ${model} --dtype float32 --mlx-path mlx_models/${model}_fp32
+python convert.py --torch-name-or-path ${model} -q --q_bits 4 --mlx-path mlx_models/${model}_quantized_4bits
+```
+Another solution would be to implement a loop that executes those commands for each of a list of Models.
```
awni, line 55: Delete this line, it is not needed imo.

whisper/README.md
```diff
@@ -39,8 +39,20 @@ To generate a 4-bit quantized model, use `-q`. For a full list of options:
 python convert.py --help
 ```
 
-By default, the conversion script will make the directory `mlx_models/tiny`
-and save the converted `weights.npz` and `config.json` there.
+By default, the conversion script will make the directory `mlx_models`
+and save the converted `weights.npz` and `config.json` there. Note that
+this does not automatically save different versions in this folder.
+However, the downloaded models are cached for future use.
```
awni, line 45:
```suggestion
and save the converted `weights.npz` and `config.json` there. 
```

awni (commented):
Thanks! Tis is a good addition. I left a few comments on wording to simplify / make it a bit more clear.

whisper/README.md
amcox886: Which line are you referring to? Sorry, on my end there's nothing marked out. I'm actually going to do another version shortly and actually implement a script like I suggested at the end. Coming shortly.

amcox886 (commented):
Just one last question there. Thanks.

whisper/README.md
amcox886: Yeah, I just pushed another update that just implemented the actual thing that I was thinking of. You know how it gets--just can't get it out of your head so you just do _the thing_ to forget about it.

amcox886: I'll mark this comment as complete since it's no longer relevant.

amcox886 (commented):
yup

whisper/README.md
```diff
@@ -39,8 +39,51 @@ To generate a 4-bit quantized model, use `-q`. For a full list of options:
...
+However, the downloaded models are cached for future use. 
+
+Consider a scripted run (e.g. BASH, zsh) like the below example to stream-line filemaking:
+
+ Here's a Python script that acts as a wrapper for calling `convert.py` with different parameters:
+
+```import sys
+import subprocess
+
+def run_convert(config):
```
awni, line 54: Do you mind removing this script? I don't think it makes sense to include. Different people will have different needs with respect to how they want serialize models.

awni (approved):

PR #531, opened 2024-03-05
feat: add update_config functionality

### LINKED ISSUE #532: fix issues with `config.json`
The config file uploaded to huggingface repos by `mlx-lm` is not sorted currently. Need updates to sort the config files first and also update the `_name_or_path` key in `config.json` with the HF `upload_repo` provided by the user.

### Example of what is expected:

- [mlabonne/NeuralMonarch-7B](https://huggingface.co/mlabonne/NeuralMonarch-7B/blob/main/config.json)

```json
{
    "_name_or_path": "mlabonne/Monarch-7B",
	...
}
```

### What MLX repos have currently vs. what the fix should implement:

- [sugatoray/mlx-alphamonarch-7b-q4bits](https://huggingface.co/sugatoray/mlx-alphamonarch-7b-q4bits/commit/9256ba904e65f0be691c488ced03039b4caad069)

```diff
{
-   "_name_or_path": "/Users/sugatoray/.cache/huggingface/hub/models--mlabonne--AlphaMonarch-7B/snapshots/83a587946ece206325bf07c3bf11134a7256feb0",
+    "_name_or_path": "sugatoray/mlx-alphamonarch-7b-q4bits",
}
```

> :bulb: Related issue: 
> - #526

sugatoray:
Adds a standardized config updating functionality

- [x] sorts the config for better readability
- [x] updates `_name_or_path` key in config with upload_repo
- [x] sets indentation of 4 spaces
- [x] ~~allows adding other key-value pairs via kwargs~~ [removed based on code-review]
- [x] reduces code duplication
- [x] standardizes saving config across mlx-lm

Also updated `CONTRIBUTING.md`:
- [x] added instruction on formatting specific set of files with `pre-commit`.

The config file uploaded to huggingface repos is not sorted currently. This sorts the config files first and also updates the `_name_or_path` key in `config.json` with the HF `upload_repo` provided by the user.

### Example of what is expected:

- [mlabonne/NeuralMonarch-7B](https://huggingface.co/mlabonne/NeuralMonarch-7B/blob/main/config.json)

```json
{
    "_name_or_path": "mlabonne/Monarch-7B",
	...
}
```

### What MLX repos have currently vs. what this will fix:

- [sugatoray/mlx-alphamonarch-7b-q4bits](https://huggingface.co/sugatoray/mlx-alphamonarch-7b-q4bits/commit/9256ba904e65f0be691c488ced03039b4caad069)

```diff
{
-   "_name_or_path": "/Users/sugatoray/.cache/huggingface/hub/models--mlabonne--AlphaMonarch-7B/snapshots/83a587946ece206325bf07c3bf11134a7256feb0",
+    "_name_or_path": "sugatoray/mlx-alphamonarch-7b-q4bits",
}
```

> :bulb: Closes #532


sugatoray:
cc: @awni

llms/mlx_lm/utils.py
```diff
@@ -550,6 +550,47 @@ def quantize_model(
...
 
 
+def update_config(
+    config: dict,
+    upload_repo: Optional[str] = None,
+    config_path: Optional[Union[str, Path]] = None,
+    indent: int = 4,
+    **kwargs,
+) -> dict:
+    """Update the model configuration and save config to the ``config_path`` (if provided).
```
mzbac, line 560: Should we make the config_path mandatory and always save the configuration? We could even name it "save_config". I don't see another use case where we would only update the configuration without saving it.

sugatoray: It's a good point. However, there could be scenarios, when you could have a use-case of updating the config dynamically on the fly, and may want to skip I/O interactions until the end -- when you just write to the disc just once.

I thought of leaving that option out there so that the same function can potentially fill in both scenarios --  dynamic in-memory config update vs. immediate file (config.json) update. But I am open to suggestions on this one.

About calling the variable `save_config`, I have a little reservation though, as `save_config` sounds like a verb and not a noun as `config_path`. Also, you could be reading config from an existing `config.json` file and then updating it later (such as sorting, adding more key-value pairs, etc.). Thus, the `config_path` refers to a path where either a new config.json file will be created or an existing config will be updated.

sugatoray:
@awni / @mzbac Please see if it is okay to setup the CI run now.

llms/mlx_lm/utils.py
mzbac: It would be clear to split this function into two: one for updating the config and another for saving the config. Then you can compose the functions for different use cases. what do you think?

awni (changes requested):
Thanks for fixing that. There's two changes I would make:

- Rename `update_config` to `save_config` as I think that is more clear (and just save it without returning it).
- Don't actually update the config in `update_config`. While I think what you did with the kwarg update could be nice in some cases, I don't know of any future use cases where will commonly want to send in kwargs and do an inplace update.. we can revisit in the future.

llms/mlx_lm/utils.py
sugatoray: Yes, maintaining single responsibility would be preferable. I will refactor it into two functions then:

- `update_config()`
- `save_config()`

What do you think about it?

sugatoray:
> Thanks for fixing that. There's two changes I would make:
> 
> * Rename `update_config` to `save_config` as I think that is more clear (and just save it without returning it).
> * Don't actually update the config in `update_config`. While I think what you did with the kwarg update could be nice in some cases, I don't know of any future use cases where will commonly want to send in kwargs and do an inplace update.. we can revisit in the future.

Thank you, @awni. I will make changes accordingly.

awni:
Hi @sugatoray sorry I just saw your question:

> Are you at all in favor of breaking the function into two?

The update config seems less useful for me since we only do an update to the one field and it happens in only a couple of places. It doesn't buy us much in terms of saving code or reducing duplication, but introduces a new function / API that obscures things a little.

sugatoray:
> Hi @sugatoray sorry I just saw your question:
> 
> > Are you at all in favor of breaking the function into two?
> 
> The update config seems less useful for me since we only do an update to the one field and it happens in only a couple of places. It doesn't buy us much in terms of saving code or reducing duplication, but introduces a new function / API that obscures things a little.

Thank you @awni. I will change it to save_config. 

I kept the kwargs in the function to allow for the user to add user-defined-metadata in the json file. But am dropping that functionality for now.

awni:
@sugatoray I'm still waiting for you to let me know when to check this again right?

sugatoray:
@awni I have made code changes suggested in the code-review. Please take a look at it and merge it if all looks good to you.

llms/mlx_lm/fuse.py
```diff
@@ -98,6 +105,8 @@ def main() -> None:
             raise ValueError(
                 "Must provide original Hugging Face repo to upload local model."
             )
+        # update the config with the upload_repo as the value of "_name_or_path" key
```
awni, line 108: 👍

awni (approved):
Thank you!

PR #534, opened 2024-03-05
Print huggingface repo url

madroidmaq:
Print the URL address of the uploaded Hugging Face repository, which can be opened by clicking on the link as a quick jump in the terminal, as follows:

image

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -470,6 +470,7 @@ def upload_to_hub(path: str, upload_repo: str, hf_path: str):
         repo_id=upload_repo,
         repo_type="model",
     )
+    print(f"Upload successful, go to https://huggingface.co/{upload_repo} for details.")
 
 
 def save_weights(


llms/mlx_lm/utils.py
```diff
@@ -470,6 +470,7 @@ def upload_to_hub(path: str, upload_repo: str, hf_path: str):
         repo_id=upload_repo,
         repo_type="model",
     )
+    print(f"Upload successful, go to https://huggingface.co/{upload_repo} for details.")
```
sugatoray, line 473: Have you tested what happens if the upload fails? If the huggingface upload does a graceful exit on upload-failure, then the print statement could still end up executing.

Perhaps consider adding a `try -- except` block and control the behavior with more certainty. For example:

```py
from huggingface_hub import HfApi, ModelCard, logging, whoami

## ...

try:
	api.upload_folder(
        folder_path=path,
        repo_id=upload_repo,
        repo_type="model",
    )
    print(f"Upload successful, go to https://huggingface.co/{upload_repo} for details.")
except Exception as e:
	print(f"Upload failed:\n{"-"*40}\n{e}")
finally:
	print(f"Huggingface User: {whoami().get('name')}")
```

awni:
This is nice! Thanks! I like @sugatoray suggestion, do you mind testing something like that out to make sure it works and then we can review / merge?

llms/mlx_lm/utils.py
awni: Yea on second thought, it might be better the way you have it now. If api raises an exception, we will see it and I don't think we want the program to exit cleanly in that case 🤔

madroidmaq: If you want to check whether the user contained in `upload_repo` matches the currently logged-in account, I think it's more reasonable to adopt a `fail-fast approach.

sugatoray: > Great suggestion, I indeed hadn't considered the exceptional cases before. However, I'm not clear on the purpose of printing `whoami()` information because I know who I am. Or did I miss some information?

The `whoami` part is to capture in logging who was running it. You may or may not include it. The `finally` part will be run no matter what. So, keeping a definitive output there could help. Hence, I added whoami there as an example.

madroidmaq: I can print the corresponding information only when ensuring the upload is successful, and display the original error message in case of an exception, so that users can see the complete error information clearly.

awni: I think it's ok for now. If the HF api call failed without raising an exception then we would need a way to detect it, otherwise the exception will propagate and the print statement will be skipped. Detecting a non-exception failure sounds a bit more involved. If you want to look at it as a future addition, feel free assuming, it doesn't add too much complexity.

awni (approved):

PR #541, opened 2024-03-06
Predict stop sequence matches during streaming

### LINKED ISSUE #524: Using token sequences as stop criteria does not work in mlx_lm
The implementation of stop_criteria in mlx_lm.server is inherently flawed. Stop sequences only get matched when the newest tokens generated perfectly match a stop sequence. However it does not stop if the stop sequence is inside of the tokens in any other way.

This only checks if the newest tokens perfectly match a sequence
```python
for stop_ids in stop_id_sequences:
    if len(tokens) >= len(stop_ids):
        if np.array_equal(tokens[-len(stop_ids) :], stop_ids):
            return StopCondition(stop_met=True, trim_length=len(stop_ids))
```

stopping_criteria only gets called when max_stop_id_sequence_len amount of tokens have been generated, which is the length of the longest stop sequence.
```python
if len(stop_sequence_buffer) > max_stop_id_sequence_len:
      if REPLACEMENT_CHAR in _tokenizer.decode(token):
          continue
      stop_condition = stopping_criteria(
          tokens,
          stop_id_sequences,
          eos_token_id,
      )
```

Example where it breaks:

I have two stop sequences, one is of length 4, and one of length 6. Once 6 tokens have been generated, stop_criteria is called. However the tokens I have generated only match the stop sequence of length 4, and the match happens at the start of the new tokens, not at the end. However since stop_criteria only checks the end of the full token list, it does not get matched and generation does not stop.

```python
stop sequence = [1, 2, 3, 4]

new tokens = [1, 2, 3, 4, 5, 6]

# stop_criteria get called and checks for len(stop_sequence) of tokens at the end of new tokens

[3, 4, 5, 6] != [1, 2, 3, 4]
```

tidely:
When streaming using mlx_lm/server.py we should predict potential stop sequence matches, and generate tokens until we know that there is no match. This prevents the server from sending parts of a stop sequence to the client before it finds the match. 

Fixes #524 

My implementation adds a new function called "sequence_overlap" which checks how much sequence 1 has overlap with sequence 2. It checks for larger overlaps first, and returns the overlap as an integer.

The server checks for overlaps, and generates more tokens before allowing the server to send them.
```python
if any((sequence_overlap(tokens, sequence) for sequence in stop_id_sequences)):
    continue
```
The sequence_overlap implementation can be tested with this example:
```python
from typing import Sequence


def sequence_overlap(s1: Sequence, s2: Sequence) -> int:
    """
    Check how much overlap two sequences have.
        Only checks the end of s1 overlapping the start of s2

    Args:
        s1 (Sequence): The first sequence, which end is checked
        s2 (Sequence): The second sequence, which beginning is checked

    Returns:
        int: The amount of overlap between s1 and s2
    """
    # Count down from the length of the smaller list -> Checks for larger overlaps first
    for index in range(min(len(s1), len(s2)), 0, -1):
        # Check if they have index amount of overlap
        if s1[-index:] == s2[:index]:
            return index
    return 0


stop_sequence = [27, 28, 29]

tokens = []
new_tokens = []

for token in range(50):
    tokens.append(token)
    new_tokens.append(token)

    # This should always be the first check, since it needs to be performed on every token
    if len(tokens) >= len(stop_sequence) and tokens[-len(stop_sequence):] == stop_sequence:
        print("Contains stop sequence:", new_tokens)
        tokens = tokens[:len(tokens) - len(stop_sequence)]
        new_tokens.clear()
        break

    # Generate tokens until we know that tokens does not contain stop sequence
    if sequence_overlap(tokens, stop_sequence):
        print("Found a possible start to a stop sequence:", new_tokens)
        continue

    # Process new tokens
    print("Processing new tokens:", new_tokens)
    new_tokens.clear()

# In the case that the generation ends with the start of a stop sequence
# We need to process leftovers, since it would call continue until a break
if new_tokens:
    print("Processing leftover new tokens", new_tokens)
    new_tokens.clear()

print("Full sequence:", tokens)

```

### DIFF
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -66,7 +66,7 @@ def sequence_overlap(s1: Sequence, s2: Sequence) -> bool:
         bool: If the two sequences have overlap
     """
     max_overlap = min(len(s1), len(s2))
-    return any(s1[-i:] == s2[:i] for i in range(1, max_overlap))
+    return any(s1[-i:] == s2[:i] for i in range(1, max_overlap + 1))
 
 
 def convert_chat(messages: List[dict], role_mapping: Optional[dict] = None):
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -535,6 +535,7 @@ class APIHandler(BaseHTTPRequestHandler):
             self.wfile.flush()
 
         # check is there any remaining text to send
+        detokenizer.finalize()
         last_segment = detokenizer.last_segment
         if last_segment:
             if stop_sequence_suffix is not None:
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -67,7 +67,7 @@ def sequence_overlap(s1: Sequence, s2: Sequence) -> int:
         int: The amount of overlap between s1 and s2
     """
     # Count down from the length of the smaller list -> Checks for larger overlaps first
-    for index in range(min(len(s1), len(s1)), 0, -1):
+    for index in range(min(len(s1), len(s2)), 0, -1):
         # Check if they have index amount of overlap
         if s1[-index:] == s2[:index]:
             return index
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -477,12 +477,13 @@ class APIHandler(BaseHTTPRequestHandler):
         stop_id_sequences: List[List[int]],
     ):
         """
-        Generate response to prompt and foward it to the client using a Server Sent Events (SSE) stream.
+        Generate response to prompt and foward it to the client using a Server
+        Sent Events (SSE) stream.
 
         Args:
             prompt (mx.array): The prompt, in token form inside of a mlx array
-            stop_id_sequences (List[List[int]]):
-                A list of stop words passed to the stopping_criteria function
+            stop_id_sequences (List[List[int]]): A list of stop words passed to
+              the stopping_criteria function
         """
         # No additional headers are needed, call end_headers
         self.end_headers()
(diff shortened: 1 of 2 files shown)


awni:
Oh much better, thank you 😄

llms/mlx_lm/server.py
```diff
@@ -329,26 +334,36 @@ def handle_stream(
...
+        # Check for remaining text
+        generated_text = TOKENIZER.decode(tokens)
+        if len(generated_text) > current_text_index:
+            new_text = generated_text[current_text_index:]
+            response = self.generate_response(new_text, "length")
             self.wfile.write(f"data: {json.dumps(response)}\n\n".encode())
             self.wfile.flush()
 
-        self.wfile.write("data: [DONE]\n\n".encode())
+        self.wfile.write(f"data: [DONE]\n\n".encode())
```
awni, line 366: why make that a format string?

awni:
What about adding your test (modified for unittest) as a test case to a new test file `test_server.py` in the tests directory: https://github.com/ml-explore/mlx-examples/tree/main/llms/tests ?

llms/mlx_lm/server.py
```diff
@@ -55,6 +55,26 @@ def stopping_criteria(
...
+        Only checks the end of s1 overlapping the start of s2
+
+    Args:
+        s1 (Sequence): The first sequence, which end is checked
+        s2 (Sequence): The second sequence, which beginning is checked
+
+    Returns:
+        int: The amount of overlap between s1 and s2
+    """
+    # Count down from the length of the smaller list -> Checks for larger overlaps first
```
awni, line 70: I'm wondering about the decision to start with larger overlaps? It seems like it could return faster if you start with smaller overlaps, but maybe I misunderstand how it's used?

llms/mlx_lm/server.py
Y4hL: Ah, I think I pasted it from the old server.py before my last PR got merged that removed it. I'll remove it

llms/mlx_lm/server.py
Y4hL: I had a version where I generated `len(stop_sequence) - overlap` tokens on a match, which required finding the longest match. But it overly complicated things for little benefit. So you have a good point, we can change it to check shorter matches first, and return a boolean instead.

Y4hL:
Yeah, I'll have a look into writing a unittest

awni:
Hi! Is this ready to be reviewed again?

Y4hL:
@awni yes

awni (approved):
Finally got around to getting this done 😅 .. thanks for the contribution 🙏

Issue #543, opened 2024-03-07
FreedomIntelligence/Apollo-0.5B failed to run


madroidmaq:
I'd like to try the inference effect of [FreedomIntelligence/Apollo-0.5B](https://huggingface.co/FreedomIntelligence/Apollo-0.5B/tree/main)'s model, which by his `model_type` is ` qwen2`, which theoretically has current model type support. However, when running it, it encounters the following problem, and I'm not sure what's causing it.

`FreedomIntelligence/Apollo-0.5B` The result of the execution is as follows:

```
python -m mlx_lm.generate --model FreedomIntelligence/Apollo-0.5B --max-tokens 50 --temp 0 --colorize --prompt "hello"
Fetching 8 files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████| 8/8 [00:00<00:00, 112977.89it/s]
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/opt/homebrew/lib/python3.11/site-packages/mlx_lm/generate.py", line 139, in <module>
    main(args)
  File "/opt/homebrew/lib/python3.11/site-packages/mlx_lm/generate.py", line 107, in main
    model, tokenizer = load(
                       ^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/mlx_lm/utils.py", line 382, in load
    model = load_model(model_path, lazy)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/mlx_lm/utils.py", line 346, in load_model
    model.load_weights(list(weights.items()))
  File "/opt/homebrew/lib/python3.11/site-packages/mlx/nn/layers/base.py", line 206, in load_weights
    raise ValueError(f"Missing parameters: {missing}.")
ValueError: Missing parameters: lm_head.weight.
```

PS: The `Qwen/Qwen1.5-0.5B-Cha`t model can run normally.

mzbac:
It looks like the `tie_word_embeddings` has been enabled.

PR #544, opened 2024-03-07
chore(mlx-lm): enable tie_word_embeddings config for qwen2

mzbac:
fix https://github.com/ml-explore/mlx-examples/issues/543


awni (approved):
Thanks!

PR #545, opened 2024-03-07
Enable unit testing in Circle and start some MLX LM tests

awni:
- I think it's time to start testing MLX LM since it's gotten a bit big and manual testing is tedious.
- Added the tests to run in Circle as well.


angeloskath (approved):
Nice, testing are the best PRs!

PR #546, opened 2024-03-07
LoRA on all linear transformer block layers

chimezie:
In general, it would be useful for the LoRa tuner and linear_to_lora_layers in mlx_lm.tuner.utils in particular to allow more control over how LoRa is applied (the ability to specify the rank and alpha, for example).  This PR is for the basic ability to indicate via CLI option (-l/--lora-all-linear) that LoRa is applied to **all** linear transformer block layers (adding the application of LoRa to k_proj and o_proj).  Currently, it only does this for the following models:

- llama
- mistral
- mixtral
- gemma
- phi
- plamo
- qwen2
- stablelm
- starcoder2

This is mainly motivated by the finding of authors of the QLoRa [paper](https://arxiv.org/abs/2305.14314)

> that the most critical LoRA hyperparameter is how many LoRA adapters are used in total and that **__LoRA on all linear transformer block layers are required__** to match full finetuning performance.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,5 +61,7 @@ workflows:
           type: approval
       - apple/authenticate:
           context: pr-approval
+      - mlx_lm_build_and_test:
+          requires: [ hold ]
       - linux_build_and_test:
           requires: [ hold ]
--- a/llms/mlx_lm/examples/lora_config.yaml
+++ b/llms/mlx_lm/examples/lora_config.yaml
@@ -48,3 +48,10 @@ test_batches: 500
 
 # Maximum sequence length.
 max_seq_length: 2048
+
+#LoRA parameters (the values below are the defaults but can be changed)
+#These can only be specified in a YAML config file
+lora_parameters:
+  lora_layer_keys: ["q_proj", "v_proj"] #Which attention layer keys to apply LoRA to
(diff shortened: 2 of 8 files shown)


awni:
I think a more sustainable way to do this is the following:
- Have a field in the Yaml which gives the layers keys to apply LoRA to
- Make this a yaml config parameter only. So you can't do it with the CLI
- Change the LoRA util to read from that

chimezie:
> I think a more sustainable way to do this is the following:
> 
> * Have a field in the Yaml which gives the layers keys to apply LoRA to
> * Make this a yaml config parameter only. So you can't do it with the CLI
> * Change the LoRA util to read from that
> 
> Wdyt?

I agree 100%.  I'll update the PR accordingly.  I'd also like to add the ability to configure LoRA alpha and rank via config parameters as well.  Should that be broken out into a separate PR?

awni:
> I'd also like to add the ability to configure LoRA alpha and rank via config parameters as well. Should that be broken out into a separate PR?

Great! I think either is fine. Up to you

chimezie:
Ok.  I have incorporated that also into this PR

awni (approved):
Thanks for adding this @chimezie, it will be super useful I think!

PR #548, opened 2024-03-08
Support for OpenAI’s fine-tuning dataset format

madroidmaq:
- [x] chat format
- [x] prompt completion pair format
- [x] huggingface chat-templates
- [x] update readme

Supports OpenAI's fine-tuning format. Currently, both methods are supported. The format is as follows:

Chat format:
```
{"messages": [{"role": "system", "content": "Marv is a factual chatbot that is also sarcastic."}, {"role": "user", "content": "What's the capital of France?"}, {"role": "assistant", "content": "Paris, as if everyone doesn't know that already."}]}
{"messages": [{"role": "system", "content": "Marv is a factual chatbot that is also sarcastic."}, {"role": "user", "content": "Who wrote 'Romeo and Juliet'?"}, {"role": "assistant", "content": "Oh, just some guy named William Shakespeare. Ever heard of him?"}]}
{"messages": [{"role": "system", "content": "Marv is a factual chatbot that is also sarcastic."}, {"role": "user", "content": "How far is the Moon from Earth?"}, {"role": "assistant", "content": "Around 384,400 kilometers. Give or take a few, like that really matters."}]}
```

prompt completion pair format：
```
{"prompt": "<prompt text>", "completion": "<ideal generated text>"}
{"prompt": "<prompt text>", "completion": "<ideal generated text>"}
{"prompt": "<prompt text>", "completion": "<ideal generated text>"}
```

For more information see: https://platform.openai.com/docs/guides/fine-tuning/example-format

Also supports  [Chat Templates](https://huggingface.co/blog/chat-templates) .

It is also compatible with current data set formats.

### DIFF
--- a/llms/mlx_lm/LORA.md
+++ b/llms/mlx_lm/LORA.md
@@ -141,37 +141,36 @@ loader expects a `test.jsonl` in the data directory.
 Currently, `*.jsonl` files support three data formats: `chat`,
 `completions`, and `text`. Here are three examples of these formats:
 
-- `chat`
+`chat`:
   
-  ```jsonl
-  {"messages": [
-      {"role": "system", "content": "You are a helpful assistant." },
-      {"role": "user", "content": "Hello."},
-      {"role": "assistant", "content": "How can I assistant you today."}
-    ]}
-  ```
+```jsonl
+{"messages": [
+  {"role": "system", "content": "You are a helpful assistant." },
+  {"role": "user", "content": "Hello."},
+  {"role": "assistant", "content": "How can I assistant you today."}
+]}
+```
 
-- `completions`
+`completions`:
   
-  ```jsonl
-  {"prompt": "What is the capital of France?", "completion": "Paris."}
-  ```
+```jsonl
+{"prompt": "What is the capital of France?", "completion": "Paris."}
+```
 
-- `text`
+`text`:
 
-  ```jsonl
-  {"text": "This is an example for the model."}
-  ```
+```jsonl
+{"text": "This is an example for the model."}
+```
 
-Note, the format is automatically determined by the dataset. Also, keys in each
-line not expected by the loader will be ignored.
+Note, the format is automatically determined by the dataset. Note also, keys in
+each line not expected by the loader will be ignored.
 
 For the `chat` and `completions` formats, Hugging Face [chat
-templates](https://huggingface.co/blog/chat-templates) are used. This apply the
-model's chat-templates definition by default. However if the model does not
-have a chat template, then Hugging Face will use a default template. For
-example, the final text of the `chat` example above with Hugging
-Face's default template becomes:
+templates](https://huggingface.co/blog/chat-templates) are used. This applies
+the model's chat template by default. If the model does not have a chat
+template, then Hugging Face will use a default. For example, the final text in
+the `chat` example above with Hugging Face's default template becomes:
 
 ```text
 <|im_start|>system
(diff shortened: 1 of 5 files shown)


fblissjr:
I think this is helpful because it indeed would cover the issue I raised that you referenced, but also allows for extensibility by other tooling without reinventing the wheel every time.

I think so long as flexible prompt templates can be passed in manually (say, as a non-default value to the OpenAI one in this PR), or by other libraries in the future, we're in a better long term state. I think it's unreasonable to try and handle every single prompt template, and makes more sense outside of it - especially until we have better standards to put a specific prompt template in the repo. 

But I do think this PR adds value by creating a different prompt templating type (or task?). 

And OpenAI is becoming a standard itself as one of those user/assistant prompt template types. Maybe the right approach now is allowing for the different tasks/types (ie: ChatML, or Alpaca-like for Instruct), and treat them as template types (Ie: pure completion, instruct, and user/assistant like this PR, making for a total of 3 in this example), rather than trying to cover every single variation.

madroidmaq:
@fblissjr I'm not sure if you've seen [Chat Templates](https://huggingface.co/blog/chat-templates), but if you haven't, I strongly suggest you take a look because I think it can answer your question (whether there is a need for real processing of dataset assembly formats similar to ChatML). I think this is a good decoupling; the dataset no longer cares about how the text will be formatted for model training, only providing the necessary content, and leaves the assembly to be handled by the tokenizer. I believe this facilitates the implementation of single responsibility.

Of course, in my view, there might also be possibilities to expand the existing dataset format to support more functionalities in the future, such as:
- Function calling, like what [CohereForAI/c4ai-command-r-v01](https://huggingface.co/CohereForAI/c4ai-command-r-v01) supports;
- RLHF (Reinforcement Learning from Human Feedback), supporting chosen and rejected options.

awni (approved):
This is great, thanks for the addition! I'm just letting the tests clear then I will merge it!

ivanfioravanti:
Thanks @madroidmaq this PR is great! Can't wait to see it merged! 🥳

alwint3r:
Very nice! Thanks

fblissjr:
> @fblissjr I'm not sure if you've seen [Chat Templates](https://huggingface.co/blog/chat-templates), but if you haven't, I strongly suggest you take a look because I think it can answer your question (whether there is a need for real processing of dataset assembly formats similar to ChatML). I think this is a good decoupling; the dataset no longer cares about how the text will be formatted for model training, only providing the necessary content, and leaves the assembly to be handled by the tokenizer. I believe this facilitates the implementation of single responsibility.
> 
> Of course, in my view, there might also be possibilities to expand the existing dataset format to support more functionalities in the future, such as:
> 
> * Function calling, like what [CohereForAI/c4ai-command-r-v01](https://huggingface.co/CohereForAI/c4ai-command-r-v01) supports;
> * RLHF (Reinforcement Learning from Human Feedback), supporting chosen and rejected options.

Thanks for the info - this helps. I agree, this PR is smart and extensible. I'm not a fan of chat templates to begin with, and I think once we get over this chat paradigm, what you said above will make even more sense.

Solido:
Hi, thanks for the update on accepted format.
Does this log any information on ignored tokens or invalid format?
Would be great to not let people train on malformed data.

Solido:
I'm also a little confused because I use the `completions` format but I do not understand this warning 

`No chat template is defined for this tokenizer - using a default chat template that implements the ChatML format (without BOS/EOS tokens!). If the default is not appropriate for your model, please set `tokenizer.chat_template` to an appropriate template. See https://huggingface.co/docs/transformers/main/chat_templating for more information.`

I've no idea where to define the next one.

Previously I was using the jsonl format with `text:`

PR #555, opened 2024-03-10
feat(mlx-lm): export the GGUF (fp16) format model weights from fuse.py

mzbac:
## Nice to Have

- [ ] Support Bpe tokenizer
- [ ] Support SentencePiece tokenizer
- [ ] export quant mlx model

## Completed Tasks

- [x] Export fp16 format model weight
- [x] Support HF tokenizer

### DIFF
--- a/llms/mlx_lm/fuse.py
+++ b/llms/mlx_lm/fuse.py
@@ -3,7 +3,6 @@ import glob
 import json
 import shutil
 from pathlib import Path
-from typing import Any, Dict, Union
 
 from mlx.utils import tree_flatten, tree_unflatten
 
--- a/llms/mlx_lm/fuse.py
+++ b/llms/mlx_lm/fuse.py
@@ -7,6 +7,7 @@ from typing import Any, Dict, Union
 
 from mlx.utils import tree_flatten, tree_unflatten
 
+from .gguf import convert_to_gguf
 from .tuner.lora import LoRALinear
 from .tuner.utils import apply_lora_layers, dequantize
 from .utils import (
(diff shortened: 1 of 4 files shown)


mzbac:
@awni As discussed here https://github.com/ml-explore/mlx/issues/814#issuecomment-1987101265, maybe we should only support converting to gguf fp16 for now. The current implementation is using hf vocab which should be good for the first cut since we are already using autoTokenizer. If there is a need to support BPE tokenizer vocab and SentencePiece tokenizer, we can add it later.

I have tested it locally for llama, mistral, and mixtral. It seems to work as expected.

llms/mlx_lm/fuse.py
```diff
@@ -90,6 +102,14 @@ def main() -> None:
     with open(save_path / "config.json", "w") as fid:
         json.dump(config, fid, indent=4)
 
+    if args.export_gguf:
+        model_type = config["model_type"]
+        if model_type not in ["llama", "mixtral", "mistral"]:
```
mzbac, line 107: Only support 'llama', 'mixtral', and 'mistral' for now.

awni:
@mzbac this is ready for review now right?

mzbac:
> @mzbac this is ready for review now right?

Yeah, it kind of works with Llama, Mixtral, and Mistral. However, I am not sure if only providing GGUF FP16 would add that much value since people still need to use llama.cpp to quantize it for more practical usage. Ideally, if we can provide Q4 quantization for GGUF, it may be more beneficial. But I may need some guidance from you on how we can convert MLX weight to GGUF Q4_0 quantization.

awni (commented):
This is pretty cool! I don't really have much to comment here. I'd like to test it out locally but otherwise I think we can merge it. Indeed I agree it would be even cooler if we could export quantized models directly. We'll have to think about if/how we can enable that in mlx.core.

awni (approved):
Thank you!!

PR #563, opened 2024-03-11
Fix image2image for SDXL

devonthomas35:


angeloskath (approved):
Nice, thanks for the update of image2image.

Issue #564, opened 2024-03-11
Qwen2: Received parameters not in model: lm_head.weight


madroidmaq:
对  `Qwen/Qwen1.5-0.5B-Chat` 进行微调，报错信息如下：

```
Loading pretrained model
Fetching 6 files: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 6/6 [00:00<00:00, 20832.64it/s]
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/madroid/develop/workspace/mlx-examples/llms/mlx_lm/lora2.py", line 164, in <module>
    model, tokenizer = load(args.model)
                       ^^^^^^^^^^^^^^^^
  File "/Users/madroid/develop/workspace/mlx-examples/llms/mlx_lm/utils.py", line 385, in load
    model = load_model(model_path, lazy)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/madroid/develop/workspace/mlx-examples/llms/mlx_lm/utils.py", line 349, in load_model
    model.load_weights(list(weights.items()))
  File "/opt/homebrew/lib/python3.11/site-packages/mlx/nn/layers/base.py", line 203, in load_weights
    raise ValueError(f"Received parameters not in model: {extras}.")
ValueError: Received parameters not in model: lm_head.weight.
```

我怀疑是 https://github.com/ml-explore/mlx-examples/pull/544 提交引入的问题，我 revert 对应提交之后再次运行，他就可以正常工作了。

mzbac:
Maybe we need to be smarter to handle the `tie_word_embeddings` configuration. It seems that qwen2 enabled the tie_word_embeddings but also provided lm_head.

mzbac:
A potential solution would be to always initialize the lm_head and, based on whether tie_word_embeddings is enabled or not, copy the weight from token embed to lm_head during load weights. Double-check that qwen2 0.5B pretty much copies the embedding to lm_head.
```
import glob
from mlx_lm.utils import get_model_path, load_model
import mlx.core as mx

model_path = get_model_path("Qwen/Qwen1.5-0.5B-Chat")

weight_files = glob.glob(str(model_path / "*.safetensors"))

weights = {}
for wf in weight_files:
    weights.update(mx.load(wf))

print(
    mx.allclose(
[...]

PR #566, opened 2024-03-12
chore(mlx-lm): fix tie_word_embeddings for qwen2

mzbac:
fix: https://github.com/ml-explore/mlx-examples/issues/564

### DIFF
--- a/llms/tests/test_models.py
+++ b/llms/tests/test_models.py
@@ -129,6 +129,47 @@ class TestModels(unittest.TestCase):
             model, args.model_type, args.vocab_size, args.num_hidden_layers
         )
 
+    def test_qwen2_tie_word_embeddings_without_lm_head_weight(self):
+        from mlx_lm.models import qwen2
+
+        args = qwen2.ModelArgs(
+            model_type="qwen2",
+            hidden_size=1024,
+            num_hidden_layers=4,
+            intermediate_size=2048,
+            num_attention_heads=4,
+            rms_norm_eps=1e-5,
+            vocab_size=10_000,
+            tie_word_embeddings=True,
+        )
+        model = qwen2.Model(args)
+        weights = {"model.embed_tokens.weight": "some_value"}
+        sanitized_weights = model.sanitize(weights)
+        self.assertIn("lm_head.weight", sanitized_weights)
+        self.assertEqual(sanitized_weights["lm_head.weight"], "some_value")
+
+    def test_qwen2_tie_word_embeddings_with_lm_head_weight(self):
+        from mlx_lm.models import qwen2
+
+        weights = {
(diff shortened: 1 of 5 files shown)


awni:
There is one thing I'm unsure about here and that is what happens when we quantize the model?

If `tie_word_embeddings = True` and we add a linear output layer, then that output layer will be quantized. I think it should work in the sense that the quantized output weights will be saved and then we will use them directly. But maybe that is not the desired behavior?

mzbac:
> There is one thing I'm unsure about here and that is what happens when we quantize the model?
> 
> If `tie_word_embeddings = True` and we add a linear output layer, then that output layer will be quantized. I think it should work in the sense that the quantized output weights will be saved and then we will use them directly. But maybe that is not the desired behavior?

Would you please elaborate on your concern a bit? It makes sense to me that once we save the quant weights which include the cloned lm_head, then when we load the quant model with saved the cloned lm_head, the model should work as expected.

I think maybe we should default tie_word_embeddings to true in the model arguments, since the transformers default tie_word_embeddings to be true. just in case some model configurations are missing the default tie_word_embeddings in the config file.

awni:
> Would you please elaborate on your concern a bit? It makes sense to me that once we save the quant weights which include the cloned lm_head, then when we load the quant model with saved the cloned lm_head, the model should work as expected.

I think it's probably minor. But, what I was thinking is maybe it's unexpected if you have `tie_word_embeddings=True` that you have a quantized LM head which is separate from the embeddings. My expectation if that flag is true is that the two layers use the exact same weights (as in the same memory blob).

llms/tests/test_models.py
```diff
@@ -129,6 +129,47 @@ def test_qwen2(self):
             model, args.model_type, args.vocab_size, args.num_hidden_layers
         )
 
+    def test_qwen2_tie_word_embeddings_without_lm_head_weight(self):
```
awni, line 132: Thanks for adding some tests ❤️

mzbac:
> > Would you please elaborate on your concern a bit? It makes sense to me that once we save the quant weights which include the cloned lm_head, then when we load the quant model with saved the cloned lm_head, the model should work as expected.
> 
> I think it's probably minor. But, what I was thinking is maybe it's unexpected if you have `tie_word_embeddings=True` that you have a quantized LM head which is separate from the embeddings. My expectation if that flag is true is that the two layers use the exact same weights (as in the same memory blob).

My thought is that some models have already cloned the token embedding to lm_head in their released weights (e.g. Qwen2 0.5B for instances), and our quantization seems to work fine with that case. I am not sure if we still need to keep those two layers the same by introducing some checks in the quantization part.

awni (approved):
Thanks for the fix! 🚀

PR #571, opened 2024-03-13
bert encoder inherits from nn.Module now

raceee:
I was using MLX to do some masked language modeling using bert model example, and I wasn't able to

```optimizer.update(model.trainable_parameters(), grads)```

Like how it is done in the MLP example documentation [here](https://ml-explore.github.io/mlx/build/html/examples/mlp.html)

After adding ```super().__init__()``` to both the ```BERT``` class and the ```BertEmbeddings``` my training loop proceeded.


awni (approved):
Oops! Good catch!

PR #574, opened 2024-03-13
Make attention faster for a some models

### LINKED ISSUE #526: Issue with fetch_from_hub tokenizer for fine-tuned models
When trying to convert fine-tuned weights of StarCode2 via 

```python
python -m mlx_lm.convert \
    --hf-path m-a-p/OpenCodeInterpreter-SC2-7B \
    -q \
    --upload-repo mlx-community/OpenCodeInterpreter-SC2-7B-4bit
```

I encountered 

```
->    tokenizer = AutoTokenizer.from_pretrained(model_path_hf)
    return cls._from_pretrained(
  File "/Users/x/miniconda3/lib/python3.10/site-packages/transformers/tokenization_utils_base.py", line 2082, in _from_pretrained
    slow_tokenizer = (cls.slow_tokenizer_class)._from_pretrained(
  File "/Users/x/miniconda3/lib/python3.10/site-packages/transformers/tokenization_utils_base.py", line 2287, in _from_pretrained
    tokenizer = cls(*init_inputs, **init_kwargs)
  File "/Users/x/miniconda3/lib/python3.10/site-packages/transformers/models/gpt2/tokenization_gpt2.py", line 187, in __init__
    with open(merges_file, encoding="utf-8") as merges_handle:
**TypeError: expected str, bytes or os.PathLike object, not NoneType**
```

I printed path and config to inspect 

```
/Users/x/.cache/huggingface/hub/models--m-a-p--OpenCodeInterpreter-SC2-7B/snapshots/ddf1d84dfc2adbc7ee9eed97298de13fc232c300
Config: Starcoder2Config {
  "_name_or_path": 
"/Users/x/.cache/huggingface/hub/models--m-a-p--OpenCodeInterpreter-SC2-7B/snapshots/ddf1d84dfc2adbc7ee9eed97298de13fc232c30
0",
```

After changing  `model_path` to non snapshotted original name issue disappeared

```python
from rich import print as rprint

def fetch_from_hub(
    model_path: Path, lazy: bool = False
) -> Tuple[nn.Module, dict, PreTrainedTokenizer]:
    model = load_model(model_path, lazy)
    print(model_path)
    config = AutoConfig.from_pretrained(model_path)
    rprint(f"Config: {config}")
    _model_path_hf = "m-a-p/OpenCodeInterpreter-SC2-7B"_
    tokenizer = AutoTokenizer.from_pretrained(model_path_hf)

    return model, config.to_dict(), tokenizer
```

@awni  maybe `utils.py` should not modify  ` "_name_or_path": `

awni:
- Use SDPA for Mixtral
- Use no repeat for phi
- Store keys in fp16 for phi, phixtral

Also closes #526

### DIFF
--- a/llms/mlx_lm/LORA.md
+++ b/llms/mlx_lm/LORA.md
@@ -167,6 +167,12 @@ of memory. Here are some tips to reduce memory use should you need to do so:
    you can do is break your examples into smaller
    sequences when making the `{train, valid, test}.jsonl` files.
 
+5. Gradient checkpointing lets you trade-off memory use (less) for computation
+   (more) by recomputing instead of storing intermediate values needed by the
+   backward pass. You can use gradient checkpointing by passing the
+   `--grad-checkpoint` flag. Gradient checkpointing will be more helpful for
+   larger batch sizes or sequence lengths with smaller or quantized models.
+
 For example, for a machine with 32 GB the following should run reasonably fast:
 
 ```
--- a/llms/mlx_lm/lora.py
+++ b/llms/mlx_lm/lora.py
@@ -61,19 +61,6 @@ def build_parser():
         "--model",
         help="The path to the local model directory or Hugging Face repo.",
     )
-    parser.add_argument(
-        "--max-tokens",
-        "-m",
-        type=int,
-        help="The maximum number of tokens to generate",
-    )
-    parser.add_argument("--temp", type=float, help="The sampling temperature")
-    parser.add_argument(
-        "--prompt",
-        "-p",
-        type=str,
-        help="The prompt for generation",
-    )
 
     # Training args
     parser.add_argument(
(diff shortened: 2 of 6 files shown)


angeloskath (approved):
This looks great!

Out of curiosity and possibly showing my ignorance, why not use `mx.fast.scaled_dot_product_attention` for phi and phixtral as well?

awni:
> Out of curiosity and possibly showing my ignorance, why not use mx.fast.scaled_dot_product_attention for phi and phixtral as well?

Yes it's really important for phi to upcast the queries (and keys implicitly) prior to the `query @ key` matmul. Then you downcast back to fp16 right after the softmax but before you multiply the scores with the values. SDPA would upcast everything so you would also do the `scores @ value` (and everything else) in fp32 which is a bit slower.

What I still don't understand is why Phi needs the `query @ key` to be in higher precision.. but it definitely doesn't work if you do that matmul in fp16. You can't even do the matmul and then down cast right after, I think the range of those values is just too large.

PR #580, opened 2024-03-14
Enable more BERT models

yzimmermann:
I changed some things that were hard coded before that can be retrieved from `config.json` to enable more models with the `BertForMaskedLM` architecture. I tested it with some other models, e.g.

> python convert.py --bert-model "unitary/toxic-bert" --mlx-model weights/toxic-bert.npz    
> python test.py --bert-model "unitary/toxic-bert" --mlx-model "weights/toxic-bert.npz" --text "Hello World!" 

> Tests pass :)

It should in theory also work with RoBERTa models, but I haven't quite gotten it to run yet.

### DIFF
--- a/bert/README.md
+++ b/bert/README.md
@@ -1,6 +1,6 @@
 # BERT
 
-An implementation of BERT [(Devlin, et al., 2019)](https://aclanthology.org/N19-1423/) within MLX.
+An implementation of BERT [(Devlin, et al., 2019)](https://aclanthology.org/N19-1423/) in MLX.
 
 ## Setup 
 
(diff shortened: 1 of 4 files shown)


yzimmermann:
> Really nice! When you say RoBERTA is not working is it giving bad output or just crashing?

Bad output.

yzimmermann:
Upon reviewing the BERT and RoBERTa architectures it seems like we do need to tweak the model file a bit.

BERT
image


RoBERTa
image

awni (approved):
Looks great!! Nice to add support for more models without making it any more complex! Thanks!

PR #581, opened 2024-03-14
chore(mlx-lm): fix print_trainable_parameters for quant models

mzbac:
address https://huggingface.co/mlx-community/Mixtral-8x7B-Instruct-v0.1-hf-4bit-mlx/discussions/1

### DIFF
--- a/llms/mlx_lm/lora.py
+++ b/llms/mlx_lm/lora.py
@@ -196,27 +196,15 @@ def load_dataset(args):
 
 
 def print_trainable_parameters(model):
+    def nparams(m):
+        if isinstance(m, nn.QuantizedLinear):
+            return m.weight.size * (32 // m.bits)
+        return sum(v.size for _, v in tree_flatten(m.parameters()))
 
-    quantization_bits = 32
-    for _, module in model.named_modules():
-        if isinstance(module, nn.QuantizedLinear):
-            quantization_bits = module.bits
-            break
-
-    all_params = tree_flatten(model.parameters())
-    quantized_params = {
-        k.rsplit(".", 1)[0] for k, _ in all_params if ".biases" in k or ".scales" in k
-    }
-    total_p = 0
-    for k, v in all_params:
-        if "weight" in k:  # skip biases and scales and lora parameters
-            base_name = k.rsplit(".", 1)[0]
-            if base_name not in quantized_params:
-                total_p += v.size
-            else:
-                total_p += v.size * (32 // quantization_bits)
-
-    total_p /= 10**6
+    leaf_modules = tree_flatten(
+        model.leaf_modules(), is_leaf=lambda m: isinstance(m, nn.Module)
+    )
+    total_p = sum(nparams(m) for _, m in leaf_modules) / 10**6
     trainable_p = (
         sum(v.size for _, v in tree_flatten(model.trainable_parameters())) / 10**6
     )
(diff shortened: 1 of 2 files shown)


llms/tests/test_lora.py
```diff
@@ -47,6 +56,69 @@ def check_config(params):
         params["keys"] = ["self_attn.k_proj"]
         check_config(params)
 
+    def test_quantized_print_trainable_parameters(self):
+        model = MagicMock()
+        model.parameters.return_value = {
+            "layer1.weight": MagicMock(size=1e6),
```
awni, line 62: That's cool, I never saw it before.

llms/mlx_lm/lora.py
```diff
@@ -209,14 +223,15 @@ def run(args, training_callback: TrainingCallback = None):
...
+    model_path = get_model_path(args.model)
+    model, config, tokenizer = fetch_from_hub(model_path)
 
     # Freeze all layers
     model.freeze()
     # Convert linear layers to lora layers and unfreeze in the process
     linear_to_lora_layers(model, args.lora_layers, args.lora_parameters)
 
-    print_trainable_parameters(model)
+    print_trainable_parameters(model, config)
```
awni, line 234: I feel like we should get this function to work without needing to pass the config as well. It would be a more general solution.

The quantized layers have `bits` and `group_size` members, maybe we should read from that?

mzbac: Yeah, that's a much better solution. I didn't notice we had it in the QuantizedLinear layer. I will update it to use that instead.

mzbac:
@awni, I have updated the PR based on your suggestion. It is now ready for review again. Please let me know if there is anything else that needs to be updated.

llms/mlx_lm/lora.py
```diff
@@ -195,7 +196,27 @@ def load_dataset(args):
...
+    total_p = 0
+    for k, v in all_params:
+        if "weight" in k:  # skip biases and scales and lora parameters
+            base_name = k.rsplit(".", 1)[0]
+            if base_name not in quantized_params:
+                total_p += v.size
+            else:
+                total_p += v.size * (32 // quantization_bits)
+
+    total_p /= 10**6
```
awni, line 219: It was not as easy as expected to use our base methods to implement this, but I think you can do something a lot simpler like the following. It's more general as it will get the bits for each model separately and so it's good to use the pattern here in case others copy it.

```suggestion
    def nparams(m):
        if isinstance(m, nn.QuantizedLinear):
            return m.weight.size * (32 // m.bits)
        return sum(v.size for _, v in tree_flatten(m.parameters()))
    leaf_modules = tree_flatten(model.leaf_modules(), is_leaf=lambda m: isinstance(m, nn.Module))
    total_p = sum(nparams(m) for _, m in leaf_modules) / 10**6
```

mzbac: Looks much clearer. Just a quick question, will this include the LoRa parameter as part of the total parameters?

awni: I will indeed...and I noticed that difference between the new and old version. I was thinking about whether it makes sense to include them or not.. 

I think it's fine (maybe better) to include, but you could also remove the trainable params from the `total_p` if you wanted to exclude them.

mzbac: @Awni updated the test and did a quick local test, everything looks good.

awni (commented):
🙏  thanks for fixing that!

awni (approved):

PR #599, opened 2024-03-20
Add dropout parameter to lora configuration

ivanfioravanti:
A dropout parameter has been added to the lora configuration settings in lora_config.yaml. The LoRALinear class in utils.py has been updated to take this new parameter. Additionally, a AttributeError: 'types.SimpleNamespace' object has no attribute 'prompt' related to `args.prompt` has been removed from lora.py.

### DIFF
--- a/llms/mlx_lm/examples/lora_config.yaml
+++ b/llms/mlx_lm/examples/lora_config.yaml
@@ -59,3 +59,4 @@ lora_parameters:
   rank: 8
   alpha: 16.0
   scale: 10.0
+  dropout: 0.05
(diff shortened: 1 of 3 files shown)


ivanfioravanti:
I added dropout even if it's not the most effective parameters in fine-tuning, but can be useful tweaking it.
I remove the check on args.prompt because it was generating an error at the end of the training and moreover it's already checked at the beginning by generate.py args parser.

mzbac:
Just a quick note, there used to be an issue when enabling lora dropoff. I'm not sure if the issue has been fixed yet. cc @awni

awni:
> I'm not sure if the issue has been fixed yet.

Indeed, It's been fixed in https://github.com/ml-explore/mlx/pull/813

awni:
I tried to push a format fix to you branch `ivanfioravanti:additional_config_params` but it didn't reflect in the PR 🤷

awni (approved):
Thanks!!

PR #602, opened 2024-03-21
chore(mlx-lm): fix the top_p implementation.

mzbac:
The top_p is supposed to sort prob in **ascending** order and select the cumulative token prob above the threshold. The current implementation works because we have a bug in reversing the sorted prob, which accidentally makes sorted prob in the correct order. So, it's better to fix the implementation to reduce the confusion.

### DIFF
--- a/llms/mlx_lm/sample_utils.py
+++ b/llms/mlx_lm/sample_utils.py
@@ -1,7 +1,7 @@
 import mlx.core as mx
 
 
-def top_p_filtering(logits: mx.array, top_p: float, temperature: float) -> mx.array:
+def top_p_sampling(logits: mx.array, top_p: float, temperature: float) -> mx.array:
     """
     Apply top-p (nucleus) filtering to logits using cumulative probabilities.
 
@@ -21,8 +21,9 @@ def top_p_filtering(logits: mx.array, top_p: float, temperature: float) -> mx.ar
     probs = mx.softmax(logits / temperature, axis=-1)
 
     # sort probs in ascending order
-    sorted_probs = mx.sort(probs, axis=-1)
     sorted_indices = mx.argsort(probs, axis=-1)
+    sorted_probs = probs[..., sorted_indices]
+
     cumulative_probs = mx.cumsum(sorted_probs, axis=-1)
 
     # select tokens with cumulative probs below threshold
--- a/llms/mlx_lm/sample_utils.py
+++ b/llms/mlx_lm/sample_utils.py
@@ -3,7 +3,7 @@ import mlx.core as mx
 
 def top_p_sampling(logits: mx.array, top_p: float, temperature: float) -> mx.array:
     """
-    Apply top-p (nucleus) filtering to logits using cumulative probabilities.
+    Apply top-p (nucleus) sampling to logits.
 
     Args:
         logits: The logits from the model's output.
--- a/llms/mlx_lm/sample_utils.py
+++ b/llms/mlx_lm/sample_utils.py
@@ -1,0 +1,38 @@
+import mlx.core as mx
+
+
+def top_p_filtering(logits: mx.array, top_p: float, temperature: float) -> mx.array:
+    """
+    Apply top-p (nucleus) filtering to logits using cumulative probabilities.
+
+    Args:
+        logits: The logits from the model's output.
+        top_p: The cumulative probability threshold for top-p filtering.
+        temperature: Temperature parameter for softmax distribution reshaping.
+    Returns:
+        token selected based on the top-p criterion.
+    """
+    if (
+        logits.dtype == mx.bfloat16
+    ):  # workaround for unable to load kernel contiguous_scan_inclusive_sum_bfloat16_bfloat16
+        logits = logits.astype(mx.float32)
+
+    # referenced implementation from https://github.com/huggingface/transformers/blob/main/src/transformers/generation/logits_process.py#L449-L460
+    probs = mx.softmax(logits / temperature, axis=-1)
+
+    # sort probs in ascending order
+    sorted_probs = mx.sort(probs, axis=-1)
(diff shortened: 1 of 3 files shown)


llms/mlx_lm/utils.py
```diff
@@ -143,17 +143,21 @@ def sample(logits: mx.array) -> Tuple[mx.array, float]:
         if temp == 0:
             token = mx.argmax(logits, axis=-1)
         else:
+            # referenced implementation from https://github.com/huggingface/transformers/blob/main/src/transformers/generation/logits_process.py#L449-L460
             if top_p > 0 and top_p < 1.0:
                 if (
                     logits.dtype == mx.bfloat16
                 ):  # workdaround for unable to load kernel contiguous_scan_inclusive_sum_bfloat16_bfloat16
                     logits = logits.astype(mx.float32)
+
```
awni, line 152: Do you mind refactoring this `top_p` into a little function we can add a test for it?

llms/mlx_lm/utils.py
```diff
@@ -143,17 +143,21 @@ def sample(logits: mx.array) -> Tuple[mx.array, float]:
...
                 probs = mx.softmax(logits / temp, axis=-1)
 
-                sorted_probs = mx.sort(probs)[::-1]
-                sorted_indices = mx.argsort(probs)[::-1]
+                # sort probs in ascending order
+                sorted_probs = mx.sort(probs, axis=-1)
```
mzbac, line 156: I am not sure if the MLX sort implementation changes or not, but I did some tests today and it seems that the MLX sort returns the array in ascending order.

llms/mlx_lm/utils.py
mzbac: Sure thing, could you help double-check if the implementation makes sense to you now? Once we are confident in the implementation, I can add tests for it.

llms/mlx_lm/utils.py
awni: It has not changed, it's always been ascending (following numpy)

mzbac: Cool, so the original implementation was wrong then, it must have been mixed up with the PyTorch sort implementation.

llms/mlx_lm/utils.py
```diff
@@ -143,17 +143,21 @@ def sample(logits: mx.array) -> Tuple[mx.array, float]:
...
+                sorted_indices = mx.argsort(probs, axis=-1)
                 cumulative_probs = mx.cumsum(sorted_probs, axis=-1)
```
awni, line 158: I still think you need to flip the probabilities here. You want the smallest set of tokens with total mass > p.

mzbac: > I still think you need to flip the probabilities here. You want the smallest set of tokens with total mass > p.

@awni would you please elaborate on this? Do you mean that when using cumsum, you want the cumulative sum to be in descending order?

awni:
Btw @mzbac looks like the CI issue is resolved for you, thanks!!

llms/mlx_lm/utils.py
mzbac: Cool, it seems there are no major blockers. I will start working on adding the test.

mzbac:
@awni I have added the unit test, so it is ready for review again. Please let me know if there is anything else that needs to be updated.

llms/mlx_lm/sample_utils.py
```diff
@@ -0,0 +1,38 @@
...
+    if (
+        logits.dtype == mx.bfloat16
+    ):  # workaround for unable to load kernel contiguous_scan_inclusive_sum_bfloat16_bfloat16
+        logits = logits.astype(mx.float32)
+
+    # referenced implementation from https://github.com/huggingface/transformers/blob/main/src/transformers/generation/logits_process.py#L449-L460
+    probs = mx.softmax(logits / temperature, axis=-1)
+
+    # sort probs in ascending order
+    sorted_probs = mx.sort(probs, axis=-1)
```
angeloskath, line 24: Why not `sorted_probs = probs[..., sorted_indices]`. It may be faster. Obviously having a multi-output primitive that returns both would be the best here.

llms/mlx_lm/sample_utils.py
```diff
@@ -0,0 +1,38 @@
+import mlx.core as mx
+
+
+def top_p_filtering(logits: mx.array, top_p: float, temperature: float) -> mx.array:
```
angeloskath, line 4: I would change the name to `top_p_sampling` rather than filtering, or even `nucleus_sampling`.

mzbac: Yeah, I renamed it to top_p_sampling because I personally feel that "top_p" is more widely used. Let me know if you still think `nucleus_sampling` is better.

angeloskath: No it's great! My comment was mostly to use sampling over filtering.

awni:
I'm not sure about introducing a separate file for that. Do you think it's necessary? Is the idea that we will add more samplers?

mzbac:
The main reason for putting it into a separate file is the current utils.py has become too large, making it difficult to maintain. So, I think we may need to start splitting up the utils.py file. sampling utils might be a good starting point. Sampling utils would include top_p, repetitive penalty and any future sampling methods that we want to support.

awni:
> So, I think we may need to start splitting up the utils.py file. 

Agreed.

> Sampling utils would include top_p, repetitive penalty and any future sampling methods that we want to support.

I'm not wild about the name `sampling_utils` but I think collecting that related functionality in a new file makes sense. Let's go with it for now but I may change the file name in the future if I can think of a better one 😄

awni (approved):
🙏 thanks for fixing that and for the tests!

PR #604, opened 2024-03-21
Configurable LR schedulers

chimezie:
Updates the mlx_lm LoRA module's YAML configuration to include sections like the following:

```yaml
schedule:
  join:
    boundaries: [101]
    schedules:
      - { name: linear_schedule,
          arguments: [0.0, 1e-5, 100] }
      - { name: cosine_decay,
          arguments: [ 1e-5, 100 ] }
```

and creates joined schedules using parameters specified in the configuration or single schedules if specified like so

```yaml
schedule:
  cosine_decay:
    arguments: [ 0.1, 10 ]
```

The schedule names refer to those in **mlx.optimizers.schedulers** and their arguments are the same as those in the functions defined there.

Currently, mlx_lm hardcodes to using optim.Adam, but this configuration framework can be extended to also allow the optimizer to be specified to use others, such as SGD, for example.


ivanfioravanti:
Great one! @chimezie

awni:
@chimezie I think we can allow a scheduler config but let's keep it simpler for now. What you've built is very flexible but I think for most cases people don't need that much flexibility and would actually benefit from some useful constraints. I would provide the option to do a warmup and maybe the choice of schedule (e.g. cosine vs exponential) with arguments. So something like:

```
schedule:
  name: cosine_decay
  warmup: 100 # 0 for no warmup
  arguments: .... # get passed to scheduler
```

llms/mlx_lm/schedule_config.py
madroidmaq, line 1: Would it be more appropriate to place it under the tuner directory, as it seems that only fine-tuning related logic would use it?

chimezie:
> @chimezie I think we can allow a scheduler config but let's keep it simpler for now. What you've built is very flexible but I think for most cases people don't need that much flexibility and would actually benefit from some useful constraints. I would provide the option to do a warmup and maybe the choice of schedule (e.g. cosine vs exponential) with arguments. So something like:
> 
> ```
> schedule:
>   name: cosine_decay
>   warmup: 100 # 0 for no warmup
>   arguments: .... # get passed to scheduler
> ```

It would be useful to also be able to specify the warmup minimum and have it default to 0 like so:

```yaml
schedule:
    name: cosine_decay
    warmup: 100 # 0 for no warmup
    minimum: 1e-6 # 0 if not specified
    arguments: ...   
```

Thoughts?

awni:
> It would be useful to also be able to specify the warmup minimum and have it default to 0 like so:
What about;

```yaml
warmup: 100 1e-1
```

If the second value is not set, it defaults to 0? If `warmup` is not set then no warmup is used.

llms/tests/test_scheduler_config.py
```diff
@@ -0,0 +1,89 @@
+import math
```
awni, line 1: Thanks for adding tests, that's really great. I do not think we need to make a separate file for this. Could you move tests to `test_lora.py`?

chimezie:
> > It would be useful to also be able to specify the warmup minimum and have it default to 0 like so:
> > What about;
> 
> ```yaml
> warmup: 100 1e-1
> ```
> 
> If the second value is not set, it defaults to 0? If `warmup` is not set then no warmup is used.

That causes the Yaml parser to interpret the argument to warmup as a string  instead of a float:

```python
import yaml
from mlx_lm.lora import yaml_loader

In [1]: CONFIG_YAML6 = """
   schedule:
      name: cosine_decay
      warmup: 100 1e-1
      minimum: 1e-6
      arguments: [ 1e-5, 20 ] """

In [2]: config = yaml.load(CONFIG_YAML6, yaml_loader)
In [3]: config
Out[3]: 
{'schedule': {'name': 'cosine_decay',
  'warmup': '100 1e-1',
  'minimum': 1e-06,
  'arguments': [1e-05, 20]}}

``` 

A separate key/value field would allow the Yaml parser to handle all the casting, etc.

awni (approved):
Nice addition, thanks!!

PR #610, opened 2024-03-24
fix(mlx-lm): sorted probs in top_p implementation.

mzbac:
Just realized that the sorted probs indexing introduced a new dimension, so we have to squeeze it out. Also, update the test to cover the shape changes.

### DIFF
--- a/llms/mlx_lm/sample_utils.py
+++ b/llms/mlx_lm/sample_utils.py
@@ -22,7 +22,8 @@ def top_p_sampling(logits: mx.array, top_p: float, temperature: float) -> mx.arr
 
     # sort probs in ascending order
     sorted_indices = mx.argsort(probs, axis=-1)
-    sorted_probs = probs[..., sorted_indices]
+    # probs shape is [B,V] and after indexing it will be [1, B, V], so we squeeze it back to [B, V]
+    sorted_probs = probs[..., sorted_indices].squeeze(0)
 
     cumulative_probs = mx.cumsum(sorted_probs, axis=-1)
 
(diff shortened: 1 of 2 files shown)


llms/mlx_lm/sample_utils.py
```diff
@@ -22,7 +22,8 @@ def top_p_sampling(logits: mx.array, top_p: float, temperature: float) -> mx.arr
 
     # sort probs in ascending order
     sorted_indices = mx.argsort(probs, axis=-1)
-    sorted_probs = probs[..., sorted_indices]
+    # probs shape is [B,V] and after indexing it will be [1, B, V], so we squeeze it back to [B, V]
+    sorted_probs = probs[..., sorted_indices].squeeze(0)
```
awni, line 26: That doesn't make sense to me, where does it get the `1` from?

```python
>>> import mlx.core as mx
>>> a = mx.zeros((4, 8))
>>> a[..., mx.array([2,3])].shape
(4, 2)
```

awni: I think I understand now.. it's probably clearer if you do 
```python
sorted_indices = mx.argsort(probs, axis=-1)
probs[..., sorted_indices.squeeze(0)]
```

llms/tests/test_sample_utils.py
```diff
@@ -18,6 +18,7 @@ def test_top_p_sampling(self, mock_categorical):
         expected_top_probs = mx.array([[0.0, 0.0, 0.0, 0.643914]])
         self.assertTrue(mx.allclose(token, expected_token))
         args, _ = mock_categorical.call_args
+        self.assertTrue(args[0].shape == expected_top_probs.shape)
```
awni, line 21: 👍

llms/mlx_lm/sample_utils.py
mzbac: The shape (1, B, V) is a bit surprising to me. I didn't really understand it, but it seems to be the same behavior in numpy.

awni (approved):
Thanks!!

PR #614, opened 2024-03-25
LoRA: Extract small function

madroidmaq:
Extract the `load_with_lora` function for easy use in other files.

### DIFF
--- a/llms/mlx_lm/examples/lora_config.yaml
+++ b/llms/mlx_lm/examples/lora_config.yaml
@@ -60,8 +60,7 @@ lora_parameters:
   # These will be applied for the last lora_layers
   keys: ["self_attn.q_proj", "self_attn.v_proj"]
   rank: 8
-  alpha: 16.0
-  scale: 10.0
+  scale: 20.0
   dropout: 0.0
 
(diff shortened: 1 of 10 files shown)


mzbac:
Maybe we should consider moving the function out of utils.py. It has becomed too large now,  it might be better to create a new file named `lora_utils.py` and place it there?

madroidmaq:
@mzbac Great suggestion. Initially, I tried placing the `load_with_lora` function in both `tuner/lora.py` and `tuner/utils.py`, but encountered a circular dependency issue. When I attempted to resolve this issue, I found that it required significant changes. So, I ended up putting it in the `utils.py`.

I will reassess the extent of changes needed to solve the circular dependency problem. The worst-case scenario would be creating a new file named `tuner/lora_utils.py`, which I will try my best to avoid.

awni:
@madroidmaq sorry for the delay! I think we can merge this. But could you first rebase to resolve conflicts?

madroidmaq:
> @madroidmaq sorry for the delay! I think we can merge this. But could you first rebase to resolve conflicts?

It's DONE.

awni (approved):
Thanks!

Issue #622, opened 2024-03-26
When loading an adapter in mlx_lm, the rank, alpha and dropout are not loaded to what was trained with


ryanmitts:
I'm not too sure if this is a problem, but when when looking through the LoRA example, it looks like when you load an adapter file back, it doesn't make sure to create the LoRALinear layers with the [same arguments](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/tuner/utils.py#L94) that were used to train.

Do these values need to be saved in a metadata file alongsize the adapters.npz file and loaded back with it?

mzbac:
The Lora adapters' weight contains information about the `r` and `alpha`, so it can be recovered by loading the weight into the Lora layer. However, I think that if a user does not use the default scale, this may cause issues when applying the Lora layer as that information is not saved anywhere. I am not sure if simply changing the scale to an mx array which will be saved as part of the Lora adapter would fix this issue or if we have to do something else to save the scale value somewhere so we can recover it when loading the adapter. @awni

rshah918:
@awni Hey, my fine tuning is disappearing after fusing. Maybe its because I set a higher scale during training but not during fusing? 

How can I configure the scale for fusing? You previously linked me to the [examples directory ](https://github.com/ml-explore/mlx-examples/blob/297a908e3db205a5d35acf806a30addbb2515788/llms/mlx_lm/examples/lora_config.yaml#L61C3-L61C14)
but seems like thats just a demo config file thats not being used by fuse.py.

awni:
There is a WIP PR for solving exactly that problem #623 

At the moment, if you change the scale through the config indeed it will not be reflected when you then reload the adapters.. that was an oversight when we made that parameter configurable. You can temporarily fix it by editing the scale here https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/tuner/lora.py#L16

rshah918:
@awni Awesome, thank you! The issue persists even when setting scale=7000. Confirmed that the pre-fused model works well. Will continue to try even larger values and report back what I observe

rshah918:
Tried scale=999999999999999.0 yet no effect on fused model output. Fusing with the command 
`python3 -m mlx_lm.fuse --model mlx_model_q --adapter-file adapters.npz --de-quantize`

rshah918:
[...]

PR #623, opened 2024-03-26
fix(mlx-lm): missing Lora scale and alpha parameters when loading Lora adapter weights

mzbac:
Make the Lora scale as an array value so it can be saved as part of the Lora weight so it can be recovered when loading the Lora weight to lora layer.

### DIFF
--- a/llms/mlx_lm/tuner/lora.py
+++ b/llms/mlx_lm/tuner/lora.py
@@ -78,6 +78,13 @@ class LoRALinear(nn.Module):
         bias: bool = False,
     ):
         super().__init__()
+        # store hyperparameters
+        self.lora_config = {
+            "r": r,
+            "alpha": alpha,
+            "dropout": dropout,
+            "scale": scale,
+        }
 
         # Regular linear layer weights
         self.linear = nn.Linear(input_dims, output_dims, bias=bias)
@@ -85,8 +92,7 @@ class LoRALinear(nn.Module):
         self.dropout = nn.Dropout(p=dropout)
 
         # Scale for low-rank update
-        self.scale = mx.array(scale * (alpha / r))
-        self.freeze(keys=["scale"], recurse=False)
+        self.scale = scale * (alpha / r)
 
         # Low rank lora weights
         scale = 1 / math.sqrt(input_dims)
--- a/llms/mlx_lm/tuner/lora.py
+++ b/llms/mlx_lm/tuner/lora.py
@@ -86,6 +86,7 @@ class LoRALinear(nn.Module):
 
         # Scale for low-rank update
         self.scale = mx.array(scale * (alpha / r))
+        self.freeze(keys=["scale"], recurse=False)
 
         # Low rank lora weights
         scale = 1 / math.sqrt(input_dims)
--- a/llms/mlx_lm/tuner/lora.py
+++ b/llms/mlx_lm/tuner/lora.py
@@ -85,7 +85,7 @@ class LoRALinear(nn.Module):
         self.dropout = nn.Dropout(p=dropout)
 
         # Scale for low-rank update
-        self.scale = scale * (alpha / r)
+        self.scale = mx.array(scale * (alpha / r))
 
         # Low rank lora weights
         scale = 1 / math.sqrt(input_dims)
--- a/llms/mlx_lm/tuner/utils.py
+++ b/llms/mlx_lm/tuner/utils.py
@@ -1,4 +1,6 @@
+import json
 import os
+import warnings
 from typing import Dict
 
 import mlx.core as mx
@@ -83,7 +85,14 @@ def apply_lora_layers(model: nn.Module, adapter_file: str) -> nn.Module:
     if not os.path.exists(adapter_file):
         raise FileNotFoundError(f"The adapter file does not exist: {adapter_file}")
 
-    adapters = list(mx.load(adapter_file).items())
+    weights, metadata = mx.load(adapter_file, return_metadata=True)
+    if not metadata:
+        warnings.warn(
+            "The metadata for the adapter is missing. Default values will be used, which may result in incorrect LoRA parameters."
+        )
+        metadata = {}
+
+    adapters = list(weights.items())
 
     linear_replacements = []
     lora_layers = set(
@@ -91,7 +100,11 @@ def apply_lora_layers(model: nn.Module, adapter_file: str) -> nn.Module:
     )
     for name, module in model.named_modules():
         if name in lora_layers:
-            replacement_module = LoRALinear.from_linear(module)
+            replacement_module = LoRALinear.from_linear(
+                module,
+                **json.loads(metadata.get("lora_config", {})),
+            )
+            print(replacement_module.lora_config)
             linear_replacements.append((name, replacement_module))
 
     model.update_modules(tree_unflatten(linear_replacements))
--- a/llms/mlx_lm/tuner/utils.py
+++ b/llms/mlx_lm/tuner/utils.py
@@ -85,11 +85,18 @@ def apply_lora_layers(model: nn.Module, adapter_file: str) -> nn.Module:
     if not os.path.exists(adapter_file):
         raise FileNotFoundError(f"The adapter file does not exist: {adapter_file}")
 
-    weights, metadata = mx.load(adapter_file, return_metadata=True)
-    if not metadata:
+    try:
+        weights, metadata = mx.load(adapter_file, return_metadata=True)
+        if not metadata:
+            warnings.warn(
+                "The metadata for the adapter is missing. Default values will be used, which may result in incorrect LoRA parameters."
+            )
+            metadata = {}
+    except ValueError:
         warnings.warn(
-            "The metadata for the adapter is missing. Default values will be used, which may result in incorrect LoRA parameters."
+            "The metadata not supported for format npz. Default values will be used, which may result in incorrect LoRA parameters."
         )
+        weights = mx.load(adapter_file)
         metadata = {}
 
     adapters = list(weights.items())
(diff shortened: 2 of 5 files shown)


llms/mlx_lm/tuner/lora.py
```diff
@@ -85,7 +85,7 @@ def __init__(
         self.dropout = nn.Dropout(p=dropout)
 
         # Scale for low-rank update
-        self.scale = scale * (alpha / r)
+        self.scale = mx.array(scale * (alpha / r))
```
mzbac, line 88: maybe we need to add ` mx.stop_gradient(self.scale)`

llms/tests/test_lora.py
```diff
@@ -45,7 +45,8 @@ def check_config(params):
                 v.size for _, v in tree_flatten(model.trainable_parameters())
             )
             self.assertEqual(
-                trainable_params, lora_layers * params["rank"] * 1024 * 2 * n_keys
+                trainable_params,
+                lora_layers * params["rank"] * 1024 * 2 * n_keys + lora_layers * n_keys,
```
mzbac, line 49: I couldn't find a way to make the scale a non-trainable constant parameter. Ideally, it shouldn't be part of trainable parameters since it's a constant parameter.

llms/mlx_lm/tuner/lora.py
awni:
```
self.freeze(keys=["scale"], recurse=False)
```

awni: Also I find it a bit odd that we have an `alpha` and a `scale` parameter since they are both just multiplicative scales 🤔 . I guess that was the original LoRA parameterization but maybe it is worth simplifying it to a single scale parameter.

awni:
Good catch! The `freeze` will work perfectly for this. Let's make it a PR and merge it!

awni (approved):
🙏 thanks!

awni:
Closes #622

llms/mlx_lm/tuner/lora.py
mzbac: Yeah, the original LoRa paper does not have the scale parameter (https://github.com/microsoft/LoRA/blob/main/loralib/layers.py#L112), but either way, we need to store the final scale parameter in the weight so that it can be retrieved from the adapter. Alternatively, we could save the LoRa configuration along with the weight.

awni: I think the way you've done it is fine in terms of making it an `mx.array`.

I would like to merge `alpha` and `scale` into a single parameter to simplify the number of inputs. But that can be done in a follow up.

mzbac:
@awni I think there is an issue with the implementation, so please don't merge it for now. The scale is frozen, so it won't be included in the weight calculations for adapters.

awni:
E.g. change L307 to:

```
    flattened_tree = tree_flatten(model.parameters())
```

awni:
Oh yes.. that's a bad idea nvm

llms/mlx_lm/tuner/lora.py
chimezie: I was actually looking at that earlier in comparison to the LoRA paper and when considering which of the two to include in hyperparameter sweeps.  Merging them would definitely simplify things

mzbac:
@awni one solution that comes to mind is to save the adapter weight in SafeTensor format and include the LoRa parameter in the metadata. This way, when we try to load the adapter weight, we can retrieve those parameters using metadata. this way we only need to change two places:
https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/tuner/trainer.py#L309
https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/tuner/utils.py#L72

llms/mlx_lm/tuner/lora.py
awni: >  To address this issue, we added the scale parameter. I am not sure why using only r and alpha caused issues in fusing models,

We needed a way to increase the size of the adapters weights so they didn't get erased in fusion, which is why I added the [`scale`](https://github.com/ml-explore/mlx-examples/pull/252/files/8b84afa63900cf3697776e97f20d1b98b055398b#r1445442551). At the time there was no `alpha` parameter, we just fixed it at `2` (which is what the LoRA paper does IIRC). In hindsight I should have just named `scale` `alpha`. Somewhere along the way we added added an additional `alpha` parameter. I don't think it makes sense to include both, the `scale` is really the same as the `alpha`.

awni: Its quite reasonable to have quite different `alpha` parameters. Keeping only `alpha` and `r` will be more consistent with what other frameworks do IMO. The `scale` is the oddball here.

awni: Btw looking at [that thread](https://github.com/ml-explore/mlx-examples/pull/252/files/8b84afa63900cf3697776e97f20d1b98b055398b#r1445442551), maybe we should experiment with other initializations. Even with a large scale, we still see issues with fusion from time to time.

mzbac:
@awni, the fix is pretty much done. Let me know what you think.

llms/mlx_lm/tuner/utils.py
```diff
@@ -83,15 +84,33 @@ def apply_lora_layers(model: nn.Module, adapter_file: str) -> nn.Module:
     if not os.path.exists(adapter_file):
         raise FileNotFoundError(f"The adapter file does not exist: {adapter_file}")
 
-    adapters = list(mx.load(adapter_file).items())
+    weights, metadata = mx.load(adapter_file, return_metadata=True)
```
mzbac, line 87: I'm not very sure if loading the np file with `return_metadata=True` will throw an error or just return None for the metadata.

awni:
To be honest using safetensors and metadata is more involved than I was hoping it would be. It kind of couples stuff (e.g. training, saving, etc) in a way that makes it harder to change..

I wonder if there is a cleaner solution 🤔

awni:
It might make more sense to simply save the config file that we use to train the model in the initial setup and then read from it when reconstructing the lora layers. It introduces an extra artifact.. but it could also be useful to retain that artifact and it should be quite a bit more general..

mzbac:
> It might make more sense to simply save the adapter config that we use to train the model in the initial setup and then read from it when reconstructing the lora layers. It introduces an extra artifact.. but it could also be useful to retain that artifact and it should be quite a bit more general..

Saving the configuration could potentially introduce a breaking change, as we need to change from  the adapter file to the adapter folder and won't still be able to decouple things in training or saving. Just curious, is there any chance we can support different data types in mlx safetensors metadata?

awni:
> Just curious, is there any chance we can support different data types in mlx safetensors metadata?

The safetensors spec is not MLX specific https://github.com/huggingface/safetensors?tab=readme-ov-file#format

As for breaking change, indeed, though we could probably mitigate it with some conditional loading.

mzbac:
I filed an issue in the safeensors repo to see if there is an easy way to extend the datatype in metadata. Otherwise, separating the configuration may be something we have to do. However, I am leaning towards the gguf way of having everything in one file, which is easier to manage.

awni:
We are already quite split among different files (base model, tokenizer and all the corresponding files, config, adapters.npz). I think its ok to have all the adapter related data in a single directory (checkpoints, final adapters, and config).  I'm not sure how that would look / if it would be cleaner but I don't mind investigating it, if you're not interested in looking into it.

Another option is to go back to making the scale a parameter and then trying to pick it up in the adapters dictionary.

mzbac:
I am not opposed to having a separate configuration for the LoRa parameters. The HuggingFace's `peft` library already uses the same approach, so users should be familiar with that method. I'm just trying to avoid any potential breaking changes if possible. Anyway, I will have a look at how to introduce the separate LoRa config today and see if that route would be cleaner.

ryanmitts:
The safetensors format for the adapters might be nice. I've been looking at loading PEFT adapter safetensor files into MLX, and being able to load conditionally on the `metadata.format` field would be nice.

mzbac:
Before proceeding to the separate configuration, @awni what do you think about serializing the LoRa parameters as a JSON string and saving it in metadata, then deserializing when loading the LoRa weight? This way, we can avoid the type conversion.

mzbac:
@awni, I have made the changes to save the separate configuration. However, it feels more complicated and introduces breaking changes for all lora, generate, fuse, and server scripts than the metadata approach. Here is a draft of the changes. Let me know if we still want to go with separate configurations.
FYI: https://github.com/ml-explore/mlx-examples/compare/main...mzbac:mlx-lm/lora/scale-2?expand=1

mzbac:
@awni, I also updated the PR to save the LoRa config as a JSON string which simplifies the implementation a bit. Would you please review this and see if the changes make the solution cleaner?

llms/mlx_lm/tuner/lora.py
ivanfioravanti: +1 for removing scale. alpha + r is better.

mzbac: The only concern I have is that using alpha normally creates a scale value around 2 or 3, but in the current alpha+ scale we are getting around 20 to 30 scaling. I'm not sure if that will make a significant difference in output or training.

ryanmitts: I actually ran into this yesterday I have a LoRA adapter trained with a scale of 1, but when running a benchark, the results were worse than the base model. When I found the problem, I forgot to set the scale back to 1 from the default when running the model for the benchmark.

awni:
@mzbac for the delay reviewing this. I think what you've implemented is actually pretty clean.

My only hesitation is that having a properly saved config might actually be useful (e.g. dumping the entire experiment config might be valuable for tracking experiments etc). If we are going to go that route it might be better to do it rather than change twice. I wanted think about this a little and also look and see what it's like to save a config / what our options are before committing here.

mzbac:
> @mzbac for the delay reviewing this. I think what you've implemented is actually really clean! It's a nice solution.
> 
> My only hesitation is that having a properly saved config might actually be useful (e.g. dumping the entire experiment config might be valuable for tracking experiments etc). If we are going to go that route it might be better to do it rather than change twice. I wanted think about this a little and also look and see what it's like to save a config / what our options are before committing here.

Yeah, we don't have to rush for this. By the way, if we want to save all the experiment configurations, we can also save them in metadata since we can serialize all the experiment parameters into a JSON string. However, if we prefer an explicit config file to make it more accessible, I think it makes sense as well.

awni:
I think we can close this and use #636?

mzbac:
Happy to close it as we fixed the issue via https://github.com/ml-explore/mlx-examples/pull/636.

PR #628, opened 2024-03-27
DBRX

awni:
## Quantize

```shell
python -m mlx_lm.convert --hf-path databricks/dbrx-base -q 
```

## Generate

```shell
python -m mlx_lm.generate --model mlx_model --prompt "Write a quicksort in Python" --trust-remote-code --max-tokens 500
```

Sample output:
```
- The partition operation will use a pair of indices that start at the two ends of the array being partitioned, then move toward each other, until they detect an inversion: a pair of elements, one larger than the pivot and one smaller, that are in the wrong order relative to each other. The inverted elements are then swapped. When
==========
Prompt: 3.415 tokens-per-sec
Generation: 14.071 tokens-per-sec
```

## QLoRA

```shell
 python -m mlx_lm.lora --model mlx_model --data ../lora/data --batch-size 1 --train --iters 200 --lora-layers 8 --steps-per-report 10 --trust-remote-code
```

Log:
```text
Iter 100: Saved adapter weights to checkpoints/100_adapters.npz.
Iter 110: Train loss 2.264, Learning Rate 1.000e-05, It/sec 0.341, Tokens/sec 26.124, Trained Tokens 8757, Peak mem 71.735 GB
Iter 120: Train loss 2.190, Learning Rate 1.000e-05, It/sec 0.340, Tokens/sec 25.812, Trained Tokens 9516, Peak mem 71.735 GB
Iter 130: Train loss 2.105, Learning Rate 1.000e-05, It/sec 0.336, Tokens/sec 27.973, Trained Tokens 10349, Peak mem 71.735 GB
Iter 140: Train loss 2.120, Learning Rate 1.000e-05, It/sec 0.334, Tokens/sec 26.627, Trained Tokens 11147, Peak mem 71.735 GB
Iter 150: Train loss 1.829, Learning Rate 1.000e-05, It/sec 0.340, Tokens/sec 23.525, Trained Tokens 11838, Peak mem 71.735 GB
```

### DIFF
--- a/llms/mlx_lm/models/dbrx.py
+++ b/llms/mlx_lm/models/dbrx.py
@@ -141,15 +141,12 @@ class SparseMoeBlock(nn.Module):
         x = x.reshape(-1, x.shape[-1])
 
         gates = self.router(x)
+        gates = mx.softmax(gates.astype(mx.float32), axis=-1)
 
-        inds = mx.stop_gradient(
-            mx.argpartition(-gates, kth=ne, axis=-1)[:, :ne]
-        )  # TODO remove it once we figure out how to fine tune TopK in MOE
-
-        scores = mx.softmax(
-            mx.take_along_axis(gates, inds, axis=-1).astype(mx.float32),
-            axis=-1,
-        ).astype(gates.dtype)
+        inds = mx.stop_gradient(mx.argpartition(-gates, kth=ne, axis=-1)[:, :ne])
+        scores = mx.take_along_axis(gates, inds, axis=-1)
+        scores = scores / mx.linalg.norm(scores, ord=1, axis=-1, keepdims=True)
+        scores = scores.astype(x.dtype)
 
         if self.training:
             inds = np.array(inds)
@@ -166,8 +163,8 @@ class SparseMoeBlock(nn.Module):
             for xt, st, it in zip(x, scores, inds.tolist()):
                 yt = mx.stack([self.experts[e](xt) for e in it], axis=-1)
                 yt = (yt * st).sum(axis=-1)
-                y.append(yt[None, :])
-            y = mx.concatenate(y)
+                y.append(yt)
+            y = mx.stack(y, axis=0)
 
         return y.reshape(orig_shape)
 
--- a/llms/mlx_lm/models/dbrx.py
+++ b/llms/mlx_lm/models/dbrx.py
@@ -1,0 +1,261 @@
+from dataclasses import dataclass
+from typing import Dict, Optional, Tuple, Union
+
+import mlx.core as mx
+import mlx.nn as nn
+import numpy as np
+
+from .base import BaseModelArgs
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    model_type: str
+    vocab_size: int
+    d_model: int
+    ffn_config: dict
+    attn_config: dict
+    n_layers: int
+    n_heads: int
+
+
+class Attention(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+        self.num_heads = args.n_heads
+        self.d_model = args.d_model
+        self.head_dim = args.d_model // args.n_heads
+        self.num_key_value_heads = args.attn_config["kv_n_heads"]
+        self.clip_qkv = args.attn_config["clip_qkv"]
+        self.rope_theta = args.attn_config["rope_theta"]
+
+        self.scale = self.head_dim**-0.5
+
+        self.Wqkv = nn.Linear(args.d_model, (self.num_key_value_heads * 2 + self.num_heads) * self.head_dim, bias=False)
+        self.out_proj = nn.Linear(args.d_model, args.d_model, bias=False)
+        self.rope = nn.RoPE(
+            self.head_dim,
+            traditional=False,
+            base=self.rope_theta,
+        )
+
+    def __call__(
+        self,
+        x: mx.array,
+        mask: Optional[mx.array] = None,
+        cache: Optional[Tuple[mx.array, mx.array]] = None,
+    ) -> mx.array:
+
+        qkv = self.Wqkv(x)
+        qkv = mx.clip(qkv, a_min=-self.clip_qkv, a_max=self.clip_qkv)
+        splits = [self.d_model, self.d_model + self.head_dim * self.num_key_value_heads]
+        queries, keys, values = mx.split(qkv, splits, axis=-1)
+
+        B, L, D = x.shape
+
+        # Prepare the queries, keys and values for the attention computation
+        queries = queries.reshape(B, L, self.num_heads, -1).transpose(0, 2, 1, 3)
+        keys = keys.reshape(B, L, self.num_key_value_heads, -1).transpose(0, 2, 1, 3)
+        values = values.reshape(B, L, self.num_key_value_heads, -1).transpose(
+            0, 2, 1, 3
+        )
+
+        if cache is not None:
+            key_cache, value_cache = cache
+            queries = self.rope(queries, offset=key_cache.shape[2])
+            keys = self.rope(keys, offset=key_cache.shape[2])
+            keys = mx.concatenate([key_cache, keys], axis=2)
+            values = mx.concatenate([value_cache, values], axis=2)
+        else:
+            queries = self.rope(queries)
+            keys = self.rope(keys)
+
+        output = mx.fast.scaled_dot_product_attention(
+            queries, keys, values, scale=self.scale, mask=mask
+        )
+        output = output.transpose(0, 2, 1, 3).reshape(B, L, -1)
+        return self.out_proj(output), (keys, values)
+
+
+class NormAttnNorm(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+        self.norm_1 = nn.LayerNorm(args.d_model, bias=False)
+        self.norm_2 = nn.LayerNorm(args.d_model, bias=False)
+        self.attn = Attention(args)
+
+    def __call__(self, x: mx.array,
+        mask: Optional[mx.array] = None,
+        cache: Optional[Tuple[mx.array, mx.array]] = None,
+    ) -> mx.array:
+        h, cache = self.attn(self.norm_1(x), mask=mask, cache=cache)
+        x = h + x
+        return x, self.norm_2(x), cache
+
+
+class MLP(nn.Module):
+    def __init__(self, d_model: int, ffn_dim: int):
+        super().__init__()
+        self.v1 = nn.Linear(d_model, ffn_dim, bias=False)
+        self.w1 = nn.Linear(d_model, ffn_dim, bias=False)
+        self.w2 = nn.Linear(ffn_dim, d_model, bias=False)
+        self.act_fn = nn.silu
+
+    def __call__(self, x: mx.array) -> mx.array:
+        current_hidden_states = self.act_fn(self.w1(x)) * self.v1(x)
+        current_hidden_states = self.w2(current_hidden_states)
+        return current_hidden_states
+
+
+class Router(nn.Module):
+    def __init__(self, d_model: int, num_experts: int):
+        super().__init__()
+        self.layer = nn.Linear(d_model, num_experts, bias=False)
+
+    def __call__(self, x: mx.array):
+        return self.layer(x)
+
+
+class SparseMoeBlock(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+        self.d_model = args.d_model
+        self.ffn_dim = args.ffn_config["ffn_hidden_size"]
+        self.num_experts = args.ffn_config["moe_num_experts"]
+        self.num_experts_per_tok = args.ffn_config["moe_top_k"]
+
+        self.router = Router(self.d_model, self.num_experts)
+        self.experts = [MLP(self.d_model, self.ffn_dim) for _ in range(self.num_experts)]
+
+    def __call__(self, x: mx.array) -> mx.array:
+        ne = self.num_experts_per_tok
+        orig_shape = x.shape
+        x = x.reshape(-1, x.shape[-1])
+
+        gates = self.router(x)
+
+        inds = mx.stop_gradient(
+            mx.argpartition(-gates, kth=ne, axis=-1)[:, :ne]
+        )  # TODO remove it once we figure out how to fine tune TopK in MOE
+
+        scores = mx.softmax(
+            mx.take_along_axis(gates, inds, axis=-1).astype(mx.float32),
+            axis=-1,
+        ).astype(gates.dtype)
+
+        if self.training:
+            mx.eval(inds)
+            inds = np.array(inds)
+            y = mx.zeros((x.shape[0], ne, x.shape[-1]), x.dtype)
+            for e, expert in enumerate(self.experts):
+                idx1, idx2 = map(mx.array, np.where(inds == e))
+                if idx1.size == 0:
+                    continue
+                y[idx1, idx2] = expert(x[idx1])
+
+            y = (y * scores[:, :, None]).sum(axis=1)
+        else:
+            y = []
+            for xt, st, it in zip(x, scores, inds.tolist()):
+                yt = mx.concatenate([self.experts[e](xt)[:, None] for e in it], axis=-1)
+                yt = (yt * st).sum(axis=-1)
+                y.append(yt[None, :])
+            y = mx.concatenate(y)
+
+        return y.reshape(orig_shape)
+
+
+class DecoderLayer(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+        self.ffn = SparseMoeBlock(args)
+        self.norm_attn_norm = NormAttnNorm(args)
+
+    def __call__(
+        self,
+        x: mx.array,
+        mask: Optional[mx.array] = None,
+        cache: Optional[Tuple[mx.array, mx.array]] = None,
+    ) -> mx.array:
+        r, h, cache = self.norm_attn_norm(x, mask, cache)
+        out = self.ffn(h) + r
+        return out, cache
+
+
+class DBRX(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+        self.vocab_size = args.vocab_size
+        self.wte = nn.Embedding(args.vocab_size, args.d_model)
+        self.blocks = [DecoderLayer(args=args) for _ in range(args.n_layers)]
+        self.norm_f = nn.LayerNorm(args.d_model, bias=False)
+
+    def __call__(
+        self,
+        inputs: mx.array,
+        cache=None,
+    ):
+        h = self.wte(inputs)
+
+        mask = None
+        T = h.shape[1]
+        if T > 1:
+            mask = nn.MultiHeadAttention.create_additive_causal_mask(T)
+            mask = mask.astype(h.dtype)
+
+        if cache is None:
+            cache = [None] * len(self.blocks)
+
+        for e, layer in enumerate(self.blocks):
+            import pdb
+            h, cache[e] = layer(h, mask, cache[e])
+
+        return self.norm_f(h), cache
+
+
+class Model(nn.Module):
+    def __init__(self, args: ModelArgs):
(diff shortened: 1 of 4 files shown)


llms/mlx_lm/models/dbrx.py
```diff
@@ -0,0 +1,268 @@
...
+            mx.argpartition(-gates, kth=ne, axis=-1)[:, :ne]
+        )  # TODO remove it once we figure out how to fine tune TopK in MOE
+
+        scores = mx.softmax(
+            mx.take_along_axis(gates, inds, axis=-1).astype(mx.float32),
+            axis=-1,
+        ).astype(gates.dtype)
+
+        if self.training:
+            mx.eval(inds)
```
mustafaaljadery, line 155: Because you convert inds to np array right after, is the mx.eval necessary?

llms/mlx_lm/models/dbrx.py
```diff
@@ -0,0 +1,268 @@
...
+            inds = np.array(inds)
+            y = mx.zeros((x.shape[0], ne, x.shape[-1]), x.dtype)
+            for e, expert in enumerate(self.experts):
+                idx1, idx2 = map(mx.array, np.where(inds == e))
```
mustafaaljadery, line 159: This may be simpler:
```python           
idx = np.where(np.array(inds) == e)
if len(idx[0]) > 0:
     y[idx[0], idx[1]] = expert(x[idx[0]])
```

llms/mlx_lm/models/dbrx.py
```diff
@@ -0,0 +1,268 @@
...
+                if idx1.size == 0:
+                    continue
+                y[idx1, idx2] = expert(x[idx1])
+
+            y = (y * scores[:, :, None]).sum(axis=1)
+        else:
+            y = []
+            for xt, st, it in zip(x, scores, inds.tolist()):
+                yt = mx.concatenate([self.experts[e](xt)[:, None] for e in it], axis=-1)
```
mustafaaljadery, line 168: `[:, None]` is added for a dimensionality, but is that not auto handled by mx.concatenate.

llms/mlx_lm/models/dbrx.py
```diff
@@ -0,0 +1,268 @@
...
+        cache=None,
+    ):
+        out, cache = self.transformer(inputs, cache)
+        return self.lm_head(out), cache
+
+    @property
+    def layers(self):
+        return self.transformer.blocks
+
+    def sanitize(self, weights):
```
mustafaaljadery, line 245: The split expert can possibly be simplified with np.split(). This is simpler for splitting the tensors. Something like this, not 100% sure: 
```python
def sanitize(self, weights):
    num_experts = self.args.ffn_config["moe_num_experts"]
    dim = self.args.ffn_config["ffn_hidden_size"]

    new_weights = {k: v for k, v in weights.items() if "experts.mlp" not in k}
    expert_weights = []

    for k, v in weights.items():
        if "experts.mlp" in k:
            expert_splits = np.split(v, num_experts, axis=0)
            for e in range(num_experts):
                new_key = k.replace(".mlp.", f".{e}.weight")
                expert_weight = expert_splits[e]
                if k.endswith(".w2"):
                    expert_weight = expert_weight.T
                expert_weights.append((new_key, expert_weight))

    new_weights.update(expert_weights)

    return new_weights
```

mustafaaljadery (commented):
This is my first time contributing, this might be the wrong practice, very sorry if it is!

llms/mlx_lm/models/dbrx.py
awni: Great point. It's also in some of our other MOE code.. but it's superfluous!

llms/mlx_lm/models/dbrx.py
awni: You cannot index `mx.array` with `np.ndarray`, maybe should add that 🤔

llms/mlx_lm/models/dbrx.py
awni: Nice, I like the idea of using split! I will borrow this!

llms/mlx_lm/models/dbrx.py
awni: Actually `concatenate` does not insert a new dimensions. But that's still a good point, we should use `stack` here which does.. We added this code before we had `stack` :).

mzbac:
Wondering what the memory requirement is for converting the model? I thought even 192GB wouldn't be enough.

awni:
@mustafaaljadery thank you for the comments, those were nice!

my-other-github-account:
This is awesome, works great!

Looks like it must be running Q4, I'm able to run it on an M3 96GB VRAM with the above prompt, doesn't drop under 15GB free even with some other stuff running on the system.

Makes sense because rough estimate 2x param count in B gives you FP16 GB RAM reqs, so 132B*2 = 264GB if running in FP16, but at Q4 you cut that in 1/4th and it can run in approx 66GB RAM.

Great work!

lin72h:
I see the the Mac Studio M3 Ultra is be perfect for this f16 finetune task

awni:
>  Wondering what the memory requirement is for converting the model? I thought even 192GB wouldn't be enough.

You shouldn't need that much RAM to quantize it. I'm trying it on an 8GB machine (I think it will work but be very slow). Definitely 32GB is plenty though.

mzbac:
> > Wondering what the memory requirement is for converting the model? I thought even 192GB wouldn't be enough.
> 
> You shouldn't need that much RAM to quantize it. I'm trying it on an 8GB machine (I think it will work but be very slow). Definitely 32GB is plenty though.

That's awesome! 🚀 🚀

my-other-github-account:
Hello,

I think there might be an issue with the generation here, or perhaps I've done something wrong.

So I go through and do:

python -m mlx_lm.convert --hf-path databricks/dbrx-instruct -q 

Then:

```

model, tokenizer = load("mlx_model", trust_remote_code=True)
```

Using the following prompt: (That I get from tokenizer.apply_chat_template() applied to the first problem from HumanEval)


```
<|im_start|>system
You are DBRX, created by Databricks. You were last updated in December 2023. You answer questions based on information available up to that point.
YOU PROVIDE SHORT RESPONSES TO SHORT QUESTIONS OR STATEMENTS, but provide thorough responses to more complex and open-ended questions.
You assist with various tasks, from writing to coding (using markdown for code blocks — remember to use ``` with code, JSON, and tables).
(You do not have real-time data access or code execution capabilities. You avoid stereotyping and provide balanced perspectives on controversial topics. You do not provide song lyrics, poems, or news articles and do not divulge details of your training data.)
This is your system prompt, guiding your responses. Do not reference it, just respond to the user. If you find yourself talking about this message, stop. You should be responding appropriately and usually that means not mentioning this.
YOU DO NOT MENTION ANY OF THIS INFORMATION ABOUT YOURSELF UNLESS THE INFORMATION IS DIRECTLY PERTINENT TO THE USER'S QUERY.<|im_end|>
<|im_start|>user
Complete the following code:
from typing import List


def has_close_elements(numbers: List[float], threshold: float) -> bool:
    """ Check if in given list of numbers, are any two numbers closer to each other than
    given threshold.
    >>> has_close_elements([1.0, 2.0, 3.0], 0.5)
    False
    >>> has_close_elements([1.0, 2.8, 3.0, 4.0, 5.0, 2.0], 0.3)
    True
    """
<|im_end|>
<|im_start|>assistant
from typing import List


def has_close_elements(numbers: List[float], threshold: float) -> bool:
    """ Check if in given list of numbers, are any two numbers closer to each other than
    given threshold.
    >>> has_close_elements([1.0, 2.0, 3.0], 0.5)
    False
    >>> has_close_elements([1.0, 2.8, 3.0, 4.0, 5.0, 2.0], 0.3)
    True
    """
```

I run the mlx version as follows:

generate(model, tokenizer, prompt=prompt, verbose=True, max_tokens=5000)

But get odd output that looks like it missed a lot of the prompt, or got otherwise corrupted: 

`itherwise, you would be able to provide information that is not available in your training data. However, you are not able to access the internet, other systems, or any data outside of your training data. You are also not able to perform any actions or tasks, you can only provide information and answer questions. You are also not able to form memories or remember any interactions, you can only provide information and answer questions based on your training data. You are also not able to learn or improve over time, you can only provide information and answer questions based on your training data. You are also not able to form opinions or make judgments, you can only provide information and answer questions. You are also not able to have personal experiences or emotions, you can only provide information and answer questions. You are also not able to have a sense of self or consciousness, you can only provide information and answer questions. You are also not able to have a sense of time or know the current date or time, you can only provide information and answer questions based on your training data. You are also not able to have a sense of location or know where you are, you can only provide information and answer questions based on your training data. [...SNIP...]`

The same prompt with HF works fine and gives:

```
def has_close_elements(numbers: List[float], threshold: float) -> bool:
    """ Check if in given list of numbers, are any two numbers closer to each other than
    given threshold.
    >>> has_close_elements([1.0, 2.0, 3.0], 0.5)
    False
    >>> has_close_elements([1.0, 2.8, 3.0, 4.0, 5.0, 2.0], 0.3)
    True
    """
    numbers.sort() # sort the list of numbers
    for i in range(len(numbers) - 1):
        if numbers[i+1] - numbers[i] < threshold:
            return True
    return False 
```
This holds true for every problem in HumanEval - good results from HF, unintelligible outputs from MLX Q4.

Any ideas where I might have gone wrong, or if there is another issue that could be causing this?

Thank you!

mezmon2:
> Hello,
> 
> I think there might be an issue with the generation here, or perhaps I've done something wrong.
> 
> So I go through and do:
> 
> python -m mlx_lm.convert --hf-path databricks/dbrx-instruct -q
> 

Did you have any bias related errors when running the inital command w/ the instruct model? I am getting this 

`self.norm_1` = nn.LayerNorm(args.d_model, bias=False)

^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

TypeError: LayerNorm.__init__() got an unexpected keyword argument 'bias'`

M-I:
@my-other-github-account Got similar output when trying to apply the chat template.

In case it helps, here's some output with no template:

 
response = generate(model, tokenizer, prompt="Here's the code in markdown for a quicksort in Python:", verbose=True, max_tokens=200)

==========
> 
> Prompt: Here's the code in markdown for a quicksort in Python:
> 
> 
> ```python
> def quicksort(arr):
>     if len(arr) <= 1:
>         return arr
>     pivot = arr[len(arr) // 2]
>     left = [x for x in arr if x < pivot]
>     middle = [x for x in arr if x == pivot]
>     right = [x for x in arr if x > pivot]
>     return quicksort(left) + middle + quicksort(right)
> 
> print(quicksort([3,6,8,10,1,2,1]))
> # Returns [1, 1, 2, 3, 6, 8, 10]
> ```
> 
> And here's the same code in a code block:
> 
> ```python
> def quicksort(arr):
>     if len(arr) <= 1:
>         return arr
>     pivot = arr[len(arr) // 2]
>     left = [x for x in arr if x < pivot]
>     middle = [x for x in arr if x
> 
==========
Prompt: 4.942 tokens-per-sec
Generation: 4.681 tokens-per-sec

M-I:
> > Hello,
> > I think there might be an issue with the generation here, or perhaps I've done something wrong.
> > So I go through and do:
> > python -m mlx_lm.convert --hf-path databricks/dbrx-instruct -q
> 
> Did you have any bias related errors when running the inital command w/ the instruct model? I am getting this
> 
> `self.norm_1` = nn.LayerNorm(args.d_model, bias=False)
> 
> ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 
> TypeError: LayerNorm.**init**() got an unexpected keyword argument 'bias'`

Just in case, make sure you're on the right branch (dbrx), and that your env doesn't have mlx-lm instaled from pip.

awni:
@mezmon2 use the latest MLX (`python -m pip install -U mlx`).

ivanfioravanti:
Great PR!

my-other-github-account:
> @my-other-github-account Got similar output when trying to apply the chat template.
> 
> In case it helps, here's some output with no template:

Not using the template sort of defeats the purpose of an instruct tuned model unfortunately, as it won't work as a traditional chatbot. At that point you are better off with the base model :/

Given that most people are using LLMs as chatbots, seems like a pretty serious issue for normal use!

Is there another way we should be applying the template so it works properly and can be customized for IFT models in a chat setting?

awni:
@mzbac @Blaizzy the instruct model here uses the default template but `tokenizer.chat_template==None`

M-I:
> Not using the template sort of defeats the purpose of an instruct tuned model unfortunately, as it won't work as a traditional chatbot. So you can't say e.g. "please give me the code for quicksort in Python" or speak to it naturally. At that point you are better off with the base model :/

Sorry for the confusion, it wasn't a suggestion to try without template. Since you had it working with HF, and this was showing reasonable text generation with mlx.  it was to indicate the issue was probably not the quantization process but maybe in the code that applies the template.

awni:
Indeed the instruct model doesn't respond well to the template.. I wonder if there is a bug in the template.. the text itself looks slightly reasonable.. unfortunately it's difficult to check if it's related to the quantization given the fp16 model doesn't in in RAM.

my-other-github-account:
In case it helps, here are my two quick reproducers:

HF:
```
from transformers import AutoTokenizer, AutoModelForCausalLM
import torch
from evalplus.data import get_human_eval_plus

tokenizer = AutoTokenizer.from_pretrained("databricks/dbrx-instruct", trust_remote_code=True)
model = AutoModelForCausalLM.from_pretrained("databricks/dbrx-instruct", device_map="auto", torch_dtype=torch.bfloat16, trust_remote_code=True)

input_prompt = list(get_human_eval_plus().items())[0][1]["prompt"]
entry_point = list(get_human_eval_plus().items())[0][1]["entry_point"]

def get_prompt(input_prompt):
    input_text = "Complete the following code:\n" + input_prompt
    messages = [{"role": "user", "content": input_text}, {"role": "assistant", "content": input_prompt}]
    prompt = tokenizer.apply_chat_template(messages, return_dict=True, tokenize=False, add_generation_prompt=False, return_tensors="pt")
    prompt = prompt[:-10]
    return prompt

prompt = get_prompt(input_prompt)
outputs = model.generate(tokenizer.encode(prompt, return_tensors="pt"), max_new_tokens=5000)
cleaned = "def "+entry_point+entry_point.join((tokenizer.decode(outputs[0]).split("<|im_start|> assistant")[1].split(entry_point)[1:]))
print(cleaned.split("<|im_end|>")[0])
```

MLX:

```
from evalplus.data import get_human_eval_plus

model, tokenizer = load("mlx_model")

input_prompt = list(get_human_eval_plus().items())[0][1]["prompt"]
entry_point = list(get_human_eval_plus().items())[0][1]["entry_point"]

def get_prompt(input_prompt):
    input_text = "Complete the following code:\n" + input_prompt
    messages = [{"role": "user", "content": input_text}, {"role": "assistant", "content": input_prompt}]
    prompt = tokenizer.apply_chat_template(messages, return_dict=True, tokenize=False, add_generation_prompt=False, return_tensors="pt")
    prompt = prompt[:-10]
    return prompt

prompt = get_prompt(input_prompt)
result = generate(model, tokenizer, prompt=prompt, verbose=True, max_tokens=5000)
print(result)
```

awni:
I'm curious, how did you run the HF model? Is it on a single GPU or sharded?

Blaizzy:
@my-other-github-account  could you share what is output of is "get_prompt(input_prompt)" on MLX and HF.

awni:
I checked the prompts @Blaizzy they are the same indeed, both of them give:

```
<|im_start|>system
You are DBRX, created by Databricks. You were last updated in December 2023. You answer questions based on information available up to that point.
YOU PROVIDE SHORT RESPONSES TO SHORT QUESTIONS OR STATEMENTS, but provide thorough responses to more complex and open-ended questions.
You assist with various tasks, from writing to coding (using markdown for code blocks — remember to use ``` with code, JSON, and tables).
(You do not have real-time data access or code execution capabilities. You avoid stereotyping and provide balanced perspectives on controversial topics. You do not provide song lyrics, poems, or news articles and do not divulge details of your training data.)
This is your system prompt, guiding your responses. Do not reference it, just respond to the user. If you find yourself talking about this message, stop. You should be responding appropriately and usually that means not mentioning this.
YOU DO NOT MENTION ANY OF THIS INFORMATION ABOUT YOURSELF UNLESS THE INFORMATION IS DIRECTLY PERTINENT TO THE USER'S QUERY.<|im_end|>
<|im_start|>user
Complete the following code:
from typing import List


def has_close_elements(numbers: List[float], threshold: float) -> bool:
    """ Check if in given list of numbers, are any two numbers closer to each other than
    given threshold.
    >>> has_close_elements([1.0, 2.0, 3.0], 0.5)
    False
    >>> has_close_elements([1.0, 2.8, 3.0, 4.0, 5.0, 2.0], 0.3)
    True
    """
<|im_end|>
<|im_start|>assistant
from typing import List


def has_close_elements(numbers: List[float], threshold: float) -> bool:
    """ Check if in given list of numbers, are any two numbers closer to each other than
    given threshold.
    >>> has_close_elements([1.0, 2.0, 3.0], 0.5)
    False
    >>> has_close_elements([1.0, 2.8, 3.0, 4.0, 5.0, 2.0], 0.3)
    True
    """
```

Blaizzy:
What is wrong then? is it something with the model?

awni:
> What is wrong then? is it something with the model?

Not sure.. could be quantization, could be something else.

Blaizzy:
Could you please describe the issue?

awni:
The issue is that the model is not generating good outputs when given the chat template. Also we think we have dialed in on a bug with the quantized gate matrices.. should have an update on that soon.

Blaizzy:
Make sense, judging by the fact that Transformers version works, it could it. 

Please tag me when you update the gates matrices 👌🏽

my-other-github-account:
I have been running it sharded on 4xA100 80GB - and indeed the above are the prompts I get with either.

I could 8 bit quant it, but llama.cpp doesn't support less than 8 bit quantization for DBRX yet as it is slightly customized. Right now I am in FP16. Would it help to try 8?

For what its worth, I have benchmarked somewhere in the neighborhood of ~50 LLMs against humaneval, in none of the cases did dropping to 4 bit affect performance substantially (3 bit did a bit, though, but not to the point of being unusable). So FWIW quantization error alone absent a bug isn't sufficient to explain such drastic results IME - i.e. dropping to a 0 on humaneval would be unheard of, which is what happens here.

awni:
Hotfix here https://github.com/ml-explore/mlx/pull/923#pullrequestreview-1967475293 will be in today's release.

mzbac:
Is it worth not quantizing the gate (route)? I can't be 100% sure, but I feel that non-quantized seems to perform better when I tried the custom MOE model in the past.

awni:
> I can't be 100% sure, but I feel that non-quantized seems to perform better when I tried the custom MOE model in the past.

Could be related to the bug we just fixed..

> Is it worth not quantizing the gate (route)? 

Maybe not worth it from a run time perspective.. it is a small part of the model. Either way I'm glad not to have a bug there. It's also simpler from an implementation standpoint if we can reliably quantize any linear layer.. keeps the model homogenous.

my-other-github-account:
I can confirm the latest changes do indeed fix the issue!

Now I get:

```
def has_close_elements(numbers: List[float], threshold: float) -> bool:
    """ Check if in given list of numbers, are any two numbers closer to each other than
    given threshold.
    >>> has_close_elements([1.0, 2.0, 3.0], 0.5)
    False
    >>> has_close_elements([1.0, 2.8, 3.0, 4.0, 5.0, 2.0], 0.3)
    True
    """

    numbers.sort()
    for i in range(len(numbers) - 1):
        if numbers[i + 1] - numbers[i] < threshold:
            return True
    return False <|im_end|>
```

As expected!

Great work, thank you all again!

awni:
Thanks for surfacing this issue @my-other-github-account and for the reproducible test case.

M-I:
@my-other-github-account  did you requantize or the just updating mlx-lm was enough?

awni:
You shouldn’t need to requantize

angeloskath (approved):
🚀

PR #632, opened 2024-03-28
Update train.jsonl

eltociear:
minor fix


awni:
That data is copied directly from Wikisql so let's keep it unchanged even if there are typos.

PR #636, opened 2024-03-29
Save lora config

awni:
- Save lora experiment config to json config
- Use safetensors for adapters
-  Read config as well as weights when loading lora model

Directory structure like:

```
adapters/
    adapters.safetensors
    adapter_config.json
```

Config looks like this:

```json
{
    "adapter_path": "adapters",
    "batch_size": 4,
    "config": null,
    "data": "../lora/data/",
    "grad_checkpoint": false,
    "iters": 11,
    "learning_rate": 1e-05,
    "lora_layers": 16,
    "lora_parameters": {
        "rank": 8,
        "alpha": 16,
        "dropout": 0.0,
        "scale": 10.0
    },
    "max_seq_length": 2048,
    "model": "Qwen/Qwen1.5-0.5B",
    "resume_adapter_file": null,
    "save_every": 10,
    "seed": 0,
    "steps_per_eval": 200,
    "steps_per_report": 10,
    "test": true,
    "test_batches": 500,
    "train": true,
    "val_batches": 25
}
```

### DIFF
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -4,6 +4,7 @@ import time
 from dataclasses import dataclass, field
 from functools import partial
 from pathlib import Path
+from typing import Union
 
 import mlx.core as mx
 import mlx.nn as nn
@@ -54,7 +55,7 @@ class TrainingArgs:
         default=2048, metadata={"help": "Maximum sequence length."}
     )
     adapter_file: str = field(
-        default="adapter.npz",
+        default="adapters.safetensors",
         metadata={"help": "Save/load path for the trained adapter weights."},
     )
     grad_checkpoint: bool = field(
@@ -172,18 +173,6 @@ def train(
 ):
     print(f"Starting training..., iters: {args.iters}")
 
-    def checkpoints_path(adapter_file) -> str:
-        checkpoints_path = Path("checkpoints")
-        if Path(adapter_file).parent:
-            checkpoints_path = Path(adapter_file).parent / "checkpoints"
-
-        checkpoints_path.mkdir(parents=True, exist_ok=True)
-
-        return str(checkpoints_path)
-
-    # Create checkpoints directory if it does not exist
-    adapter_path = checkpoints_path(args.adapter_file)
-
     if args.grad_checkpoint:
         grad_checkpoint(model.layers[0])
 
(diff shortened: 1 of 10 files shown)


awni:
@mzbac would you mind looking at this.

llms/mlx_lm/tuner/trainer.py
```diff
@@ -287,13 +276,10 @@ def step(batch):
...
 
-        # Save adapter weights if needed
+        # Save adapter weights
         if (it + 1) % args.steps_per_save == 0:
-            checkpoint_adapter_file = (
-                f"{adapter_path}/{it + 1}_{Path(args.adapter_file).name}"
-            )
-            save_adapter(model=model, adapter_file=checkpoint_adapter_file)
-            print(f"Iter {it + 1}: Saved adapter weights to {checkpoint_adapter_file}.")
+            save_adapter(model=model, adapter_file=args.adapter_file)
```
mzbac, line 281: Is this going to keep overriding the same adapter file, causing us to lose all the intermediate checkpoint weights?

mzbac:
Looks good, just a question about how to handle the weights for the checkout point. Also, we may need to update the server.py script to use adapter-path instead of adapter-file.

llms/mlx_lm/tuner/trainer.py
mzbac: I am thinking maybe we can do something like this:
```
├── lora_adapters/
│   ├── adapter_1/
│   │   ├── adapter_config.json  
│   │   └── adapters.safetensors
│   ├── adapter_2/
│   │   ├── adapter_config.json    
│   │   └── adapters.safetensors
│   ├── adapter_N/
│   |   ├── adapter_config.json    
│   |   └── adapters.safetensors
│   └── adapter_final/
│       ├── adapter_config.json    
│       └── adapters.safetensors
```
It may seem a bit redundant with the `adapter_config.json`, but we would save more precise information for that training iteration. For example, the learning rate of the scheduled learning rate, loss, and anything else related to the training iteration.

awni: I went with
```
lora_adapters/
  -adapters.safetensors
  -adapter_config.json
  -00100_adapters.safetensors
...
``

That fits the way we load better from resumed files and saves from needing to pass all the args into the traiing file or hacking a way around copying the config.

awni:
Comments addressed, thanks @mzbac !

awni:
Closes #622

angeloskath (approved):

llms/mlx_lm/tuner/trainer.py
mzbac: Alright, maybe we could add an optional adapter file name parameter to allow users to specify a specific checkpoint weight during the inference so they don't have to copy the checkpoint weight into a folder and rename it in order to load it. If you are okay with this approach, I can work on a PR for adding that support.

awni: I'm ok with that.. I just didn't think it was a common thing people did?

awni: Do you do it yourself? If so, what for?

mzbac: Yeah, maybe not common use cases, but something from my personal experience. It's not always the adapter with the lowest loss that performs the best, so I normally test out a few adapters to see which one is the best. However, I'm happy to just wait and see if this is something people want to add.

awni: Do you test them on a downstream task? I wonder if we saved the adapter based on validation loss maybe that would help.

PR #644, opened 2024-04-02
Add Qwen2MoE

bozheng-hit:
# Adding Qwen2MoE
This PR adds the support of codes for the Qwen2MoE models. For information about Qwen, please visit https://github.com/QwenLM/Qwen.

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -35,6 +35,8 @@ linear_class_predicate = (
     lambda m: isinstance(m, nn.Linear)
     and m.weight.shape[0]
     != 8  # avoid quantizing gate layers, otherwise we have to re-quant and upload all the mixtral models
+    and m.weight.shape[0]
+    not in [1, 60] # avoid quantizing qwen2_moe gate layers
 )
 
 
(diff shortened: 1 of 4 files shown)


llms/mlx_lm/utils.py
```diff
@@ -35,6 +35,8 @@
     lambda m: isinstance(m, nn.Linear)
     and m.weight.shape[0]
     != 8  # avoid quantizing gate layers, otherwise we have to re-quant and upload all the mixtral models
+    and m.weight.shape[0]
+    not in [1, 60] # avoid quantizing qwen2_moe gate layers
```
awni, line 39: Is it necessary? Would prefer to keep the model consistently quantized to avoid these hacks (the one above is just for back-compatability).

awni:
@Blaizzy @bozheng-hit we have two PRs for the same thing. This one is a little more complete so I would default to use it (e.g. it has a test and supports LoRA). Or @Blaizzy you can pull in those changes and we can make @bozheng-hit a co-author on your commit? Let me know how you would like to proceed

Blaizzy:
I can pull these changes and make @bozheng-hit co-author :)

llms/mlx_lm/utils.py
bozheng-hit: I did not quantize those gating layers in the GPTQ-Int4 version either since routing function is sensitive.

Issue #620, opened 2024-03-26
Getting nan values for training and validation loss


chimezie:
I'm getting nan values for training and validation loss with recent git versions of mlx and mlx_lm (see commit hashes below).

This is the configuration file I'm using:

```yaml
config.yaml:
model: "/path/to/mlx/model"

train: true
lora_layers: 16
batch_size: 4
iters: 2596
steps_per_report: 25
steps_per_eval: 259
val_batches: 32
learning_rate: 7e-5
seed: 4
lora_parameters:
  keys: ["self_attn.q_proj", "self_attn.v_proj", "self_attn.k_proj", "self_attn.o_proj"]
  rank: 64
  alpha: 128
  dropout: 0.0
  scale: 10.0
```

This is how I'm launching the LoRA training:

```bash
% python -m mlx_lm.lora --data mlx-examples/lora/data -c config.yaml                                  
Loading configuration file config.yaml
Loading pretrained model
You set `add_prefix_space`. The tokenizer needs to be converted from the slow tokenizers
Special tokens have been added in the vocabulary, make sure the associated word embeddings are fine-tuned or trained.
Trainable parameters: 0.376% (27.263M/7241.748M)
Loading datasets
Training
Starting training..., iters: 2596
Iter 1: Val loss 2.381, Val took 16.942s
Iter 25: Train loss 8.108, Learning Rate 7.000e-05, It/sec 0.984, Tokens/sec 384.079, Trained Tokens 9761, Peak mem 6.906 GB
Iter 50: Train loss 6.509, Learning Rate 7.000e-05, It/sec 0.934, Tokens/sec 364.859, Trained Tokens 19531, Peak mem 7.058 GB
Iter 75: Train loss 6.027, Learning Rate 7.000e-05, It/sec 0.908, Tokens/sec 359.912, Trained Tokens 29444, Peak mem 8.382 GB
Iter 100: Train loss 5.575, Learning Rate 7.000e-05, It/sec 0.925, Tokens/sec 366.800, Trained Tokens 39358, Peak mem 8.906 GB
Iter 100: Saved adapter weights to checkpoints/100_MrGrammaticaOntology-corpus-pretraining.npz.
Iter 125: Train loss 5.393, Learning Rate 7.000e-05, It/sec 0.952, Tokens/sec 364.491, Trained Tokens 48927, Peak mem 8.906 GB
[...]

PR #645, opened 2024-04-03
Added code for Full fine tune

ziozzang:
The code presented here is derived from the original `lora.py` file, with minimal modifications. The primary addition is the inclusion of full fine-tuning functionality, while preserving the core structure of the original code. This revised version offers a potential starting point for testing the training process on more powerful Mac devices.

Efforts were made to avoid altering any code within the `tuner/*` directory, ensuring that this update does not introduce any conflicts with the legacy codebase.

The code has been successfully tested on a Mac M2 Studio model with 192GB of memory, demonstrating its compatibility with high-performance hardware configurations.

### DIFF
--- a/llms/mlx_lm/FULL.md
+++ b/llms/mlx_lm/FULL.md
@@ -66,6 +66,33 @@ the output location with `--model-file`.
 You can resume fine-tuning with an existing model with
 `--model-file <path_to_model.npz>`.
 
+* NOTICE:
+  - After finetune, you get the model-file only,
+  - You have to model only file, put it back to original model folder. (copy or else. as you wish to.)
+  - You can convert model format npz to safetensor, like this.
+
+```
+# Conversion of file format
+import sys
+from mlx.core import *
+
+#IN: npz format file, OUT: safetensors format file
+def convert_npz_to_safetensor(input_file, output_file):
+    try:
+        # Load the npz file
+        data = load(input_file)
+        
+        # Save the data as safetensor
+        save_safetensors(output_file, data)
+        
+        print(f"Conversion successful. Output saved as {output_file}")
+    except FileNotFoundError:
+        print(f"File not found: {input_file}")
+    except Exception as e:
+        print(f"An error occurred during conversion: {str(e)}")
+
+```
+
 ### Evaluate
 
 To compute test set perplexity use:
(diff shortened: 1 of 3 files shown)


N8python:
Do you perform your full fine-tune in float32?

ziozzang:
Fixed load/save function. fully tested with Phi-2 2.8B model.
- model file saving works well and resuming of model training works well also.

> Do you perform your full fine-tune in float32?

no. I just copy code from mlx-lm/lora.py and fixed to run as full fine-tune. to make compatible original tuner/* code.

ziozzang:
test example.
- M2 studio / 192GB
- Model: phi-2 2.8B
- Training set: chat completion / 400 items. (only iter set to 10, to show running demo)

```
$ python -m mlx_lm.full -c full_config.yaml

Loading configuration file full_config.yaml
Loading pretrained model
model file loaded: model.safetensors
Loading datasets
Training
Starting training..., iters: 10
Iter 1: Val loss 1.405, Val took 2.793s
Iter 5: Train loss 1.094, Learning Rate 1.000e-05, It/sec 0.956, Tokens/sec 779.347, Trained Tokens 4077, Peak mem 22.747 GB
Iter 10: Train loss 1.311, Learning Rate 1.000e-05, It/sec 0.668, Tokens/sec 589.665, Trained Tokens 8490, Peak mem 26.198 GB
Iter 10: Val loss 1.542, Val took 2.936s
Saved final adapter weights to adapter.npz.
Saved final model weights to model.safetensors.
```

N8python:
Tried training qwen-1.8b. NaN loss immediately. Will try phi-2.

ziozzang:
> Tried training qwen-1.8b. NaN loss immediately. Will try phi-2.

when I tried Gemma-2b, same NaN loss. maybe, it's  foundation code issue. maybe in models/* ? I didn't check.

N8python:
Think its the float16.

N8python:
Just checked - NaN w/ phi.

chimezie:
I've opened an older issue (#620) regarding training error NaN values

awni:
This is cool, and I think it would be nice to support. We might be able to do it with a far smaller diff however. Something like:
- Have a training type field in the config 
- If it's `full_fine_tune` then don't freeze the model / don't use LoRA layers

Everything else should be the same. Wdyt?

awni:
This landed in #903

Issue #647, opened 2024-04-04
Server produce gibberish output for some model


alwint3r:
The issue is observed on a 4-bit quantized `cognitivecomputations/dolphin-2.8-mistral-7b-v02` model.

Here's a curl request to reproduce the issue:
```bash
curl -X POST localhost:8080/v1/chat/completions \
    -H "Content-Type: application/json" \
    -d '{
            "messages":[
                  {
                        "role": "user",
                        "content": "Write a sentence of opening line for a dune-like movie"
                  }
            ],
            "max_tokens":500,
            "temperature":0.9
      }'
```

I would expect a response with the following line from the model

```
In a world of shifting sands and endless winds, a stranger's arrival sets off a chain of events that would determine the fate of a people and the land.
```

Here's the complete response that I received instead:

```json
{
  "id": "chatcmpl-85aa4e55-521b-4d35-960b-a2a4333b68dc",
  "system_fingerprint": "fp_beaf8d29-15d9-4c27-b583-2817518b51a6",
  "object": "chat.completions",
  "model": "default_model",
  "created": 1712188000,
  "choices": [
    {
      "index": 0,
      "logprobs": null,
      "finish_reason": "stop",
      "message": {
        "role": "assistant",
[...]

PR #648, opened 2024-04-04
Add workaround for apply_chat_template on server.py

alwint3r:
This PR is a workaround for the issue related to the transformers library as described in issue #647.

### DIFF
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -372,9 +372,10 @@ class APIHandler(BaseHTTPRequestHandler):
         if hasattr(TOKENIZER, "apply_chat_template") and TOKENIZER.chat_template:
             prompt = TOKENIZER.apply_chat_template(
                 body["messages"],
-                tokenize=True,
+                tokenize=False,
                 add_generation_prompt=True,
             )
+            prompt = TOKENIZER.encode(prompt)
         else:
             prompt = convert_chat(body["messages"], body.get("role_mapping"))
             prompt = TOKENIZER.encode(prompt)


mzbac:
After reading the documentation [here](https://huggingface.co/docs/transformers/main/chat_templating#what-template-should-i-use), it appears that disabling add_special_tokens is intentional. Based on the documentation, it seems like we should address this issue in the chat template.

mzbac:
I just double-checked on my local and found that updating the chat_template in tokenizer_config.json to `"chat_template": "{% if not add_generation_prompt is defined %}{% set add_generation_prompt = false %}{% endif %}+'<s>'+{% for message in messages %}{{'<|im_start|>' + message['role'] + '\n' + message['content'] + '<|im_end|>' + '\n'}}{% endfor %}{% if add_generation_prompt %}{{ '<|im_start|>assistant\n' }}{% endif %}"` fixed the issue as well.

alwint3r:
Interesting! Adding the `<s>` before the messages fixed the issue. Thanks @mzbac, I'll close the issue and PR for now.

mzbac:
> Interesting! Adding the `<s>` before the messages fixed the issue. Thanks @mzbac, I'll close the issue and PR for now.

Apologies for leading you in the wrong direction. I didn't know that Transformers prefer handling special characters in chat templates over the tokenizer.

alwint3r:
> > Interesting! Adding the `<s>` before the messages fixed the issue. Thanks @mzbac, I'll close the issue and PR for now.
> 
> Apologies for leading you in the wrong direction. I didn't know that Transformers prefer handling special characters in chat templates over the tokenizer.

No worries @mzbac. Thanks for the help.

PR #651, opened 2024-04-05
Add streaming detokenizers

angeloskath:
Enables true streaming detokenization for SPM and BPE models while simplifying the generate function a tiny bit (moving the logic in the `tokenizer_utils` basically).

```
              | Before | After
--------------+--------+--------
Mistral-4 bit |  95.76 | 96.60
Qwen1.5 1.8B  |  89.70 | 90.31
```

It also makes them true linear complexity detokenizers regardless of the length of the longest line.


llms/mlx_lm/tokenizer_utils.py
```diff
@@ -0,0 +1,296 @@
...
+
+    @property
+    def tokens(self):
+        return self._tokens
+
+
+class SPMStreamingDetokenizer(StreamingDetokenizer):
+    """A streaming detokenizer for SPM models.
+
+    It adds tokens to the text if the next token starts with the special SPM
```
mzbac, line 102: Will this fix the issue of decoding token by token, causing the missing whitespace issue in the Hugging Face tokenizer implementation?

angeloskath: Yes exactly. The tokenizers library is designed to encode/decode the full text at a time. That means that decoders like `{"type": "Strip", "content": " ", "start": 1, "stop": 0}` can't be called token by token.

Our original code was calling `decode` on the full list of tokens each time which is `O(T^2)` for `T` being the number of tokens generated and then we changed it to flush the token list for every line (which I think is also done in the HF transformers library).

This is one more step where you actually just decode token by token. The problem is that it depends on the tokenizer implementation so if it isn't exactly this BPE configuration or the SPM configuration it will fallback to line by line.

awni:
Looks great! I assume you noticed there is a test failing? O/w I think we should merge it.

llms/mlx_lm/tokenizer_utils.py
mzbac: Nice, I'll take a look and see if we can adapt this in the server.py as well.

angeloskath:
@awni yeah, had pushed a wrong fix. It should be good to go now, feel free to stamp whenever you want.

awni (approved):
This is very cool, the API is 🔥.  I think it has the potential to be useful in a lot of places even outside of MLX LM.

PR #654, opened 2024-04-05
Update model card describe

madroidmaq:
- Add full link jump
- Add the address of the model uploader's Hugging Face homepage
- Add the -U parameter to always install the latest version of dependencies

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -18,6 +18,7 @@ from mlx.utils import tree_flatten
 from transformers import AutoConfig, AutoTokenizer, PreTrainedTokenizer
 
 from .sample_utils import top_p_sampling
+
 # Local imports
 from .tuner.utils import apply_lora_layers
 from .tuner.utils import dequantize as dequantize_model
@@ -32,8 +33,8 @@ MAX_FILE_SIZE_GB = 5
 
 linear_class_predicate = (
     lambda m: isinstance(m, nn.Linear)
-              and m.weight.shape[0]
-              != 8  # avoid quantizing gate layers, otherwise we have to re-quant and upload all the mixtral models
+    and m.weight.shape[0]
+    != 8  # avoid quantizing gate layers, otherwise we have to re-quant and upload all the mixtral models
 )
 
 
@@ -326,7 +327,7 @@ def load_model(model_path: Path, lazy: bool = False) -> nn.Module:
             vocab_size = config["vocab_size"]
             extended_linear_class_predicate = (
                 lambda layer: linear_class_predicate(layer)
-                              and layer.weight.shape[0] != vocab_size
+                and layer.weight.shape[0] != vocab_size
             )
             nn.QuantizedLinear.quantize_module(
                 model,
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/utils.py
```diff
@@ -436,17 +436,25 @@ def upload_to_hub(path: str, upload_repo: str, hf_path: str):
...
     card.data.tags = ["mlx"] if card.data.tags is None else card.data.tags + ["mlx"]
     card.text = dedent(
         f"""
         # {upload_repo}
-        This model was converted to MLX format from [`{hf_path}`]() using mlx-lm version **{__version__}**.
-        Refer to the [original model card](https://huggingface.co/{hf_path}) for more details on the model.
+        
+        The Model [{upload_repo}](https://huggingface.co/{upload_repo}) was converted to MLX format from [{hf_path}](https://huggingface.co/{hf_path}) using mlx-lm version **{__version__}**.
+        
+        Model added by [{fullname}](https://huggingface.co/{name}).
```
awni, line 452: I'm not against it.. but I'm also not sure everyone will want their username by default in the model care 🤔

llms/mlx_lm/utils.py
```diff
@@ -436,17 +436,25 @@ def upload_to_hub(path: str, upload_repo: str, hf_path: str):
...
+        
         ## Use with mlx
 
         ```bash
-        pip install mlx-lm
+        pip install -U mlx mlx-lm
```
awni, line 457: I don't think you need the `mlx` there. If it's not installed or too old installing MLX LM will update it.

llms/mlx_lm/utils.py
madroidmaq: Initially, I wasn't sure about this either. However, when I saw people adding their personal information to the model cards (e.g., [mlx-community/dbrx-instruct-4bit](https://huggingface.co/mlx-community/dbrx-instruct-4bit), [mlx-community/gemma-1.1-2b-it-4bit](https://huggingface.co/mlx-community/gemma-1.1-2b-it-4bit), etc.), I felt that they should be proud of uploading their models. At the same time, when I use models from the mlx community, I sometimes want to know who contributed to them.

Of course, some people might not want to display their information. This adjustment is not a must for me, and if needed, I can remove that part of the content.

llms/mlx_lm/utils.py
madroidmaq: I will restore it to the original way.

llms/mlx_lm/utils.py
awni: It's totally ok if people want to add their contact info in the cards for models they upload, but I prefer not to make it the default behavior as it may surprise people who are just experimenting or something like that.

awni:
@madroidmaq are you coming back to this?

madroidmaq:
> @madroidmaq are you coming back to this?

@awni I just adjusted the description of the connection to see if there is anything else that needs to be adjusted.

awni (approved):
Thank you!!

PR #657, opened 2024-04-06
Set mlx version to 0.8.1

julianadler:
The script [lora/lora.py](https://github.com/ml-explore/mlx-examples/blob/main/lora/lora.py) does not seem to work with `mlx` version `0.9.0`.

Forcing it to `0.8.1`, allows training using `lora.py`.

### DIFF
--- a/lora/requirements.txt
+++ b/lora/requirements.txt
@@ -1,3 +1,3 @@
-mlx>=0.8.0
+mlx==0.8.1
 transformers
 numpy


madroidmaq:
@julianadler Can you provide your error message, maybe someone else in the community can fix this based on the latest mlx version (0.9.0) instead of locking the mlx version.

awni:
I agree with @madroidmaq. Also I just tried it with 0.9.1 and it works fine for me 🤔 

```
python lora.py --model mlx-community/NeuralBeagle14-7B-4bit-mlx --train --iters 20
```

Please send an issue with the command you used and error you got and we can debug.

julianadler:
Thanks for the feedback, I'll get the error message when I'll get a chance

PR #661, opened 2024-04-08
NLLB Model

ToluClassics:
Add NLLB (No Language Left Behind) translation models in MLX


awni:
@ToluClassics this is very cool, sorry for the delay in taking a look.

I'm wondering if we should add it to MLX examples or not. On the one hand we don't have a pure translation example. Would this be a good canonical example to serve as a reference?

On the other hand we also have T5 which is basically the same architecture and can do translation... so I'm not sure how useful this example is in addition to that?

An alternative would be for you to start a `mlx-translation` repo or `mlx-nllb` and we could point people to it.

Wdyt?

ToluClassics:
Oh yeah that’s fine, it’s actually included as an example in [mlx-transformers](https://github.com/ToluClassics/mlx-transformers/tree/main) for anyone that wants to try it out.

awni:
Oh awesome!

PR #665, opened 2024-04-08
Fix for empty initial string

### LINKED ISSUE #653: Enso error when running llama-2- w/70B parameters
Hi!
I was recently experimenting with running the llama 2 models and encountered a strange error where the 70B parameter model would just give me "enso" over and over again for any prompt (In this instance I was just testing with who is Abe Linconl which the 13B and 7B parameter were able to answer) Does anybody know what could be causing this? And not sure if it is an MLX issue versus a llama 2 issue.

Thanks!

awni:
As title.

Closes #653

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -249,7 +249,9 @@ def generate(
 
         if verbose:
             s = tokenizer.decode(tokens)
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
👍

PR #674, opened 2024-04-11
Always resume downloads

da-z:


awni (approved):
Nice! I wonder why it's defaulted to `False`.

awni:
I think you need to run the pre-commit hooks. The linting test failed.

Issue #671, opened 2024-04-10
Mixtral Bug 


yuhai-china:
I try to make a mlx version of my MOE model cloudyu/Mixtral_34Bx2_MoE_60B https://huggingface.co/cloudyu/Mixtral_34Bx2_MoE_60B

But the mlx version doesn't work, I don't know the reason.

it report:
mlx-examples/llms/mlx_lm/models/mixtral.py", line 137, in call
mx.argpartition(-gates, kth=ne, axis=-1)[:, :ne]
ValueError: [argpartition] Received invalid kth 2 along axis -1 for array with shape: (1,2)

I read the code
https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/mixtral.py

at 130 line:

it should be ne = self.num_experts_per_tok - 1  instead of ne = self.num_experts_per_tok

ne = self.num_experts_per_tok that means 1+self.num_experts_per_tok experts working.

awni:
Yup, that's a bug.. do you want to send a PR with your suggested fix?

yuhai-china:
> Yup, that's a bug.. do you want to send a PR with your suggested fix?

https://github.com/ml-explore/mlx-examples/pull/673

PR #676, opened 2024-04-11
Fix argpartition call in  Mixtral and other MOES

awni:
Fixes an obo error in the argpartition call but it was not really a bug since our argpartition uses sort. Better to fix though.


awni:
Closes #671

angeloskath (approved):

PR #679, opened 2024-04-13
More async eval

awni:
Companion to https://github.com/ml-explore/mlx/pull/998

Benchmarks generating with Gemma on M1 Max:

```
MLX_MAX_OPS_PER_BUFFER=10 python -m mlx_lm.generate --model mlx_model --prompt "Write a story about Einstein" --temp 0.0 --max-tokens 256
```

Pre: `128.018 tokens-per-sec`
Post: `140.112 tokens-per-sec`

Pre + Max Ops 80: `143.687 tokens-per-sec`
Post + Max Ops 80: `164.307 tokens-per-sec`

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -188,14 +188,15 @@ def generate_step(
                 repetition_context = repetition_context[-repetition_context_size:]
         return y, prob
 
-    y, prob = _step(y)
+    y, p = _step(y)
 
+    sync = mx.async_eval(y)
     while True:
-        sync = mx.async_eval(y)
-        next_out = _step(y)
+        next_y, next_p = _step(y)
+        next_sync = mx.async_eval(next_y)
         sync.wait()
-        yield y.item(), prob
-        y, prob = next_out
+        yield y.item(), p
+        sync, y, p = next_sync, next_y, next_p
 
 
 def generate(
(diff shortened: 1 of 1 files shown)


ivanfioravanti:
I did several tests on M3 Max, output result is consistent, performance similar,  but it's slower with --colorize

python -m mlx_lm.generate --model mlx-community/Mistral-7B-Instruct-v0.2-4bit --prompt "Write a story on Albert Einstein of 40 words" --seed 42 
PRE:  75.0 t/s POST: 75.3

adding --colorize
PRE:  73.8 t/s POST: 70.4

awni:
Are you using the right version of MLX? This depends on a branch that is not in main yet..

awni:
You can try it if you want but you have to build this branch: https://github.com/ml-explore/mlx/pull/992

And it's pretty experimental right now, so be warned 😉

ivanfioravanti:
No I just applied the changed utils.py in the PR to test it against main. 
I'll try the whole branch to test the speed! 🚀

ivanfioravanti:
Branch is always slightly faster in tests 🚀. 
I confirm the slow down using --colorize, but this is not an issue no? At the end is like running the model in debug mode.

awni:
> I confirm the slow down using --colorize, but this is not an issue no? At the end is like running the model in debug mode.

Yea less concerning. But will take a look to make sure nothing strange is going on there. I can't imagine why it would be slower than before with `--colorize`.. so that sounds a bit fishy.

awni:
Landed in #680

PR #680, opened 2024-04-14
Quantize embedding / Update quantize API

awni:
As title. Relies on https://github.com/ml-explore/mlx/pull/994


angeloskath (approved):
Nice!

PR #681, opened 2024-04-15
Adapt SPM detokenizer to optionally not trim initial space

angeloskath:
This basically allows streaming detokenization for the Gemma models.

If this continues (as in we find more decoders that need special cases) it might be worth checking if composable streaming detokenizers would be any slower. In that case we could just implement `ByteFallback`, `Strip`, `Replace` and a couple of others and compose them according to the `tokenizer.json` file.


awni (approved):
🙏

PR #682, opened 2024-04-16
Create executables for generate, lora, server, merge, convert

lkhphuc:
This PR create executable commands on installation: `mlx_lm.generate, mlx_lm.lora, mlx_lm.server, mlx_lm.merge, mlx_lm.convert`, without having to invoke `python -m`.

This is especially helpful to use with global installation tool such as `pipx, condax, pixi global` etc as user won't have to manage their virtualenv manually.

My motivation for this is that I want a quick way to startup a local language server like ollama, but using mlx.

### DIFF
--- a/llms/README.md
+++ b/llms/README.md
@@ -66,7 +66,7 @@ To see a description of all the arguments you can do:
 You can also use `mlx-lm` from the command line with:
 
 ```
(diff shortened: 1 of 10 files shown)


awni (approved):
Nice addition, thank you!!

PR #685, opened 2024-04-16
MiniCPM implementation

Goekdeniz-Guelmez:
Added support for the MiniCPM Architecture in the mlx_lm folder. 

## Changes Made
- Added `minicpm.py` in `mlx_lm/models` which contains the implementation of the MiniCPM model architecture.
- Modified `LORA.md` to include the MiniCPM model.
- Updated the `trainer/utils.py` to include the new model type in the `linear_to_lora_layers` function.


llms/mlx_lm/models/minicpm.py
```diff
@@ -0,0 +1,203 @@
...
+                values = mx.concatenate([value_cache, values], axis=2)
+            else:
+                queries = self.rope(queries)
+                keys = self.rope(keys)
+
+            attn_output = mx.fast.scaled_dot_product_attention(
+                queries, keys, values, scale=self.scale, mask=mask
+            )
+
+            # attn_output = attn_output.reshape(B, L, self.hidden_size) # Originale impementation
```
awni, line 109: That comment seems unecessary?

llms/mlx_lm/models/minicpm.py
```diff
@@ -0,0 +1,203 @@
...
+        out = self.lm_head(out / (self.args.hidden_size / self.args.dim_model_base))
+        return out, cache
+
+    def sanitize(self, weights):
+        if "lm_head.weight" not in weights:
+            weights["lm_head.weight"] = weights["model.embed_tokens.weight"]
+        # Remove unused precomputed rotary freqs
+        return {
+            k: v for k, v in weights.items() if "self_attn.rotary_emb.inv_freq" not in k
+        }
```
awni, line 199: Do you need this sanitize method? It's better not to have it if you don't need.

awni:
Looks great. Should I use one of these models for testing: https://huggingface.co/openbmb/MiniCPM-2B-sft-bf16 ?

awni:
Or maybe this one looks like it has safetensors: https://huggingface.co/openbmb/MiniCPM-V

llms/mlx_lm/models/minicpm.py
Goekdeniz-Guelmez: yes, sorry for the unnecessary code blocks, they can be removed.

Goekdeniz-Guelmez:
> ~Or maybe this one looks like it has safetensors: https://huggingface.co/openbmb/MiniCPM-V~ has the wrong model type: https://huggingface.co/openbmb/MiniCPM-V/blob/main/config.json#L23

I noticed that most of the models from the original creators are in the wrong format, therefore I used katuni4ka/tiny-random-minicpm, and yes, https://huggingface.co/openbmb/MiniCPM-V is a Vision Language model (from what I understood) which is the wrong model type. So I suggest you to try https://huggingface.co/katuni4ka/tiny-random-minicpm. It's small so it won't use much memory too.

llms/mlx_lm/models/minicpm.py
Goekdeniz-Guelmez: unfortunately when training without the sanitize method, I get the same error we had when training the Qwen2 models.

File "/Users/gokdenizgulmez/Desktop/1/mlx-examples/llms/mlx_lm/utils.py", line 349, in load_model
                            ^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'items'

so from my understanding, the sanitize method is needed for weight management.

llms/mlx_lm/models/minicpm.py
Goekdeniz-Guelmez: yes, your right, I have removed it 👍

llms/mlx_lm/models/minicpm.py
awni: That error message doesn't make sense. I don't know why you are receiving it or why the `sanitize` removes it but the `sanitize` is not likely the right answer. We should debug what's actually causing that.

Goekdeniz-Guelmez: I think I found out, what's causing it. Upon reviewing the actual model structure (`katuni4ka/tiny-random-minicpm`), I found that lm_head.weight is indeed represented by model.embed_tokens.weight. I have removed the rope code parts in the sanitized method and it works fine now. That error is still very weird and I have never gotten it after too. I think that that was a layer 8 error lol. When removing the sanitize method and running it, I get `ValueError: Missing parameters: lm_head.weight.`, which is a more understandable error.

```python
def sanitize(self, weights):
        if "lm_head.weight" not in weights:
            weights["lm_head.weight"] = weights["model.embed_tokens.weight"]
        return weights
 ```
 
So the sanitize method is to ensure that the lm_head.weight key is present and correctly assigned, especially when it's not explicitly set during model initialization.

That's at least from my understanding.

Comment (awni) [reply llms/mlx_lm/models/minicpm.py]: Great! So rather than setting the `lm_head` to the embedding weight, it would be better to use it as a linear layer directly. That way parameters don't get duplicated taking up more memory.

I would do the following:
1. Remove `lm_head` from the model
2. Do `out @ model.embed_tokens.weight.T` to produce the final output.

Sound good?

Comment (Goekdeniz-Guelmez) [reply llms/mlx_lm/models/minicpm.py]: sounds great!

The new Model class is then:
```python
class Model(nn.Module):
    def __init__(self, args: ModelArgs):
        super().__init__()
        self.args = args
        self.model_type = args.model_type
        self.model = MiniCPMModel(args)

    def __call__(
        self,
        inputs: mx.array,
        cache=None,
    ):
        out, cache = self.model(inputs, cache)
        out = out @ self.model.embed_tokens.weight.T
        return out, cache
 ```
 
 
 Also training and inference is way better now too. Thanks for the quick response.

Comment (Goekdeniz-Guelmez) [reply llms/mlx_lm/models/minicpm.py]: Well when using the original model I get this error:

```
File "/Users/gokdenizgulmez/Desktop/1/mlx-examples/llms/mlx_lm/utils.py", line 349, in load_model
  File "/Users/gokdenizgulmez/miniconda3/envs/mlx/lib/python3.12/site-packages/mlx/nn/layers/base.py", line 211, in load_weights
    raise ValueError(f"Received parameters not in model: {extras}.")
ValueError: Received parameters not in model: lm_head.weight.
```

Adding the lm_head part and the rest, trains the model right. It looks like the previous fix worked for `https://huggingface.co/katuni4ka/tiny-random-minicpm` but not for the original:  `https://huggingface.co/openbmb/MiniCPM-2B-128k/tree/aff9444f5c3f1795c2baddadfcf7be54d397ca11`

Comment (awni) [reply llms/mlx_lm/models/minicpm.py]: Yup in that case you'll need to read the flag in the config `tie_word_embeddings` https://huggingface.co/openbmb/MiniCPM-2B-128k/blob/aff9444f5c3f1795c2baddadfcf7be54d397ca11/config.json#L33

If it is `True` (and maybe unset) then you would do as you have now. If it is `False` add separate `lm_head`

Comment (Goekdeniz-Guelmez) [reply llms/mlx_lm/models/minicpm.py]: Thank you for pointing me in the right direction regarding the tie_word_embeddings configuration. I have added the conditional handling of lm_head based on the flag, and the models are now performing as expected.

**Comment by awni** [comment]
> therefore I used katuni4ka/tiny-random-minicpm

What's a good way to verify it's working? I tried it with an english prompt and it looked pretty bad, but I'm not sure if it's supposed to work.

I'm also not sure about adding this model given there are no supported versions in Hugging Face.

**Comment by Goekdeniz-Guelmez** [comment]
Yea the tiny-random-minicpm has not been trained, the 128K context model uses different RoPE calculations that are not implemented yet, but here is the https://huggingface.co/openbmb/MiniCPM-2B-sft-bf16/tree/ed3546a920d3005a1495609e6919c991bd7f036d. Works as intended when generating, but you have to set `"model_type": "minicpm"` in the config.json file. But after that it should work.

**Comment by awni** [comment]
Cool, I will try that. Also, thanks for uploading them.

Feel free to do it in the mlx-community or your own space.

The community is here: https://huggingface.co/mlx-community

You can upload with something like `mlx_lm.convert --hf-path <path-to-model> --mlx-path <output-path> --upload-repo mlx-community/minicpm-sft-2b-4bit -q`

**Comment by awni** [comment]
I tried the model you pointed me to and it generates garbage output for the prompts I gave it. Were you able to get decent output?

For example:
```
python -m mlx_lm.generate --model openbmb/MiniCPM-2B-sft-bf16 --prompt "山东省最高的山是哪座山, 它比黄山高还是矮？差距多少？" --max-tokens 256 --temp=0.0
```

**Comment by Goekdeniz-Guelmez** [comment]
that seems weird, i get normal outputs like:

```sh
(mlx) gokdenizgulmez@MacBook-Air-von-Gokdeniz llms % python -m mlx_lm.generate \
    --model /Users/gokdenizgulmez/Desktop/MiniCPM-2B-sft-bf16-4bit \
    --prompt "hello how are you." \
    --colorize
None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
==========
Prompt: <用户>hello how are you.<AI>
Hello! I'm an AI language model, and I'm doing well. How about you?
==========
Prompt: 42.488 tokens-per-sec
Generation: 23.780 tokens-per-sec
```

Fine tuning, and quantizing works just as expected too. Did you use the safetensors variant?

I am uploading the model and will tell you when it's uploaded.

awni:
Working now, thanks!

awni (approved):
Thanks!!

Goekdeniz-Guelmez:
No Problem! Also thanks for adding it in. I’ll probably try to push more models in the future.

fblissjr:
Has anyone been able to get minicpm v2 to work? 

```
python -m mlx_lm.convert --hf-path openbmb/MiniCPM-V-2 --mlx-path ./minicpm-v2-mlx
The cache for model files in Transformers v4.22.0 has been updated. Migrating your old cache. This is a one-time only operation. You can interrupt this and resume the migration later on by calling `transformers.utils.move_cache()`.
0it [00:00, ?it/s]
[INFO] Loading
modeling_minicpmv.py: 100%|██████████████████████████████████████████████████████████████████████████████████████| 19.5k/19.5k [00:00<00:00, 36.7MB/s]
configuration_minicpm.py: 100%|██████████████████████████████████████████████████████████████████████████████████| 10.9k/10.9k [00:00<00:00, 78.5MB/s]
resampler.py: 100%|██████████████████████████████████████████████████████████████████████████████████████████████| 5.59k/5.59k [00:00<00:00, 16.0MB/s]
modeling_minicpm.py: 100%|███████████████████████████████████████████████████████████████████████████████████████| 71.2k/71.2k [00:00<00:00, 1.03MB/s]
config.json: 100%|███████████████████████████████████████████████████████████████████████████████████████████████| 1.18k/1.18k [00:00<00:00, 16.7MB/s]
generation_config.json: 100%|████████████████████████████████████████████████████████████████████████████████████████| 111/111 [00:00<00:00, 1.68MB/s]
special_tokens_map.json: 100%|███████████████████████████████████████████████████████████████████████████████████████| 651/651 [00:00<00:00, 8.45MB/s]
model.safetensors.index.json: 100%|██████████████████████████████████████████████████████████████████████████████| 54.6k/54.6k [00:00<00:00, 2.17MB/s]
tokenizer.json: 100%|████████████████████████████████████████████████████████████████████████████████████████████| 6.20M/6.20M [00:00<00:00, 23.7MB/s]
tokenizer_config.json: 100%|█████████████████████████████████████████████████████████████████████████████████████| 3.34k/3.34k [00:00<00:00, 30.6MB/s]
tokenizer.model: 100%|███████████████████████████████████████████████████████████████████████████████████████████| 1.99M/1.99M [00:00<00:00, 12.7MB/s]
model-00002-of-00002.safetensors: 100%|██████████████████████████████████████████████████████████████████████████| 1.88G/1.88G [00:34<00:00, 54.4MB/s]
model-00001-of-00002.safetensors: 100%|██████████████████████████████████████████████████████████████████████████| 4.99G/4.99G [01:01<00:00, 81.4MB/s]
Fetching 13 files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████| 13/13 [01:01<00:00,  4.74s/it]
ERROR:root:Model type minicpmv not supported.█████████████████████████████████████████████████████████████████████| 4.99G/4.99G [01:01<00:00, 108MB/s]

ValueError: Model type minicpmv not supported.
```

PR #687, opened 2024-04-17
Add multi_commits to uploads

versae:
This PR adds an optional flag `--upload-multi-commits` to use an experimental feature in the HuggingFace API that creates multiple commits per upload. It is very convenient for big models with multiple files. It also allows for resuming the upload if something fails by simply re-running the command. I've tested it with `convert`, it looks [like this](https://huggingface.co/mlx-community/zephyr-orpo-141b-A35b-v0.1-8bit/discussions/2).

### DIFF
--- a/llms/mlx_lm/convert.py
+++ b/llms/mlx_lm/convert.py
@@ -43,6 +43,12 @@ def configure_parser() -> argparse.ArgumentParser:
         default=None,
     )
     parser.add_argument(
+        "--upload-multi-commits",
+        help="Use multiple commits to upload the model, which allows for resuming.",
+        action="store_true",
+        default=False,
+    )
+    parser.add_argument(
         "-d",
         "--dequantize",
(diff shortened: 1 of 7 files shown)


awni:
Sorry for the delay! That's nice. I wonder if we should just make it the default behavior, seems quite useful? Any downside?

versae:
The only downside is if HF decides to change the API for some reason, as the feature is still marked as experimental. Maybe a risk worth taking.

awni:
I am cool with that risk. If they change it we can reverse this / patch it quickly.

versae:
Oh, you are thinking just not giving the option of not using the multi commits? If that's the case, yes, the PR will be much cleaner. I'll open a new one if that's the case.

awni:
That's what I was thinking. I guess it would just be a two line change in the call to upload?

PR #688, opened 2024-04-17
Add NanoLLaVA Example

Blaizzy:
This PR adds NanoLlaVA (a Vision Language Model) example and updates the VLMs folder.
Twitter: @Prince_Canuma

PR co-authored with @qnguyen3, the creator of NanoLlaVA.
Twitter: @stablequan

Todo:
- [ ] Refactor code
- [ ] VLM Quantization
- [ ] Tests


lin72h:
Prince, thanks for working on this. I noticed  @qnguyen3 also the author of [bunny model](https://huggingface.co/BAAI/Bunny-v1_0-2B-zh), and both nanollava and bunny architecturally very similar, So does this PR also works for bunny?

qnguyen3:
> Prince, thanks for working on this. I noticed @qnguyen3 also the author of [bunny model](https://huggingface.co/BAAI/Bunny-v1_0-2B-zh), and both nanollava and bunny architecturally very similar, So does this PR also works for bunny?

Hi @lin72h , I am not the author of the Bunny series, however, I reused the code base from Bunny Qwen for the training. This code should work with the Bunny Qwen1.5-1.8B + SigLIP variant

awni:
@Blaizzy this might make more sense in a standalone repo or a separate mlx-vllm package. This repo is more intended to have a canonical example of each type of model rather than be a model zoo for all models which would be awesome but we don't have the bandwidth to maintain it yet.

MLX LM is of course a big exception to that (but ideally it would be pulled out of this and put into something standalone in the future). 

Wdyt about starting a community lead mlx-vllm package? We could always reconsider incorporating it into MLX LM or here if it makes sense in the future.

lin72h:
> Hi @lin72h , I am not the author of the Bunny series, however, I reused the code base from Bunny Qwen for the training. This code should work with the Bunny Qwen1.5-1.8B + SigLIP variant

Thanks for clarification

KE7:
> @Blaizzy this might make more sense in a standalone repo or a separate mlx-vllm package. This repo is more intended to have a canonical example of each type of model rather than be a model zoo for all models which would be awesome but we don't have the bandwidth to maintain it yet.
> 
> MLX LM is of course a big exception to that (but ideally it would be pulled out of this and put into something standalone in the future).
> 
> Wdyt about starting a community lead mlx-vllm package? We could always reconsider incorporating it into MLX LM or here if it makes sense in the future.

I don't feel strongly about this, but given the current state, my 2 cent's would be to include it now since llms is likely to remain here as well. It would be a confusing developer experience to do a pip install for everything except vlms

awni:
> I don't feel strongly about this, but given the current state, my 2 cent's would be to include it now since llms is likely to remain here as well. It would be a confusing developer experience to do a pip install for everything except vlms

`pip install mlx-lm` *only* gives you LLMs in `llms/mlx_lm`. It does not pip install all the examples.

KE7:
> > I don't feel strongly about this, but given the current state, my 2 cent's would be to include it now since llms is likely to remain here as well. It would be a confusing developer experience to do a pip install for everything except vlms
> 
> `pip install mlx-lm` _only_ gives you LLMs in `llms/mlx_lm`. It does not pip install all the examples.

My fault. Too many things to keep track of 😅

Blaizzy:
> @Blaizzy this might make more sense in a standalone repo or a separate mlx-vllm package. This repo is more intended to have a canonical example of each type of model rather than be a model zoo for all models which would be awesome but we don't have the bandwidth to maintain it yet.
> 
> MLX LM is of course a big exception to that (but ideally it would be pulled out of this and put into something standalone in the future).
> 
> Wdyt about starting a community lead mlx-vllm package? We could always reconsider incorporating it into MLX LM or here if it makes sense in the future.

@awni @lin72h @qnguyen3 
I understand that, and thought the same. 

That's why 3 days I ago I went ahead and started building mlx-vllm :). I will announce it tomorrow on X.

https://github.com/Blaizzy/mlx-vllm

If you would like to contribute, reach out to me. I need help with developing tests, quantisation and developing the inference API. 🚀

PR #690, opened 2024-04-18
fix(mlx-lm): broken server.py

mzbac:
There are two changes breaking the current server.py. The async eval now returns the token item instead of just the token, causing the "item" method to not be found. Additionally, the entryscript is making the global MODEL and TOKENIZER not found. Therefore, this PR is fixing those two issues and also adapting to use the new tokenizer wrapper's detokenizer to improve decoding speed.


awni (approved):
Awesome, thanks for the fixes and sorry for breaking that. The tests will keep us honest in the future 😄

awni:
Merging even with the broken test because the patch is in #680

PR #691, opened 2024-04-18
Support for Meta Llama 3

js-ts:
Since Llama 3 uses [tiktoken](https://github.com/meta-llama/llama3/blob/main/llama/tokenizer.py#L20)  and V1 and V2 use sentencepiece, needed to create a separate script to run Llama3

```
 % python llama3.py --prompt "hello"  
[INFO] Loading model from mlx_model/weights.npz.
mlx_model
Press enter to start generation
------
hello
, I am a new member here. I am a 25 year old male and I am interested in learning more about the world of photography. I have a camera and I have taken a few photos, but I am not sure if I am doing it right. I am looking for some advice and tips on how to improve my photography skills.
Welcome to the community! I'm happy to help you with your photography journey.

To get started, can you tell me a bit more about your current setup
```

### DIFF
--- a/llms/llama/requirements.txt
+++ b/llms/llama/requirements.txt
@@ -2,3 +2,4 @@ mlx>=0.8.0
 sentencepiece
 torch
 numpy
+tiktoken==0.4.0
--- a/llms/llama/README.md
+++ b/llms/llama/README.md
@@ -52,12 +52,16 @@ the converted `weights.npz`, `tokenizer.model`, and `config.json` there.
 ### Run
 
 Once you've converted the weights to MLX format, you can interact with the
-LlamA model:
+LlamA 1,2 model:
 
 ```
 python llama.py --prompt "hello"
 ```
+LlamA 3 model:
 
+```
+python llama3.py --prompt "hello"
+```
 Run `python llama.py --help` for more details.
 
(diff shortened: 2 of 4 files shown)


llms/llama/requirements.txt
```diff
@@ -2,3 +2,4 @@ mlx>=0.8.0
 sentencepiece
 torch
 numpy
+tiktoken==0.4.0
```
KE7, line 5: Necessary to pin? Seems the older tokenizer didn't need it

js-ts: Did the old tokenizer had it already installed ?, I added this specific version because it was listed in the requirements of [Llama3](   https://github.com/meta-llama/llama3/blob/bb55334adcedfa9f5da66d2e1ed64e6f3dbd82ed/requirements.txt#L4)

awni:
Thanks for adding this! However, moving forward we are adding new models to MLX LM rather than back-porting the original examples. Llama 3 already works in MLX LM so we should be good there.

llms/llama/requirements.txt
KE7: Well adding `tiktoken` is fine but pinning it to `0.4.0` didn't seem necessary

PR #693, opened 2024-04-19
Fix dequantization

awni:
Broke dequantization in #680 when started quantizing embeddings.


KE7 (approved):
Would be nice to have a unit test that catches this so it doesn't break again in the future but there aren't many tests in general so LGTM

angeloskath (approved):
👍

PR #694, opened 2024-04-19
Support Llama-3-8B-Instruct Model 2 types of EOS tokens - Update utils.py

joshuachak:
To fit with Llama-3-8B-Instruct Model, which have 2 types of EOS tokens ("<|end_of_text|>" and "<|eot_id|>"), I add an optional parameter (eos_token_ids) of generate function to receive an array of EOS tokens for Llama-3-8B-Instruct Model and changed the for-loop break check as "if token in eos_token_ids:". If the function does not receive any eos_token_ids as an input, it is by default using "[tokenizer.eos_token_id]" which keep the function can break for-loop, just like the previous version.

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -9,7 +9,7 @@ import shutil
 import time
 from pathlib import Path
 from textwrap import dedent
-from typing import Any, Callable, Dict, Generator, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Generator, Optional, Tuple, Union, List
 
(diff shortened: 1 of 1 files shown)


awni:
My understanding is that the instruct model uses `<|eot_id|>` not `<|end_of_text|>`. I don't think we need to support multiple end of sentence tokens. The instruct models just need to have the eos token configured properly as they do in the mlx-community: https://huggingface.co/mlx-community/Meta-Llama-3-8B-Instruct-4bit/blob/main/special_tokens_map.json#L10

PR #697, opened 2024-04-19
Add support for logit bias

KE7:
It would be cool to support grammar's at some point like https://github.com/ggerganov/llama.cpp/pull/1773/files but for now this will suffice.


awni (approved):
Waiting for tests to clear.

PR #701, opened 2024-04-20
Configuration-based use of HF hub-hosted datasets for training

chimezie:
Per the title, allow a structured `hf_dataset` YAML configuration parameter for specifying an HF hub-hosted dataset (via `name`) to use with training and the ability to use [datasets](https://huggingface.co/docs/datasets/en/index)' local file system caching, named splits, named configurations (via `configuration`), split slicing syntax for specifying train, validation, and test datasets, etc.

The dataset _feature names_ that correspond to attributes or keys for prompt/completion value pairs in the datasets or those with single pure text values (via `text_feature` in that case) can be specified. 

Added YAML parameters, for example (train on the first 1K in the train split and validate with the last 100 (no test data set):

```yaml
hf_dataset:
  name: "billsum"
  train_split: "train[:1000]"
  valid_split: "train[-100:]"
  prompt_feature: "text"
  completion_feature: "summary"
```

See: [Splits and Configurations](https://huggingface.co/docs/datasets-server/en/configs_and_splits), [billsum](https://huggingface.co/datasets/billsum), & [HF Dataset API](https://huggingface.co/docs/datasets/en/access)


chimezie:
Motivated by need to reproduce #620 with an open dataset

llms/mlx_lm/tuner/datasets.py
```diff
@@ -99,6 +205,6 @@ def load_dataset(args, tokenizer: PreTrainedTokenizer):
         )
     if args.test and len(test) == 0:
         raise ValueError(
-            "Test set not found or empty. Must provide test set for evaluation."
+            "Test set not found or empty. Must provide test set for ev aluation."
```
awni, line 208:
```suggestion
            "Test set not found or empty. Must provide test set for evaluation."
```

llms/mlx_lm/tuner/datasets.py
```diff
@@ -83,12 +84,117 @@ def create_dataset(path: Path, tokenizer: PreTrainedTokenizer = None):
         )
 
 
+class TextHFDataset(Dataset):
+    """
+    A Huggingface dataset for a single blob of text per record
+    https://huggingface.co/docs/datasets/en/access
+    """
+
+    def __init__(self, hf_dataset, tokenizer, text_feature):
```
awni, line 93: You can make `text_feature` configurable in the regular dataset, default it to `text` and then just reuse that class.


Same comment for the completions dataset.

awni (commented):
This is very nice, opens up a lot of datasets without too much complexity. I'd like to simplify the dataset classes (see inline comment). After that, I think we should merge it.

fblissjr:
I like this a lot.

Once everyone finetunes for the first time, they quickly see it's all about the data.  

Mixing private and public HF datasets in the same config would be fantastic as well.

awni (approved):
Thanks for the addition!!

PR #702, opened 2024-04-20
fix(mlx-lm): unicode decoding in server.py

mzbac:
I misunderstood how the new detokenizer works, didn't realize that detokenizer.finalize() would force decoding of the current token, which breaks the unicode decoding that requires more than 1 token.
before:
image
after:
image


awni (approved):
🙏

PR #703, opened 2024-04-21
Load fused model with transformers

AlexandrosChrtn:
Was unable to load the fused model on an ubuntu environment. Was able to load it when I changed save_safetensors metadata to 'pt' (there was a related [GH](https://github.com/ml-explore/mlx/issues/633) issue I believe) and produced model.safetensors.index.json. If this patch is not optimal, would still be better imo to let users know about this incompatibility, until it's fixed.

### DIFF
--- a/lora/fuse.py
+++ b/lora/fuse.py
@@ -48,11 +48,6 @@ if __name__ == "__main__":
         help="Generate a de-quantized model.",
         action="store_true",
     )
-    parser.add_argument(
-        "--save-format",
-        help="Format for saved safetensors. Examples: 'mlx', 'pt', 'tf', 'flax'.",
-        default="mlx",
-    )   
 
     print("Loading pretrained model")
     args = parser.parse_args()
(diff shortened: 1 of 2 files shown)


awni:
I should work if the format is `mlx` https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L510

So can you remove the need to specify the format and just hard code it to MLX?

AlexandrosChrtn:
I can do it for sure, but I couldn't load it with mlx format, I got 
```
ValueError: Incompatible safetensors file. File metadata is not ['pt', 'tf', 'flax'] but mlx
```
transformers==4.38.2 in google colab with
```
import torch
from transformers import AutoModelForCausalLM, AutoTokenizer
MODEL_PATH = "/content/drive/MyDrive/mlx_test/lora_fused_model"
tokenizer = AutoTokenizer.from_pretrained(MODEL_PATH)
model = AutoModelForCausalLM.from_pretrained(
    MODEL_PATH,
    device_map="auto",
    torch_dtype="auto",
    trust_remote_code=True,
    use_safetensors=True
)
```
where lora_fused_model is the fuse.py output with mlx format. Not sure if I'm missing something

awni:
Oh maybe you need to upgrade transformers:

`pip install -U transformers`

awni:
Can you run the pre-commit hooks, looks like your code failed the linting.

AlexandrosChrtn:
transformers==4.40.0 solves it, hardcoding "mlx" is enough for users to load models on linux. The argument is not needed indeed as colab / kaggle / whatever else is out there transitions to transformers==4.40.0 by default.

ps. sorry my bad, missed it, should be good

awni (approved):
Thanks!!

PR #707, opened 2024-04-22
🔢 Add score property to clip.py file

kadirnar:
I added a cosine function to give the CLIP model output to the score 🥳  I also added a logging function instead of print ❤️


awni:
I'm not entirely sure the purpose of this diff / why it's useful here. Could you share more on your thought process for the addition?

awni:
Closing as inactive. Feel free to comment here to discuss this change further if you still plan to add it.

PR #715, opened 2024-04-23
Pad mask with zeros for non-square attention matrices

kevmo314:
The current implementation of the mask assumes the attention matrix is square, which is true if there is no cache. However, if one wishes to produce multiple tokens at a time, such as in speculative decoding implementations, a rectangular mask is necessary.

This change pads the bottom of the mask with zeros so multi-token decoding with a cache works correctly.

### DIFF
--- a/llms/mlx_lm/models/llama.py
+++ b/llms/mlx_lm/models/llama.py
@@ -154,6 +154,9 @@ class LlamaModel(nn.Module):
         mask = None
         if h.shape[1] > 1:
             mask = nn.MultiHeadAttention.create_additive_causal_mask(h.shape[1])
+            if cache is not None:
+                pad_size = cache[0][0].shape[2]
+                mask = mx.pad(mask, ((0, 0), (0, pad_size)))
             mask = mask.astype(h.dtype)
 
         if cache is None:
--- a/llms/mlx_lm/models/llama.py
+++ b/llms/mlx_lm/models/llama.py
@@ -131,6 +131,13 @@ class TransformerBlock(nn.Module):
         return out, cache
 
 
+def create_additive_causal_mask(N: int, offset: int = 0):
+    rinds = mx.arange(offset + N)
+    linds = mx.arange(offset, offset + N) if offset else rinds
+    mask = linds[:, None] < rinds[None]
+    return mask * -1e9
+
+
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/models/llama.py
```diff
@@ -154,6 +154,9 @@ def __call__(
         mask = None
         if h.shape[1] > 1:
             mask = nn.MultiHeadAttention.create_additive_causal_mask(h.shape[1])
+            if cache is not None:
+                pad_size = cache[0][0].shape[2]
+                mask = mx.pad(mask, ((0, 0), (0, pad_size)))
```
awni, line 159: This seems like a useful change but it would be better to unify it across models.

Maybe we can have a new `create_additive_causal_mask` in the `base.py` file: https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/base.py and it takes two variables instead of one? Then we can use it in all the model files instead of `create_additive_causal_mask`?

kevmo314: That seems pretty reasonable to me. Do you have any thoughts on how it's implemented? I noticed the speculative decoding example also includes this behavior however in a different implementation. Happy to work with any implementation if you have any preferences. https://github.com/ml-explore/mlx-examples/blob/main/llms/speculative_decoding/model.py#L256

kevmo314: Actually, would this change be better suited to upstream into mlx itself? https://github.com/ml-explore/mlx/blob/ec8578d41a49d0dd4dc966a2a9e5ef9edca9f172/python/mlx/nn/layers/transformer.py#L102

awni: Oh nice, the one in the spec decoding example looks quite reasonable. I'm not sure if it is faster than doing the square + padding as you have it here. Probably either is fine as it's a very small part of the model.

awni:
@kevmo314 are you coming back to this?

kevmo314:
> @kevmo314 are you coming back to this?

Yes I will come back to this shortly.

kevmo314:
@awni I've updated the implementation!

awni:
Looks like the test failed? Could you take a look?

awni (approved):
Thanks!

PR #717, opened 2024-04-24
Add Supported Quantized Phi-3-mini-4k-instruct gguf Weight

Jaykef:
The official phi-3-mini-4k-instruct gguf weight is currently unsupported (of type array(15, dtype=uint32)) by mlx, I uploaded a supported 4bit quantized gguf weight, download here: https://huggingface.co/Jaward/phi-3-mini-4k-instruct.Q4_0.gguf.

Also included a print of model info and quant type.

Screenshot 2024-04-24 at 13 47 35

### DIFF
--- a/llms/gguf_llm/models.py
+++ b/llms/gguf_llm/models.py
@@ -18,6 +18,7 @@ class ModelArgs:
     num_attention_heads: int
     rms_norm_eps: float
     vocab_size: int
+    context_length: int
     num_key_value_heads: int = None
     rope_theta: float = 10000
     rope_traditional: bool = False
@@ -157,6 +158,20 @@ class LlamaModel(nn.Module):
             TransformerBlock(args=args) for _ in range(args.num_hidden_layers)
         ]
         self.norm = nn.RMSNorm(args.hidden_size, eps=args.rms_norm_eps)
+        # model info
+        print(
+            f"Model info\n"
+            f"==========\n"
+            f"Context length: {args.context_length}\n"
+            f"Hidden size: {args.hidden_size}\n"
+            f"Num layers: {args.num_hidden_layers}\n"
+            f"Num attention heads: {args.num_attention_heads}\n"
+            f"Num key value heads: {args.num_key_value_heads}\n"
+            f"RMSNorm epsilon: {args.rms_norm_eps}\n"
+            f"Vocab size: {args.vocab_size}\n"
+            f"RoPE theta: {args.rope_theta}\n"
+            f"RoPE traditional: {args.rope_traditional}\n"
+        )
 
     def __call__(
         self,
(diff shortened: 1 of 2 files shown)


llms/gguf_llm/models.py
```diff
@@ -18,6 +18,7 @@ class ModelArgs:
     num_attention_heads: int
     rms_norm_eps: float
     vocab_size: int
+    context_length: int
```
awni, line 21: Do the other models in the list have this field?

Jaykef: Yes they all do with no errors during generation.

Screenshot 2024-04-25 at 12 28 27

Jaykef: Is everything okay for a merge now? @awni

awni: Yea I think so. I was wondering about a nit: if we wanted to print all the model info by default.

Jaykef: Since the use-case here is primarily for research I believe having to see relevant model dets is essential, also it's not all the model info, just some relevant ones. I learned that from karpathy's gpt repos:)

Jaykef: Interestingly, the official [Phi-3-mini-4k-instruct-gguf](https://huggingface.co/microsoft/Phi-3-mini-4k-instruct-gguf)'s is using llama architecture, you can see even in the metadata we are printing, prints out because of that.

Screenshot 2024-04-27 at 07 30 50

Jaykef: > Yea I think so. I was wondering about a nit: if we wanted to print all the model info by default.

Interestingly, the official [Phi-3-mini-4k-instruct-gguf](https://huggingface.co/microsoft/Phi-3-mini-4k-instruct-gguf) is using llama architecture, you can see even in the metadata we are printing, prints out because of that.

Screenshot 2024-04-27 at 07 30 50

Jaykef: @awni still no merge?

llms/gguf_llm/models.py
```diff
@@ -265,13 +277,17 @@ def load(gguf_file: str, repo: str = None):
     if gguf_ft == 0 or gguf_ft == 1:
         # ALL_F32 or MOSTLY_F16
         quantization = None
+        print("Non-quantized Float32 model")
```
awni, line 280: This print is not right. It should be either float32 or float16.

awni (commented):
@Jaykef thanks, I left a minor comment. Could you address and then we can merge?

Jaykef:
Okay

Jaykef:
> @Jaykef thanks, I left a minor comment. Could you address and then we can merge?

Addressed

Jaykef:
@awni I have removed that print all together since the last condition already prints warning for when quantization is None.

llms/gguf_llm/models.py
Jaykef: @awni I have deleted that print, since last condition already prints warning for unsupported quants

awni (approved):
Thanks!!

PR #719, opened 2024-04-24
Add support for OpenELM

Blaizzy:
This PR adds support for openELM.
Twitter: @Prince_Canuma

### DIFF
--- a/llms/mlx_lm/models/openelm.py
+++ b/llms/mlx_lm/models/openelm.py
@@ -73,7 +73,7 @@ class Attention(nn.Module):
         self.model_dim = model_dim = args.model_dim
 
         self.n_heads = n_heads = args.num_query_heads[layer_id]
-        self.n_kv_heads = n_kv_heads = args.num_kv_heads[layer_id]s
+        self.n_kv_heads = n_kv_heads = args.num_kv_heads[layer_id]
         self.scale = head_dim**-0.5
 
         op_size = (n_heads + (n_kv_heads * 2)) * head_dim
@@ -114,10 +114,8 @@ class Attention(nn.Module):
         ).transpose(0, 2, 1, 3)
 
         # [B, (q_h + k_h + v_h), S, h] --> [B, q_h, S h], [B, k_h, S, h], [B, v_h, S, h]
-        queries = qkv[:, : self.n_heads, :, :]
-        keys = qkv[:, self.n_heads : (self.n_heads + self.n_kv_heads), :, :]
-        values = qkv[:, (self.n_heads + self.n_kv_heads) :, :, :]
-
+        queries, keys, values = mx.split(qkv, [self.n_heads, self.n_heads + self.n_kv_heads], axis=1)
+        
         # Prepare the queries, keys and values for the attention computation
         if self.normalize_qk_projections:
             queries = self.q_norm(queries)
--- a/llms/mlx_lm/models/openelm.py
+++ b/llms/mlx_lm/models/openelm.py
@@ -74,9 +74,7 @@ class Attention(nn.Module):
             self.k_norm = nn.RMSNorm(head_dim, eps=args.rms_norm_eps)
 
         self.rope = nn.RoPE(
-            head_dim,
-            traditional=args.rope_traditional,
-            base=args.rope_theta
+            head_dim, traditional=args.rope_traditional, base=args.rope_theta
         )
 
     def __call__(
--- a/llms/mlx_lm/models/openelm.py
+++ b/llms/mlx_lm/models/openelm.py
@@ -146,15 +146,15 @@ class MLP(nn.Module):
         dim = args.model_dim
         ffn_multiplier = args.ffn_multipliers[layer_id]
 
-        hidden_dim = int(
+        intermediate_dim = int(
             make_divisible(
                 ffn_multiplier * args.model_dim,
                 divisor=args.ffn_dim_divisor,
             )
         )
 
-        self.proj_1 = nn.Linear(dim, 2 * hidden_dim, bias=False)
-        self.proj_2 = nn.Linear(hidden_dim, dim, bias=False)
+        self.proj_1 = nn.Linear(dim, 2 * intermediate_dim, bias=False)
+        self.proj_2 = nn.Linear(intermediate_dim, dim, bias=False)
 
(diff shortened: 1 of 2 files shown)


Blaizzy:
@awni I couldn't replicate the qkv splitting logic from the transformers implementation.

```python
# [B, (q_h + k_h + v_h), S, h] --> [B, q_h, S h], [B, k_h, S, h], [B, v_h, S, h]
queries, keys, values = qkv.split(
      [self.num_q_heads, self.num_k_heads, self.num_v_heads], dim=1
)
```

Here is what I tried:
```
queries, kv = mx.split(qkv, [self.n_heads], axis=1)
keys , values = mx.split(kv, 2, axis=1)
```

The outputs don't yet make sense with this approach.

Goekdeniz-Guelmez:
have you tried splitting them in one go like: `queries, keys, values = mx.split(qkv, 3, axis=-1)`, why do you have to split them into query and kv and then the kv again into key and value pairs?

awni:
Closes #723

Blaizzy:
> @awni I couldn't replicate the qkv splitting logic from the transformers implementation.
> 
> 
> 
> ```python
> 
> # [B, (q_h + k_h + v_h), S, h] --> [B, q_h, S h], [B, k_h, S, h], [B, v_h, S, h]
> 
> queries, keys, values = qkv.split(
> 
>       [self.num_q_heads, self.num_k_heads, self.num_v_heads], dim=1
> 
> )
> 
> ```
> 
> 
> 
> Here is what I tried:
> 
> ```python
> 
> queries, kv = mx.split(qkv, [self.n_heads], axis=1)
> 
> keys , values = mx.split(kv, 2, axis=1)
> 
> ```
> 
> and
> 
> ```python
> 
> queries = qkv[:, :self.n_heads, :, :]
> 
> kv = qkv[:, self.n_kv_heads:, :, :]
> 
> keys, values = kv
> 
> ```
> 
> 
> 
> However, the outputs don't yet make sense with this approach.


@awni the issue still remains. 

The model loads well but produces nonsensical responses.

I think mx.split works differently from the one they used. 

I'm trying to think of different options but I'm bit stuck.

awni:
You should use split in one go rather than calling it multiple times. See the [docs here](https://ml-explore.github.io/mlx/build/html/python/_autosummary/mlx.core.split.html#mlx.core.split)

As for why the model doesn't generate good output, I'm not sure there. I haven't looked at it in detail. Maybe double check the reference implementation? A good next step if you are stuck is to

1. step through the reference implementation and make sure both models are generating the same outputs up to the same point.
2. Once you find the point at which the outputs differ, that should help you diagnose the problem in the code.

Blaizzy:
@awni thank you very for the advice!

Thank God MLX has similar tensor/array slicing as torch and numpy!

I knew the error was in the kv slicing. I just needed to sitdown and focus :)

It's done now ✅

llms/mlx_lm/tuner/utils.py
```diff
@@ -88,6 +88,8 @@ def linear_to_lora_layers(
         keys = set(["att_proj"])
     elif model.model_type == "phi3":
         keys = set(["self_attn.qkv_proj"])
+    elif model.model_type == "openelm":
```
awni, line 91:
```suggestion
    elif model.model_type in ["phi3", "openelm"]:
        keys = set(["self_attn.qkv_proj"])
```

awni:
Looks great, thanks for the addition! Could you address the two comments I left, then we can merge?

llms/mlx_lm/models/openelm.py
```diff
@@ -0,0 +1,249 @@
...
+
+        # [B, S, (q_h + k_h + v_h) * h] --> [B, S, (q_h + k_h + v_h), h] -> [B, (q_h + k_h + v_h), S, h]
+        qkv = qkv.reshape(
+            B, L, self.n_heads + (self.n_kv_heads * 2), self.head_dim
+        ).transpose(0, 2, 1, 3)
+
+        # [B, (q_h + k_h + v_h), S, h] --> [B, q_h, S h], [B, k_h, S, h], [B, v_h, S, h]
+        queries = qkv[:, : self.n_heads, :, :]
+        keys = qkv[:, self.n_heads : (self.n_heads + self.n_kv_heads), :, :]
+        values = qkv[:, (self.n_heads + self.n_kv_heads) :, :, :]
```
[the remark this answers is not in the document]
Blaizzy: I was looking for exactly that. The MLX.core documentation doesn't explain how it works well, perhaps  some examples would help.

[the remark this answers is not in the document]
Blaizzy: It's done ✅ 

Also, I ran some tests and there is little difference in speed.

llms/mlx_lm/tuner/utils.py
Blaizzy: Done ✅

Blaizzy:
> Looks great, thanks for the addition! Could you address the two comments I left, then we can merge?

Most welcome! This was an interesting architecture.

The comments are resolved ✅

llms/mlx_lm/models/openelm.py
```diff
@@ -0,0 +1,247 @@
...
+    def __post_init__(self):
+        if self.rope_scaling:
+            required_keys = {"long_factor", "type"}
+            if not all(key in self.rope_scaling for key in required_keys):
+                raise ValueError(f"rope_scaling must contain keys {required_keys}")
+
+            if self.rope_scaling["type"] != "linear":
+                print(
+                    "[WARNING] rope_scaling 'type' currently only supports 'linear' setting rope scaling to false."
+                )
```
awni, line 38: What is the purpose of this? I don't think this model uses rope scaling, so could you remove the stuff on rope scaling and just default it to 1.0?

Blaizzy: Yes, will do !

I thought the logic defaults to 1 anyways.

Blaizzy: Done ✅

awni:
Do you mind to run the pre-commit hooks? Looks like the linter failed.

Blaizzy:
Yeah, it's because I removed scale argment.

It's done ✅

awni (approved):
Thanks!!

PR #729, opened 2024-04-26
Upload uses multi_commits when uploading to HF

versae:
This PR uses an experimental feature in the HuggingFace API that creates multiple commits per upload. It is very convenient for big models with multiple files. It also allows for resuming the upload if something fails by simply re-running the command. I've tested it with `convert`, it looks [like this](https://huggingface.co/mlx-community/zephyr-orpo-141b-A35b-v0.1-8bit/discussions/2).


awni (approved):
Nice and simple. Hopefully we can just stick to this moving forward.

KE7 (approved):
lgtm & nice improvement

PR #731, opened 2024-04-26
Validate server params & fix logit bias bug

KE7:
Assumed a diff format for logit bias previously. This one is test on llama-3-8b and working. Also added parameter validations to the server so that user's get error messages at the api level instead of having to check the server logs

image


llms/mlx_lm/utils.py
```diff
@@ -136,7 +136,10 @@ def generate_step(
     """
 
     def sample(logits: mx.array) -> Tuple[mx.array, float]:
-        logits = logits + logit_bias if logit_bias else logits
+        if logit_bias:
+            indicies = mx.array(list(logit_bias.keys()))
+            values = mx.array(list(logit_bias.values()))
+            logits[:, indicies] += values
```
awni, line 142: indicies -> indices

llms/mlx_lm/server.py
```diff
@@ -159,6 +161,40 @@ def do_POST(self):
         method = self.handle_stream if self.stream else self.handle_completion
         method(prompt, stop_id_sequences)
 
+    def validate_model_parameters(self):
```
awni, line 164: This is nice.

llms/mlx_lm/server.py
```diff
@@ -159,6 +161,40 @@ def do_POST(self):
...
+        """
+        Validate the model parameters passed in the request for the correct
+            types and values
```
awni, line 167:
```suggestion
        Validate the model parameters passed in the request for the correct
        types and values
```

llms/mlx_lm/server.py
```diff
@@ -159,6 +161,40 @@ def do_POST(self):
...
+        if not isinstance(self.repetition_context_size, int) or self.repetition_context_size < 0:
+            raise ValueError("repetition_context_size must be a non-negative integer")
+        
+        if not isinstance(self.logit_bias, dict):
+            raise ValueError("logit_bias must be a dict of int to float")
+        
+        try:
+            self.logit_bias = {int(k): v for k, v in self.logit_bias.items()}
+        except ValueError:
+            raise ValueError("logit_bias must be a dict of int to float")
```
awni, line 193: What's the purpose of this? I don't think it makes sense to allow strings here does it?

awni (commented):
Looks nice, thanks! I left a few minor comments. Could you please address then we can merge?

llms/mlx_lm/server.py
KE7: It's to support those who may call through postman:
image

llms/mlx_lm/server.py
KE7: I had the tab to be consistent with other docstrings in the same file. For example, https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/server.py#L171

awni: I think the style we usually use in MLX is that this should be 1 line and end in a period. The others should be fixed as well.

awni:
Could you check the failed tests? Also please run the pre-commit hooks for formatting as that also failed.

awni (approved):
Thanks!!

PR #735, opened 2024-04-28
Update lora_config.yaml

jincdream:
Update LoRa config YAML, replacing the adapter file argument with the adapter path argument.


awni (approved):
Thanks!!

PR #739, opened 2024-04-29
Whisper: Add pip distribution configuration to support pip installations.

madroidmaq:
Currently the whisper module can only be used in the current project, and there are some difficulties when other projects want to build on it.
For example, some of the possible actions are to copy the source code into their own project or to publish it in a separate package (e.g. [lightning-whisper-mlx](https://github.com/mustafaaljadery/lightning-whisper-mlx)), which will make it difficult for them to benefit from the changes made to this repository when the whisper module is updated. is updated, it is difficult for them to take advantage of the benefits of the whisper module.

In addition to that, it could be used as a separate command-line tool, like openai whisper.

- install

```
pip install mlx_whisper
```

- use
```
import mlx_whisper

result = mlx_whisper.transcribe(
    audio="audio.mp3",
    path_or_hf_repo="mlx-community/whisper-large-v3-mlx-8bit",
)

print(result["text"])
```

tasks:
- [x] Add setup.py config 
- [x] Update readme.md


madroidmaq:
@awni  Currently my local release works, but there are some paths in the code that are written dead, so there are changes that need to be made if I want to use it. I'd like to know if this part of the current PR changes will be accepted so I can decide if I want to go ahead and commit.

awni:
> but there are some paths in the code that are written dead,

What do you mean by that?

I think we could accept this change, it's relatively straightforward and seems pretty usefule. We'd (as in MLX maintainers) should manage the PyPi pacakage directly if we are going to distribute it through MLX Examples, so I will put it on PyPi once this is merged.

madroidmaq:
> What do you mean by that?

@awni  After I release mlx_whisper and run the code in another directory, I encounter the following problem:

```
python mlx-w.py
  File "/Users/madroid/Desktop/pdf/mlx-w.py", line 4, in <module>
    result = transcribe(
             ^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/mlx_whisper/transcribe.py", line 150, in transcribe
    mel = log_mel_spectrogram(audio, n_mels=model.dims.n_mels, padding=N_SAMPLES)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/mlx_whisper/audio.py", line 167, in log_mel_spectrogram
    filters = mel_filters(n_mels)
              ^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/mlx_whisper/audio.py", line 96, in mel_filters
    return mx.load(filename)[f"mel_{n_mels}"]
           ^^^^^^^^^^^^^^^^^
ValueError: [load_npz] Input must be a zip file or a file-like object that can be opened with zipfile.ZipFile
```

After I located it, I found that the error occurred in the access to the ` mel_filters.npz` file in the audio.py file as follows:
https://github.com/ml-explore/mlx-examples/blob/5513c4e57dd78bf0d929b49ea2818ee0ac9644b2/whisper/whisper/audio.py#L95

Maybe I should package the contents of the `assets` file as well, but files like `ls_test.flac` don't seem to need to be packaged either. I'm not familiar with this part, can you give some advice?

awni:
> Maybe I should package the contents of the assets file as well, but files like ls_test.flac don't seem to need to be packaged either. I'm not familiar with this part, can you give some advice?

Yes I think we'd need to package the mel filters. I think you can use either `MANIFEST.in` or `package_data` to include it in the distribution, I'm not sure which is the preferred route these days 😅

madroidmaq:
I chose the `package_data` approach because it looks like he modified less. Now the local package can be run successfully from any directory.

awni (approved):
Thanks @madroidmaq looks great. I will upload a distribution shortly!

PR #752, opened 2024-05-04
chore(mlx-lm): fix the number of validation batches configuration.

mzbac:
Fixing the issue where setting `-1` for the number of validations does not evaluate the full validation set.

### DIFF
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -137,8 +137,11 @@ def evaluate(
 ):
     all_losses = []
     ntokens = 0
+
+    index_iterator = iter(range(num_batches)) if num_batches != -1 else iter(int, 1)
+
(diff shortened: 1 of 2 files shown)


llms/mlx_lm/tuner/trainer.py
```diff
@@ -137,8 +137,11 @@ def evaluate(
 ):
     all_losses = []
     ntokens = 0
+
+    index_iterator = iter(range(num_batches)) if num_batches != -1 else iter(int, 1)
+
     for it, batch in zip(
-        range(num_batches),
+        index_iterator,
```
awni, line 144: Nit: might be good to have `it` be `_` since it's not a valid counter

awni (approved):
Nice catch, thanks for the fix and the test! I left a very minor comment o/w we can merge shortly.

PR #770, opened 2024-05-10
Add `model_config` parameter to `load()` and `load_model()`

JosefAlbers:
For easy editing of the loaded model configuration (e.g., increasing RoPE base frequency or PI). Example:

```python
model, tokenizer = load("mlx-community/Phi-3-mini-4k-instruct-4bit-no-q-embed", model_config={"rope_theta":50000.0})
response = generate(model, tokenizer, prompt)
```

Output with default rope_theta of 10000.0:
image

Output with modified rope_theta:
image

### DIFF
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -64,7 +64,7 @@ class TrainingArgs:
 
 
 def default_loss(model, inputs, targets, lengths):
-    logits, _ = model(inputs)
+    logits = model(inputs)
     logits = logits.astype(mx.float32)
 
     length_mask = mx.arange(inputs.shape[1])[None, :] < lengths[:, None]
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -64,7 +64,7 @@ class TrainingArgs:
 
 
 def default_loss(model, inputs, targets, lengths):
-    logits = model(inputs)
+    logits, _ = model(inputs)
     logits = logits.astype(mx.float32)
 
     length_mask = mx.arange(inputs.shape[1])[None, :] < lengths[:, None]
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -299,17 +299,21 @@ def load_config(model_path: Path) -> dict:
     return config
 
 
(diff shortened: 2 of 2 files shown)


awni:
I'm a little confused by that example. Shouldn't the RoPE theta in the config be the correct value? If its not the correct value, we should fix it in the original config right?

JosefAlbers:
> I'm a little confused by that example. Shouldn't the RoPE theta in the config be the correct value? If its not the correct value, we should fix it in the original config right?

I apologize for any confusion caused by the example. It was intended to illustrate the potential utility of the proposed `model_config` parameter. LLMs often produce nonsensical output as they approach their pretrained context limit, but techniques like position interpolation or adjusting base frequency with RoPE can extend this limit without fine-tuning. I believe that the proposed change may facilitate further experimentation and development of ways to enhance the usefulness of LLMs by the users.

P.S. I also apologize for the clutter I caused with the commit and revert actions for an unrelated issue in this thread.

awni:
>  LLMs often produce nonsensical output as they approach their pretrained context limit, but techniques like position interpolation or adjusting base frequency with RoPE can extend this limit without fine-tuning. 

That makes sense! Thanks for the explanation

> P.S. I also apologize for the clutter I caused with the commit and revert actions for an unrelated issue in this thread.

No problem. It looks like the fix is actually needed so you would be great to add it here or send a new PR with it. O/w I will add it shortly.

aquiferer:
> > LLMs often produce nonsensical output as they approach their pretrained context limit, but techniques like position interpolation or adjusting base frequency with RoPE can extend this limit without fine-tuning.
> 
> That makes sense! Thanks for the explanation
> 
> > P.S. I also apologize for the clutter I caused with the commit and revert actions for an unrelated issue in this thread.
> 
> No problem. It looks like the fix is actually needed so you would be great to add it here or send a new PR with it. O/w I will add it shortly.

Thanks so much, will do!

JosefAlbers:
Thanks so much, will do!

awni:
Looks like some tests failed. Would you mind checking @JosefAlbers ?

JosefAlbers:
Sorry for the inconvenience,

1. Moved the new `model_config` argument to the end.
2. Fixed styling (`black`).

awni (approved):
Thank you!

PR #775, opened 2024-05-11
fix: Added dedicated error handling to load and get_model_path

AtakanTekparmak:
Added proper error handling to `load` and `get_model_path` by adding a dedicated exception class, because when the local path is not right, it still throws the huggingface `RepositoryNotFoundError`, which is not fully informative to the end user, they might have just entered their local path wrong by one character.

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -14,6 +14,7 @@ from typing import Any, Callable, Dict, Generator, Optional, Tuple, Union
 import mlx.core as mx
 import mlx.nn as nn
 from huggingface_hub import snapshot_download
+from huggingface_hub.utils._errors import RepositoryNotFoundError
 from mlx.utils import tree_flatten
 from transformers import AutoTokenizer, PreTrainedTokenizer
 
@@ -90,7 +91,7 @@ def get_model_path(path_or_hf_repo: str, revision: Optional[str] = None) -> Path
                 )
             )
         except RepositoryNotFoundError:
-            raise RepoNotFoundError(f"No local or Hugging Face repository found for path: {path_or_hf_repo}.")
+            raise RepoNotFoundError(f"Model Not Found for {path_or_hf_repo}.\n Please make sure you specified the local path or Hugging Face repo id correctly.\n If you are trying to access a private or gated Hugging Face repo, make sure you are authenticated.")
     return model_path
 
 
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/utils.py
```diff
@@ -397,7 +405,10 @@ def load(
         FileNotFoundError: If config file or safetensors are not found.
         ValueError: If model class or args class are not found.
     """
-    model_path = get_model_path(path_or_hf_repo)
+    try:
+        model_path = get_model_path(path_or_hf_repo)
+    except RepoNotFoundError:
+        raise RepoNotFoundError(f"No local or Hugging Face repository found for path: {path_or_hf_repo}. Please check the provided path again.")
```
awni, line 411: Why do we need to wrap it twice (once here and once above)?

llms/mlx_lm/utils.py
```diff
@@ -69,20 +74,23 @@ def get_model_path(path_or_hf_repo: str, revision: Optional[str] = None) -> Path
...
+                        "*.py",
+                        "tokenizer.model",
+                        "*.tiktoken",
+                        "*.txt",
+                    ],
+                )
             )
-        )
+        except RepositoryNotFoundError:
+            raise RepoNotFoundError(f"No local or Hugging Face repository found for path: {path_or_hf_repo}.")
```
awni, line 93: Couldn't this fail for other reasons? TBH I'm not sure what the goal of this new exception is. Maybe you can clarify the purpose?

AtakanTekparmak: In the case where the user tries to load a local model with the name `example_model_name` and makes a typo (for example `example_model-name`, the code still throws the huggingface's `RepositoryNotFoundError`error, with the following error message:

```bash
Repository Not Found for url: https://huggingface.co/api/models/llama3-17b-instruc/revision/main.
Please make sure you specified the correct `repo_id` and `repo_type`.
If you are trying to access a private or gated repo, make sure you are authenticated.
```

for the following line of code:
```python
model, tokenizer = load("llama3-17b-instruc")
```
which is not very informative, as the user is trying to access a local model. Not essential by any means, just something i encountered my first time using the package.

awni: I see. Maybe it is worth augmenting that error message instead because it's possible they intended to download from HF in which case the message is actually useful.

Something like:

```
Model Not Found for {model_path}.
Please make sure you specified the local path or Hugging Face repo id correctly.
If you are trying to access a private or gated Hugging Face repo, make sure you are authenticated.
```

AtakanTekparmak: This looks good, are you going to add the edits or should I?

awni: Would you mind making them?

awni (approved):
Thanks!!

PR #778, opened 2024-05-12
Update LORA.md

konarkm:
Add API generation documentation for when user wants to generate using fine tuned adapters.

### DIFF
--- a/llms/mlx_lm/LORA.md
+++ b/llms/mlx_lm/LORA.md
@@ -95,6 +95,28 @@ mlx_lm.generate \
     --prompt "<your_model_prompt>"
 ```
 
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/LORA.md
```diff
@@ -95,6 +95,28 @@ mlx_lm.generate \
     --prompt "<your_model_prompt>"
 ```
 
+For generation using the API:
```
awni, line 98: This doesn't look like the completion or chat format we use for LoRA. Shouldn't we follow one of those formats here (and specify it)?

awni:
Closing as inactive

PR #782, opened 2024-05-16
Block sparse MM MoEs

angeloskath:
Implements all MoEs with the new `mx.block_sparse_mm` and `mx.block_sparse_qmm` (the latter is at https://github.com/ml-explore/mlx/pull/1124 ).

One notable change is in the quantization predicate which accepts modules that define `to_quantized()` and delegates the check for the existence of quantized weights to the module using `is_quantized()`. This allows us to define `SwitchMLP` for each model and a quantized equivalent and auto convert to the quantized one.

### Performance comparison at 4bits

Before
```
Qwen1.5-MoE-A2.7B-Chat     : 56.3 tps
Mixtral-8x7B-Instruct-v0.1 : 40.3 tps
phixtral-4x2_8             : 47.2 tps
```

After
```
Qwen1.5-MoE-A2.7B-Chat     : 121.7 tps
Mixtral-8x7B-Instruct-v0.1 :  59.3 tps
phixtral-4x2_8             :  92.8 tps
```

### DIFF
--- a/llms/mlx_lm/tuner/utils.py
+++ b/llms/mlx_lm/tuner/utils.py
@@ -9,8 +9,9 @@ import mlx.nn as nn
 import mlx.optimizers as opt
 from mlx.utils import tree_unflatten
 
+from ..models.switch_layers import QuantizedSwitchLinear, SwitchLinear
 from .dora import DoRALinear
-from .lora import LoRALinear
+from .lora import LoRALinear, LoRASwitchLinear
 
 
 def build_schedule(schedule_config: Dict):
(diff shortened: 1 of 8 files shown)


angeloskath:
@awni I think this is ready for a review. I didn't implement the `DoRASwitchLinear` but it can easily be added in a later PR that also adds quantization to DoRA layers.

llms/mlx_lm/tuner/utils.py
```diff
@@ -58,11 +59,17 @@ def linear_to_lora_layers(
...
-    cls = DoRALinear if use_dora else LoRALinear
 
-    def to_lora(lin):
-        return cls.from_linear(
-            lin,
+    def to_lora(layer):
+        if isinstance(layer, (nn.Linear, nn.QuantizedLinear)):
+            LoRALayer = DoRALinear if use_dora else LoRALinear
+        elif isinstance(layer, (SwitchLinear, QuantizedSwitchLinear)):
+            LoRALayer = LoRASwitchLinear
```
awni, line 67: Should we add `use_dora` to the condition here? Or throw if trying to DoRA-fy a switch linear layer (until we support Dora)?

awni (approved):
🚀  thanks for fixing the quantized models!

PR #789, opened 2024-05-20
[mlx-whisper] Use namespace package to get rid of package install warning

### LINKED ISSUE #788: Package 'mlx_whisper.assets' is absent from the `packages` configuration
Installing, `whisper` by `pip install . -v` I get:
```
:
  writing manifest file 'mlx_whisper.egg-info/SOURCES.txt'
  /opt/homebrew/Caskroom/miniconda/base/envs/mlx/lib/python3.12/site-packages/setuptools/command/build_py.py:207: _Warning: Package 'mlx_whisper.assets' is absent from the `packages` configuration.
  !!

          ********************************************************************************
          ############################
          # Package would be ignored #
          ############################
          Python recognizes 'mlx_whisper.assets' as an importable package[^1],
          but it is absent from setuptools' `packages` configuration.

          This leads to an ambiguous overall configuration. If you want to distribute this
          package, please make sure that 'mlx_whisper.assets' is explicitly added
          to the `packages` configuration field.

          Alternatively, you can also rely on setuptools' discovery methods
          (for example by using `find_namespace_packages(...)`/`find_namespace:`
          instead of `find_packages(...)`/`find:`).

          You can read more about "package discovery" on setuptools documentation page:

          - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

          If you don't want 'mlx_whisper.assets' to be distributed and are
          already explicitly excluding 'mlx_whisper.assets' via
          `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,
          you can try to use `exclude_package_data`, or `include-package-data=False` in
          combination with a more fine grained `package-data` configuration.

          You can read more about "package data files" on setuptools documentation page:

          - https://setuptools.pypa.io/en/latest/userguide/datafiles.html


          [^1]: For Python, any directory (with suitable naming) can be imported,
                even if it does not contain any `.py` files.
                On the other hand, currently there is no concept of package data
                directory, all directories are treated like packages.
          ********************************************************************************

  !!
    check.warn(importable)
  /opt/homebrew/Caskroom/miniconda/base/envs/mlx/lib/python3.12/site-packages/setuptools/_distutils/cmd.py:66: SetuptoolsDeprecationWarning: setup.py install is deprecated.
  !!

          ********************************************************************************
          Please avoid running ``setup.py`` directly.
          Instead, use pypa/build, pypa/installer or other
          standards-based tools.

          See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
          ********************************************************************************

  !!
    self.initialize_options()
  installing to build/bdist.macosx-11.1-arm64/wheel
:
```

No idea whether this is intentional or not.

awni:
Closes #788

### DIFF
--- a/whisper/setup.py
+++ b/whisper/setup.py
@@ -3,7 +3,7 @@
 import sys
 from pathlib import Path
 
-from setuptools import find_packages, setup
+from setuptools import find_namespace_packages, setup
 
 package_dir = Path(__file__).parent / "mlx_whisper"
 
@@ -26,7 +26,7 @@ setup(
     url="https://github.com/ml-explore/mlx-examples",
     license="MIT",
     install_requires=requirements,
-    packages=find_packages(),
+    packages=find_namespace_packages(),
     include_package_data=True,
     python_requires=">=3.8",
 )


angeloskath (approved):
👍

PR #790, opened 2024-05-21
Add support for Phi-3 Medium

Blaizzy:
This PR adds support for Phi-3 Medium.
Twitter: @Prince_Canuma

### DIFF
--- a/llms/mlx_lm/models/phi3.py
+++ b/llms/mlx_lm/models/phi3.py
@@ -44,8 +44,9 @@ class Attention(nn.Module):
         dim = args.hidden_size
         self.n_heads = n_heads = args.num_attention_heads
         self.n_kv_heads = n_kv_heads = args.num_key_value_heads
+        self.num_hidden_layers = args.num_hidden_layers
 
-        head_dim = args.hidden_size // n_heads
+        self.head_dim = head_dim = args.hidden_size // n_heads
         self.scale = head_dim**-0.5
 
         op_size = n_heads * head_dim + 2 * (n_kv_heads * head_dim)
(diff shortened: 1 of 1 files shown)


Blaizzy:
@awni for the Phi-3 small I need time and probably a bit of help :)

llms/mlx_lm/models/phi3.py
```diff
@@ -73,7 +74,12 @@ def __call__(
         B, L, D = x.shape
 
         qkv = self.qkv_proj(x)
-        queries, keys, values = mx.split(qkv, 3, axis=-1)
+        if self.num_hidden_layers == 40:
+            query_pos = self.n_heads * self.head_dim
+            queries, keys, values = mx.split(qkv, [query_pos, query_pos + self.n_kv_heads * self.head_dim], axis=-1)
+        else:
+            queries, keys, values = mx.split(qkv, 3, axis=-1)
```
awni, line 81: I think you should just remove the condition on the number of layers and use the first split. It will give you the right thing when the `self.n_kv_heads == self.n_heads`.

awni:
> @awni for the Phi-3 small I need time and probably a bit of help :)

What's up with that model? It's not working?

llms/mlx_lm/models/phi3.py
Blaizzy: Read my mind!

It was a late this side to test it.

But I made the change this morning :)

Blaizzy:
> > @awni for the Phi-3 small I need time and probably a bit of help :)
> 
> What's up with that model? It's not working?

It's a complex implementation to map to MLX.

It has kernels for their blocksparse layers built with triton and torch.

- https://huggingface.co/microsoft/Phi-3-small-8k-instruct/blob/main/triton_blocksparse_attention_layer.py
- https://huggingface.co/microsoft/Phi-3-small-8k-instruct/blob/main/triton_flash_blocksparse_attn.py

awni:
Huh the Phi-2 small is quite different than the other two models. I haven't seen that block sparse attention before 🤔

Blaizzy:
Yap, it's completely different.

We can merge this PR for now.

And with time I will take few stabs at the Phi-3 Small. I will need your help, together we can potentially port it to MLX and learn a lot.

Note: This might take a while.

awni (approved):

PR #794, opened 2024-05-23
Port of phi3small

### LINKED ISSUE #805: link to Phi-2 example in readme broken
404 not found when url is clicked

awni:
All pretty basic so far except the block sparse layers.

Also closes #805

### DIFF
--- a/llms/mlx_lm/models/phi3small.py
+++ b/llms/mlx_lm/models/phi3small.py
@@ -23,24 +23,8 @@ class ModelArgs(BaseModelArgs):
     mup_use_scaling: bool = True
     mup_embedding_multiplier: float = 10.0
     mup_width_multiplier: float = 8.0
-    rope_theta: float = 10000
-    rope_traditional: bool = False
-    rope_scaling: Optional[Dict[str, Union[float, str]]] = None
-
-    def __post_init__(self):
-        if self.num_key_value_heads is None:
-            self.num_key_value_heads = self.num_attention_heads
-
-        if self.rope_scaling:
-            required_keys = {"long_factor", "type"}
-            if not all(key in self.rope_scaling for key in required_keys):
-                raise ValueError(f"rope_scaling must contain keys {required_keys}")
-
-            if self.rope_scaling["type"] != "linear":
-                print(
-                    "[WARNING] rope_scaling 'type' currently only supports 'linear' setting rope scaling to false."
-                )
-                self.rope_scaling = None
+    rope_embedding_base: float = 1000000
+    rope_position_scale: float = 1.0
 
 
 def quick_gelu(x):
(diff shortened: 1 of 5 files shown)


awni:
Model generates fine now:
```
python -m mlx_lm.generate --model microsoft/Phi-3-small-8k-instruct --prompt "What about solving an 2x + 3 = 7 equation?" --temp 0.0 --max-tokens 512 --trust-remote-code
```

Outputs:
```
==========
Prompt: <|endoftext|><|user|>
What about solving an 2x + 3 = 7 equation?<|end|>
<|assistant|>

To solve the equation 2x + 3 = 7, follow these steps:

Step 1: Subtract 3 from both sides of the equation.
2x + 3 - 3 = 7 - 3
2x = 4

Step 2: Divide both sides of the equation by 2.
2x / 2 = 4 / 2
x = 2

So, the solution to the equation 2x + 3 = 7 is x = 2.<|end|>

==========
Prompt: 46.940 tokens-per-sec
Generation: 34.704 tokens-per-sec
```

But it will be wrong once the context gets big enough where the sparse attention makes a difference.

Blaizzy:
Let's Go! 🚀

Blaizzy:
I tried a  similar approach yesterday.

I thought we could potentially ignore those block sparse code and use normal attention, I'm glad to see it working!

awni:
Yea exactly. For short sequences the block sparsity is not in effect. For longer sequences though this is probably not going to work.

Blaizzy:
That's great news! 

Have you found the context cutoff where block sparse attn is needed? Also what got you to that realisation?  

I haven't started looking into it deeply yet, was finishing a release.

Blaizzy:
> phi-3-small uses alternative layers of dense attention and a novel blocksparse attention to further optimize on KV cache savings while maintaining long context retrieval performance. - https://arxiv.org/pdf/2404.14219


Got it...

awni:
@angeloskath @jagrit06 I think we can merge this and I will update to use block_masked_mm when it is faster and to remove the `dense_mask` if we add a fill_value to `block_masked_mm`.

llms/mlx_lm/models/phi3small.py
```diff
@@ -0,0 +1,318 @@
...
+        # TODO get rid of dense mask if we have a fill value
+        block_mask, dense_mask = self._block_sparse_mask(L, keys.shape[-2])
+        scores = queries @ mx.swapaxes(keys, -1, -2)
+        # TODO, uncomment when faster
+        # scores = mx.block_masked_mm(
+        #   queries,
+        #   mx.swapaxes(keys, -1, -2),
+        #   mask_out=block_mask,
+        #   block_size=self.blocksparse_block_size,
+        # )
```
awni, line 140: @jagrit06 see comment here

angeloskath (approved):
🚀

PR #800, opened 2024-05-28
feat: Add support for force downloading the model from Hugging Face

malibayram:
This commit adds a new command-line option for `lora/convert.py` as `--force-download` to the `fetch_from_hub` function in `utils.py`. When this option is provided, the model will be forcefully downloaded from Hugging Face, even if it already exists locally. This can be useful when there are updates to the model that need to be fetched.

### DIFF
--- a/clip/convert.py
+++ b/clip/convert.py
@@ -63,7 +63,7 @@ def save_weights(save_path: Union[str, Path], weights: Dict[str, Any]) -> None:
         )
 
 
-def get_model_path(path_or_hf_repo: str) -> Path:
+def get_model_path(path_or_hf_repo: str, force_download: bool = False) -> Path:
     model_path = Path(path_or_hf_repo)
     if not model_path.exists():
         model_path = Path(
@@ -74,6 +74,7 @@ def get_model_path(path_or_hf_repo: str) -> Path:
                     "*.json",
                     "*.txt",
                 ],
+                force_download=force_download,
             )
         )
     return model_path
@@ -107,9 +108,15 @@ if __name__ == "__main__":
         type=str,
         default="float32",
     )
+    parser.add_argument(
+        "-f",
+        "--force-download",
+        help="Force download the model from Hugging Face.",
+        action="store_true",
+    )
     args = parser.parse_args()
 
-    torch_path = get_model_path(args.hf_repo)
+    torch_path = get_model_path(args.hf_repo, args.force_download)
     mlx_path = Path(args.mlx_path)
     mlx_path.mkdir(parents=True, exist_ok=True)
 


awni:
> snapshot_download() downloads the latest revision by default. If you want a specific repository revision, use the revision parameter:

That's from the [docs on snapshot_download](https://huggingface.co/docs/huggingface_hub/en/guides/download#download-an-entire-repository). Shouldn't that update the files to the latest version if they are out of data?

malibayram:
Thank you for pointing out the existing functionality of `snapshot_download`. You are correct that it updates files to the latest version if they are outdated.

However, I believe there are specific scenarios where an explicit `--force-download` option is beneficial:
- **Cache Corruption:** Local caches might occasionally become corrupted or inconsistent. The `--force-download` option ensures a fresh copy is downloaded, bypassing any potential issues with the existing cache.
- **Manual Control:** This option provides users with explicit control to refresh their local copy, regardless of its current state.
- **Consistency Across Tools:** Other tools and scripts in the ecosystem use a similar `--force-download` parameter, so maintaining consistency in user experience is valuable.

Considering these points, I think retaining the `--force-download` option adds robustness and flexibility for users.

awni:
> (Actually, I faced this problem today and that is why I needed to do this change on my fork)

Interesting. Could you say more about that? What exactly was the problem / how did the file cache get corrupted?

malibayram:
To elaborate, I encountered an issue when I initially downloaded the meta-llama--Meta-Llama-3-8B-Instruct model using lora/convert.py. Although I don't recall the exact details of the exceptions I received, I noticed that some files in my cache folder were smaller than expected, as shown in the terminal under `/Users/users/.cache/huggingface/hub/models--meta-llama--Meta-Llama-3-8B-Instruct.`

Uncertain of the cause, I decided to use the --force-download option to re-download the model. This resolved the issue, resulting in a consistent cache folder structure with files matching the expected sizes.

The --force-download option provides a crucial safeguard against such issues, allowing users to recover from incomplete or corrupted downloads by fetching a fresh copy directly from Hugging Face. This feature is particularly valuable in environments where reliability is paramount, ensuring that users always work with an uncorrupted, up-to-date model.

The current model cache folder tree is:

```shell
├── blobs
│   ├── 0fd8120f1c6acddc268ebc2583058efaf699a771
│   ├── 1bfd1146b7bde6168de7bb673bfe0acaea6e684c
│   ├── 31553d63b8f18df5d6d440860751533ada3759e4
│   ├── 3acdd690e65c24f42a24581b8467af98bd3ca357444580f8012aacd2bd607921
│   ├── 65f16bb074ab2ba456331a8ea3d50c3fdbf343bf
│   ├── 67e9ad31c8c32abf3a55ee7fc7217b3ecb35fd3c74d98a5bd233e0e4d6964f46
│   ├── 8675015d20a4b12ee2ee0197a0937beba318122a
│   ├── 8d4782b4a69ef03845159ce1a15e272aadaaf134dc138d68f616098e8531729c
│   ├── b197f72effb9d5ed16ee0f5663e11e4cfac2ba62
│   ├── d8cd5076496dbe4be2320312abc10adc43097b81
│   └── d8cf9c4d0dd972e1a2131bfe656235ee98221679711a3beef6d46dadf0f20b5c
├── refs
│   └── main
└── snapshots
    ├── c4a54320a52ed5f88b7a2f84496903ea4ff07b45
    │   ├── config.json -> ../../blobs/31553d63b8f18df5d6d440860751533ada3759e4
    │   ├── generation_config.json -> ../../blobs/8675015d20a4b12ee2ee0197a0937beba318122a
    │   ├── model-00001-of-00004.safetensors -> ../../blobs/d8cf9c4d0dd972e1a2131bfe656235ee98221679711a3beef6d46dadf0f20b5c
    │   ├── model-00002-of-00004.safetensors -> ../../blobs/8d4782b4a69ef03845159ce1a15e272aadaaf134dc138d68f616098e8531729c
    │   ├── model-00003-of-00004.safetensors -> ../../blobs/3acdd690e65c24f42a24581b8467af98bd3ca357444580f8012aacd2bd607921
    │   ├── model-00004-of-00004.safetensors -> ../../blobs/67e9ad31c8c32abf3a55ee7fc7217b3ecb35fd3c74d98a5bd233e0e4d6964f46
    │   ├── model.safetensors.index.json -> ../../blobs/0fd8120f1c6acddc268ebc2583058efaf699a771
    │   ├── original
    │   │   └── params.json -> ../../../blobs/65f16bb074ab2ba456331a8ea3d50c3fdbf343bf
    │   ├── special_tokens_map.json -> ../../blobs/d8cd5076496dbe4be2320312abc10adc43097b81
    │   ├── tokenizer.json -> ../../blobs/b197f72effb9d5ed16ee0f5663e11e4cfac2ba62
    │   └── tokenizer_config.json -> ../../blobs/1bfd1146b7bde6168de7bb673bfe0acaea6e684c
    └── e1945c40cd546c78e41f1151f4db032b271faeaa
        ├── config.json -> ../../blobs/31553d63b8f18df5d6d440860751533ada3759e4
        ├── generation_config.json -> ../../blobs/8675015d20a4b12ee2ee0197a0937beba318122a
        ├── model-00001-of-00004.safetensors -> ../../blobs/d8cf9c4d0dd972e1a2131bfe656235ee98221679711a3beef6d46dadf0f20b5c
        ├── model-00002-of-00004.safetensors -> ../../blobs/8d4782b4a69ef03845159ce1a15e272aadaaf134dc138d68f616098e8531729c
        ├── model-00003-of-00004.safetensors -> ../../blobs/3acdd690e65c24f42a24581b8467af98bd3ca357444580f8012aacd2bd607921
        ├── model-00004-of-00004.safetensors -> ../../blobs/67e9ad31c8c32abf3a55ee7fc7217b3ecb35fd3c74d98a5bd233e0e4d6964f46
        ├── model.safetensors.index.json -> ../../blobs/0fd8120f1c6acddc268ebc2583058efaf699a771
        ├── original
        │   └── params.json -> ../../../blobs/65f16bb074ab2ba456331a8ea3d50c3fdbf343bf
        ├── special_tokens_map.json -> ../../blobs/d8cd5076496dbe4be2320312abc10adc43097b81
        ├── tokenizer.json -> ../../blobs/b197f72effb9d5ed16ee0f5663e11e4cfac2ba62
        └── tokenizer_config.json -> ../../blobs/1bfd1146b7bde6168de7bb673bfe0acaea6e684c
```

malibayram:
> It seems like it'd be better to verify the model's size/hash and output an error message and/or automatically resume the download. If this is part of huggingface, maybe upstream a PR to improve robustness? Data integrity isn't really in scope for mlx, ne?

Given these points, while verifying data integrity and implementing automatic resume functionality (and this function deprecated in huggingface codes) are excellent ideas, the --force-download option provides a simple and effective solution for ensuring users can recover from a variety of potential issues quickly. It adds robustness and flexibility without significantly increasing the complexity of the implementation.

awni:
> Uncertain of the cause, I decided to use the --force-download option to re-download the model. 

What's not clear to me is if the `--force-download` was necessary. Did you try just rerunning the download? I've encountered situations where the download gets interrupted and just redoing it works as expected. Sorry to belabor this, I'm just trying to understand in which cases one needs to use the `force-download` option.

Either way there is no hash check when we load the model.. though it should fail if not all the weights are present

malibayram:
I encountered the same issue again today, and the solution was to use the `--force-download` option. You can find a related discussion here: [Hugging Face Issue #2197](https://github.com/huggingface/huggingface_hub/issues/2197#issuecomment-2047170683).

As discussed in that thread, the only reliable solution to this problem is to delete the cache. The `--force-download` option effectively performs this task in a simpler and more user-friendly manner. Reopening this PR to highlight the importance and utility of this feature for ensuring model integrity.

malibayram:
@awni

awni:
I never encountered that issue. But since it's such a minor change let's merge this. I will wait until the tests clear.

awni (approved):
🙏

PR #806, opened 2024-05-30
Logprobs info to completion API

chimezie:
Motivated by attempt to add mlx support [to](https://github.com/EleutherAI/lm-evaluation-harness/pull/1902) lm-evaluation-harness, this extends the HTTP Model Server's completion API implementation to include logprobs information in the choices result JSON
- The log probabilities per token generated
- The [_top_logprobs_](https://platform.openai.com/docs/api-reference/completions/object#completions/object-choices) the tokens with the highest probability at each token position in what the model generated.

To be able to calculate top_logprobs, the  **generate_step** function was modified to be able to generate a triple instead of a 2-item tuple, where the last item is the _step logits_, i.e., the logits at each step of the generation for use in capturing and returning the top N tokens at each step of the generated token sequence.  A boolean parameter named **return_step_logits** was added to this function, which is False by default and causes the function to include the step logits if it is True. 

I could not find clarity on what this (legacy) API expects as the data structure the server should return to embed this information.  In addition, some literature and implementations return strings represented by the tokens instead of the tokens themselves, so this just follows the data structure that lm-evaluation-harness [expects](https://github.com/EleutherAI/lm-evaluation-harness/blob/main/lm_eval/models/openai_completions.py#L32).

Similarly, it interprets the *logprobs* OpenAI API parameter as specifying the number of top tokens to return at each token position of what is generated (by default, it doesn't do this).

The current version of lm-evaluation-harness currently from github invokes the completions API with max_tokens set to 0, which causes mlx to not generate anything in response to evaluation requests (see: [Issue 1903](https://github.com/EleutherAI/lm-evaluation-harness/issues/1903) ).


llms/mlx_lm/server.py
```diff
@@ -289,21 +313,42 @@ def handle_completion(
...
+
+            log_probs = mx.log(mx.softmax(step_logits, axis=-1))
+
+            if self.logprobs > 0:
+                sorted_indices = mx.argpartition(-log_probs, kth=self.logprobs, axis=-1)
+                top_n_indices = sorted_indices[:, : self.logprobs]
+                top_n_log_probs = mx.take_along_axis(log_probs, top_n_indices, axis=1)
+
+                top_token_info = {
+                    int(t.item()): l_prob.item()
```
chimezie, line 343: This cast to int is redundant but added because it seems the current server is serializing the tokens as integer strings despite the Python data dictionary having integers (the tokens as keys)

llms/mlx_lm/server.py
```diff
@@ -183,6 +185,11 @@ def validate_model_parameters(self):
         ):
             raise ValueError("repetition_penalty must be a non-negative float")
 
+        if self.logprobs != -1 and not (0 < self.logprobs <= 10):
```
chimezie, line 188: [Completion create API: logprobs](https://platform.openai.com/docs/api-reference/completions/create#completions-create-logprobs)

llms/mlx_lm/server.py
```diff
@@ -457,8 +510,12 @@ def handle_text_completions(self) -> mx.array:
 
         assert "prompt" in self.body, "Request did not contain a prompt"
         prompt_text = self.body["prompt"]
-
-        prompt = self.tokenizer.encode(prompt_text)
+        logging.debug(f"Prompt type: '{type(prompt_text)}'")
```
chimezie, line 511: Support for prompt submitted as a token sequence  instead of a string (would only work if lm-evaluation-harness is invoked with information about the tokenizer)

llms/mlx_lm/utils.py
```diff
@@ -206,15 +210,18 @@ def _step(y):
         if repetition_context_size:
             if len(repetition_context) > repetition_context_size:
                 repetition_context = repetition_context[-repetition_context_size:]
-        return y, prob
+        return y, prob, logits
```
awni, line 213: I would simplify this a bit:
1. Always just return `y, log_probs` and get rid of `prob` and get rid of the kwarg `return_step_logits`
2. Fix this to return the `log_probs` and not the `logits` which I think is wrong
3. Update the generate function to compute the exp of the `log_probs[t]` when it needs the actual token prob

chimezie: > I would simplify this a bit:
> 
> 1. Always just return `y, log_probs` and get rid of `prob` and get rid of the kwarg `return_step_logits`
> 2. Fix this to return the `log_probs` and not the `logits` which I think is wrong
> 3. Update the generate function to compute the exp of the `log_probs[t]` when it needs the actual token prob

@awni The OpenAI completions response expects the log probability for each returned token as well as the top N most likely tokens at each token sequence position and their corresponding log probabilities.  How can those be calculated without the logits at each generation step?

awni: > log probability for each returned token as well as the top N most likely tokens

You can get both of those from the `log_probs`. By `log_probs` I mean the vector of log probabilities for every token.

chimezie: @awni When you say **log_probs**, which variable are you referring to in that __step_ function? There is only the **logits** variable, which has the vector of raw, conditional probabilities for every token in the vocabulary and **prob**, which results from a call to the _sample_ function.  The _sample_ function returns the softmax probability for only the sampled token, losing the probabilities for the other tokens.

Without the probabilities for the other tokens in the vocabulary, how can you calculate the top N most likely ones?

llms/mlx_lm/server.py
awni: Can you explain this? It looks like it its a list we assume the tokens are already encoded and in the first item of the list? What's in the rest of the list? It seems a bit brittle to me.

awni:
@chimezie I made some modifications here. Could you check that the server works for your use case?

chimezie:
> @chimezie I made some modifications here. Could you check that the server works for your use case?
> 
> If its good we can land this shortly. Though it might be worth having a cleaner solution for the how we take already tokenized prompts.

@awni Yes, this works for the use case.  Thanks! I just made a few small changes to fix the logprobs response format, but nothing else.

llms/mlx_lm/server.py
chimezie: @awni For the completion API, the prompt can be provided as raw tokens [as well](https://github.com/openai/openai-python/blob/main/src/openai/types/completion_create_params.py#L24)  and the server was not handling that situation properly

awni: >  The prompt(s) to generate completions for, encoded as a string, array of strings, array of tokens, or array of token arrays.

Right I get that. It assumes if its a list its tokens and it only takes the first element?

chimezie: So, for the completion API, the inputs are:

> prompt(s) to generate completions for, encoded as a string, array of
>     strings, array of tokens, or array of token arrays.

So, ideally **handle_text_completions**,  **handle_completion**,and **generate_response** would have to change to handle these alternative prompt structures instead of just a single string. The use case I was adding this for, lm-evaluation-harness, can send multiple choice question/answer completions batched in a single HTTP request, in which case the prompt is a list of token sequences when it comes in and each should be completed and end up as a _choices_ ouput with multiple dictionaries, rather than just one with an index of 0 in the response.

awni: @chimezie I would prefer to add this change in a follow on PR as a separate thing and discuss it further. I'm not sure what the right thing to do is there, but I don't want to block the rest of this PR on figuring that out.

awni (approved):
Thanks for the contribution @chimezie !

fblissjr:
Apologize if I missed this, but I had a use case where I needed to return the token text in the server response.

Made a small change locally to do calculate and return text representations of tokens instead of just token IDs, along with text offsets based on the actual token text lengths.

I suppose the client could do this as well, but found it helpful when using a notebook mode UI to alter text.

Happy to add it in if anyone finds it helpful, or if it belongs on the client side, that works too.

fblissjr:
In case anyone is interested, I pushed the updated `server.py` with logprob texts (instead of just token IDs) here: https://github.com/fblissjr/mlx-examples/commit/6eefa51261345c21b5c83cf4d8c937e547b6dd50

@@ -351,6 +351,26 @@ def generate_response(
        }

    choice = response["choices"][0]
    if self.logprobs > 0:
        detokenizer = self.tokenizer.detokenizer
        detokenizer.reset()

        choice["logprobs"]["tokens"] = []
        choice["logprobs"]["text_offset"] = [0]
        accumulated_text = ""

        for token in tokens:
            detokenizer.add_token(token)
            token_text = detokenizer.last_segment
            choice["logprobs"]["tokens"].append(token_text)
            accumulated_text += token_text
            choice["logprobs"]["text_offset"].append(len(accumulated_text))

        choice["logprobs"]["token_logprobs"] = token_logprobs
        choice["logprobs"]["top_logprobs"] = [
            {self.tokenizer.decode([int(token)]): prob for token, prob in top_logprob.items()}
            for top_logprob in top_tokens
        ]

    # Add dynamic response
    if self.object_type.startswith("chat.completion"):

PR #810, opened 2024-06-02
Tweaks to run dspy-produced calls to the server, with gemma template.

namin:
Hi,

I am not sure if that is of general interest, but I thought I'd submit it in case it's useful.

I had to tweak the server a little to get it to interact with dspy following the following comment https://github.com/stanfordnlp/dspy/issues/385#issuecomment-1998939936.

The two tweaks are to (1) relax the validation and convert from int to float as needed, (2) to work around some chat template not taking the system role (such as Google Gemma's chat template).

can try it out with:
```sh
python -m server --model mlx-community/gemma-1.1-7b-it-4bit --port 1143
```
modulo patching the relative imports in server.py
```
-from .tokenizer_utils import TokenizerWrapper
-from .utils import generate_step, load
+from mlx_lm.tokenizer_utils import TokenizerWrapper
+from mlx_lm.utils import generate_step, load
```

and then, ont the dspy side:
```python
import dspy
lm = dspy.OpenAI(model_type="chat", api_base="http://localhost:11434/v1/", api_key="not_needed", max_tokens=250)
lm("hello")
```

Thanks!

### DIFF
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -51,22 +51,6 @@ def stopping_criteria(
     return StopCondition(stop_met=False, trim_length=0)
 
 
-def fold_in_system_prompt(messages: List[dict]):
-    n = len(messages)
-    if n>0 and messages[0]["role"] == "system":
-        message_system = messages[0]
-        if n>1:
-            message_user = messages[1]
-            assert message_user["role"] == "user"
-            message_user.update(
-                {"content": "System: " + message_system["content"] + "\n" + message_user["content"]}
-            )
-            messages = [message_user] + messages[2:]
-        else:
-            messages_system.update({"role": "user"})
-    return messages
-
-
 def convert_chat(messages: List[dict], role_mapping: Optional[dict] = None):
     default_role_mapping = {
         "system_prompt": "A chat between a curious user and an artificial intelligence assistant. The assistant follows the given rules no matter what.",
@@ -449,11 +433,8 @@ class APIHandler(BaseHTTPRequestHandler):
             hasattr(self.tokenizer, "apply_chat_template")
             and self.tokenizer.chat_template
         ):
-            messages = body["messages"]
-            if "raise_exception('System role not supported')" in self.tokenizer.chat_template:
-                messages = fold_in_system_prompt(messages)
             prompt = self.tokenizer.apply_chat_template(
-                messages,
+                body["messages"],
                 tokenize=True,
                 add_generation_prompt=True,
             )
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -544,6 +544,13 @@ def main():
         required=False,
     )
     parser.add_argument(
+        "--chat-template",
+        type=str,
+        default="",
+        help="Specify a chat template for the tokenizer",
+        required=False,
+    )
+    parser.add_argument(
         "--use-default-chat-template",
         action="store_true",
         help="Use the default chat template",
@@ -566,6 +573,8 @@ def main():
 
     # Building tokenizer_config
     tokenizer_config = {"trust_remote_code": True if args.trust_remote_code else None}
+    if args.chat_template:
+        tokenizer_config['chat_template'] = args.chat_template
 
     model, tokenizer = load(
         args.model, adapter_path=args.adapter_path, tokenizer_config=tokenizer_config
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -51,6 +51,19 @@ def stopping_criteria(
     return StopCondition(stop_met=False, trim_length=0)
 
 
+def openai_to_common_denominator(messages: List[dict]):
+    # fuse system and user, and strictly alternate with roles
+    for m in messages:
+        m["role"] = m["role"].replace("system", "user")
+    t_messages = []
+    for m in messages:
+        if t_messages == [] or t_messages[-1]["role"] != m["role"]:
+            t_messages.append(m)
+        else:
+            t_messages[-1]["content"] = "\n" + m["content"]
+    return t_messages
+
+
 def convert_chat(messages: List[dict], role_mapping: Optional[dict] = None):
     default_role_mapping = {
         "system_prompt": "A chat between a curious user and an artificial intelligence assistant. The assistant follows the given rules no matter what.",
@@ -433,8 +446,10 @@ class APIHandler(BaseHTTPRequestHandler):
             hasattr(self.tokenizer, "apply_chat_template")
             and self.tokenizer.chat_template
         ):
+            messages = body["messages"]
+            messages = openai_to_common_denominator(messages)
             prompt = self.tokenizer.apply_chat_template(
-                body["messages"],
+                messages,
                 tokenize=True,
                 add_generation_prompt=True,
             )
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -60,7 +60,7 @@ def openai_to_common_denominator(messages: List[dict]):
         if t_messages == [] or t_messages[-1]["role"] != m["role"]:
             t_messages.append(m)
         else:
-            t_messages[-1]["content"] = "\n" + m["content"]
+            t_messages[-1]["content"] += "\n" + m["content"]
     return t_messages
 
 
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -187,20 +187,14 @@ class APIHandler(BaseHTTPRequestHandler):
         if not isinstance(self.max_tokens, int) or self.max_tokens < 0:
             raise ValueError("max_tokens must be a non-negative integer")
 
-        if isinstance(self.temperature, int):
-            self.temperature = float(self.temperature)
-        if not isinstance(self.temperature, float) or self.temperature < 0:
+        if not isinstance(self.temperature, (float, int)) or self.temperature < 0:
             raise ValueError("temperature must be a non-negative float")
 
-        if isinstance(self.top_p, int):
-            self.top_p = float(self.top_p)
-        if not isinstance(self.top_p, float) or self.top_p < 0 or self.top_p > 1:
+        if not isinstance(self.top_p, (float, int)) or self.top_p < 0 or self.top_p > 1:
             raise ValueError("top_p must be a float between 0 and 1")
 
-        if isinstance(self.repetition_penalty, int):
-            self.top_p = float(self.repetition_penalty)
         if (
-            not isinstance(self.repetition_penalty, float)
+            not isinstance(self.repetition_penalty, (float, int))
             or self.repetition_penalty < 0
         ):
             raise ValueError("repetition_penalty must be a non-negative float")
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -140,7 +140,8 @@ class APIHandler(BaseHTTPRequestHandler):
         self.validate_model_parameters()
 
         # Get stop id sequences, if provided
-        stop_words = self.body.get("stop", [])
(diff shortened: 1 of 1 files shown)


awni:
Cool! Btw you shouldn't need to change the imports if you run it as a package like so:

```
mlx_lm.server --model mlx-community/gemma-1.1-7b-it-4bit --port 1143
```

llms/mlx_lm/server.py
```diff
@@ -171,12 +187,18 @@ def validate_model_parameters(self):
         if not isinstance(self.max_tokens, int) or self.max_tokens < 0:
             raise ValueError("max_tokens must be a non-negative integer")
 
+        if isinstance(self.temperature, int):
+            self.temperature = float(self.temperature)
         if not isinstance(self.temperature, float) or self.temperature < 0:
```
awni, line 192: I think if we just allow integers for these it will be simpler. E.g. 
```suggestion
        if not isinstance(self.temperature, (float, int)) or self.temperature < 0:
```

llms/mlx_lm/server.py
```diff
@@ -433,8 +455,11 @@ def handle_chat_completions(self) -> mx.array:
             hasattr(self.tokenizer, "apply_chat_template")
             and self.tokenizer.chat_template
         ):
+            messages = body["messages"]
+            if "raise_exception('System role not supported')" in self.tokenizer.chat_template:
```
awni, line 459: Could you explain what's happening here?

namin: The chat template for Gemma raises an exception if a system role is used. Here, I hard-code searching for the raised exception, and fold in the system message into the first user message. It's very hacky, and is actually not needed for running DSPy. I can remove all the logic related to this hack? Then the PR is just about (float, int) :) Thanks!

namin:
Done with your comments. Let me know if you prefer me to rebase. Thanks!

namin:
I tried a different application (OpenDevin), and to get it working with this server and gemma, I had to convert from OpenAI to Gemma messages as follows:
```
def openai2gemma(messages: List[dict]):
    print("before:", pretty_json(messages))
    for m in messages:
        m["role"] = m["role"].replace("system", "user")
    t_messages = []
    for m in messages:
        if t_messages == [] or t_messages[-1]["role"] != m["role"]:
            t_messages.append(m)
        else:
            t_messages[-1]["content"] = "\n" + m["content"]
    print("after:", pretty_json(t_messages))
    return t_messages
```

In short, replace system roles with user role, and make sure roles are strictly alternating.

Transforming from OpenAI messages to conform to a specific chat template is probably out of scope of this PR...

namin:
After further experimentation... the openai2gemma function seems useful for other models too, like llama3. In opendevin, mlx-community/Meta-Llama-3-70B-Instruct-4bit will silently work without the conversion function, but it produces questionable results.

llms/mlx_lm/server.py
```diff
@@ -544,6 +564,11 @@ def main():
     model, tokenizer = load(
         args.model, adapter_path=args.adapter_path, tokenizer_config=tokenizer_config
     )
+
+    if args.use_default_chat_template:
+        if tokenizer.chat_template is None:
+            tokenizer.chat_template = tokenizer.default_chat_template
```
awni, line 570: What's the purpose of that?

namin: DBRX (https://huggingface.co/mlx-community/dbrx-instruct-4bit) needs an explicit default chat template. This is how it's done in generate.py. See also https://huggingface.co/mlx-community/dbrx-instruct-4bit/discussions/7

llms/mlx_lm/server.py
```diff
@@ -433,8 +446,10 @@ def handle_chat_completions(self) -> mx.array:
             hasattr(self.tokenizer, "apply_chat_template")
             and self.tokenizer.chat_template
         ):
+            messages = body["messages"]
+            messages = openai_to_common_denominator(messages)
```
namin, line 450: this could be behind a flag?

llms/mlx_lm/server.py
```diff
@@ -51,6 +51,19 @@ def stopping_criteria(
...
+    # fuse system and user, and strictly alternate with roles
+    for m in messages:
+        m["role"] = m["role"].replace("system", "user")
+    t_messages = []
+    for m in messages:
+        if t_messages == [] or t_messages[-1]["role"] != m["role"]:
+            t_messages.append(m)
+        else:
+            t_messages[-1]["content"] = "\n" + m["content"]
+    return t_messages
```
awni, line 64: This chat template stuff makes my head spin a little. But shouldn't this be handled on the client side? It's a bit odd to me to hardcode the prompt modification on the server when the client can specify whatever they want in the prompt.

llms/mlx_lm/server.py
awni: Right, sounds good, thanks!

llms/mlx_lm/server.py
awni: Yes, maybe it is better to put it behind a flag for now. I left a comment elsewhere.. I'm still a little confused by it tbh. But at least if it's behind a flag it won't break anything by default.

llms/mlx_lm/server.py
namin: I agree it's not great. But my use case is precisely to use the OpenAI server to serve MLX LLMs when I have no control over the client, e.g. using an app or a library that supports the OpenAI API but not MLX out of the box. I am OK if you think this is out of scope. Thanks and sorry for the head spin :)

awni: I just have a question about that. Is the reason you have to do the mapping because other LLMs have different chat templates than what OpenAI uses? So the client side calls MLX server as if it's OpenAI but on the backend the model expects a different template?

But in that case this is pretty specific to a given LM right? Which LMs does it work for?

I'm just thinking about how to handle this in a more flexible way.

namin: Yes, that's exactly the reason... It seems pretty specific to the LM, but trying many of them, it seems the common denominator that they can all accept is alternating user/assistant messages.

namin:
@ awni a much better solution to deal with chat templates occurred to me. Just provide a --chat-template flag that is passed to the tokenizer. The provider is then responsible to adhere to the OpenAI chat model. For gemma, this template seems to work:
```sh
--chat-template "{{ bos_token }}{% set extra_system = '' %}{% for message in messages %}{% if (message['role'] == 'assistant') %}{% set role = 'model' %}{% else %}{% set role = message['role'] %}{% endif %}{% if role == 'system' %}{% set extra_system = extra_system + message['content'] %}{% else %}{% if role == 'user' and extra_system %}{% set message_system = 'System: ' + extra_system %}{% else %}{% set message_system = '' %}{% endif %}{{ '<start_of_turn>' + role + '\n' + message_system + message['content'] | trim + '<end_of_turn>\n' }}{% endif %}{% endfor %}{% if add_generation_prompt %}{{'<start_of_turn>model\n'}}{% endif %}"
```
It basically does the convoluted logic in the jinja2 code.

What do you think?

Thanks!

awni:
I think that's great! Though I'm curious, where did you get the chat template for Gemma from? Is there a standard process for coming up with those?

namin:
Cool :)

I started with the original template, found like this:
```sh
% python
>>> from mlx_lm import load
>>> 
>>> model, tokenizer = load("mlx-community/gemma-1.1-7b-it-4bit")
>>> tokenizer.chat_template
"{{ bos_token }}{% if messages[0]['role'] == 'system' %}{{ raise_exception('System role not supported') }}{% endif %}{% for message in messages %}{% if (message['role'] == 'user') != (loop.index0 % 2 == 0) %}{{ raise_exception('Conversation roles must alternate user/assistant/user/assistant/...') }}{% endif %}{% if (message['role'] == 'assistant') %}{% set role = 'model' %}{% else %}{% set role = message['role'] %}{% endif %}{{ '<start_of_turn>' + role + '\n' + message['content'] | trim + '<end_of_turn>\n' }}{% endfor %}{% if add_generation_prompt %}{{'<start_of_turn>model\n'}}{% endif %}"
```

and made it lenient instead of throwing exceptions when a system role is found or when roles do not alternate.

So in particular, like I was doing in the workaround in Python here earlier, I fold in the system messages to the next user message, prepending `System: ` and all the system contents accumulated in the meantime. That seems to work for most use cases. I don't bother checking alternation of roles at the moment.

llms/mlx_lm/server.py
```diff
@@ -540,10 +553,17 @@ def main():
...
     # Building tokenizer_config
     tokenizer_config = {"trust_remote_code": True if args.trust_remote_code else None}
+    if args.chat_template:
+        tokenizer_config['chat_template'] = args.chat_template
 
     model, tokenizer = load(
         args.model, adapter_path=args.adapter_path, tokenizer_config=tokenizer_config
     )
+
+    if args.use_default_chat_template:
```
awni, line 563: Should it be an error to specify a chat-template and `use-default-chat-template`?

llms/mlx_lm/server.py
```diff
@@ -140,7 +140,8 @@ def do_POST(self):
         self.validate_model_parameters()
 
         # Get stop id sequences, if provided
-        stop_words = self.body.get("stop", [])
+        stop_words = self.body.get("stop")
+        stop_words = stop_words or []
```
awni, line 144: Is that so the set value can be `None`?

namin: Exactly.

llms/mlx_lm/server.py
awni: Sounds good!

namin:
I did some debugging on my gemma template, and I ran into the issue/misconception in jinja2 that variables outside a loop cannot be updated inside! I used a namespace as a workaround, as suggested here: https://stackoverflow.com/questions/46939756/setting-variable-in-jinja-for-loop-doesnt-persist-between-iterations.

So this is my updated chat-template parameter:
```
"{{ bos_token }}{% set ns = namespace(extra_system='') %}{% for message in messages %}{% set role = message['role'] %}{% if (message['role'] == 'assistant') %}{% set role = 'model' %}{% endif %}{% if (role == 'system') %}{% set ns.extra_system = ns.extra_system + message['content'] %}{% else %}{% set message_system = '' %}{% if (role == 'user') %}{% if (ns.extra_system == '') %}{% else %}{% set message_system = 'System: ' + ns.extra_system + '\\n' %}{% set ns.extra_system = '' %}{% endif %}{% endif %}{{ '<start_of_turn>' + role + '\\n' + message_system + message['content'] | trim + '<end_of_turn>\\n' }}{% endif %}{% endfor %}{% if add_generation_prompt %}{{'<start_of_turn>model\\n'}}{% endif %}"
```

For readability:
```
{{ bos_token }}
{% set ns = namespace(extra_system='') %}
{% for message in messages %}
  {% set role = message['role'] %}
  {% if (message['role'] == 'assistant') %}
    {% set role = 'model' %}
  {% endif %}
  {% if (role == 'system') %}
    {% set ns.extra_system = ns.extra_system + message['content'] %}
  {% else %}
    {% set message_system = '' %}
    {% if (role == 'user') %}
      {% if (ns.extra_system == '') %}
      {% else %}
        {% set message_system = 'System: ' + ns.extra_system + '\n' %}
        {% set ns.extra_system = '' %}
      {% endif %}
    {% endif %}
    {{ '<start_of_turn>' + role + '\n' + message_system + message['content'] | trim + '<end_of_turn>\n' }}
  {% endif %}
{% endfor %}
{% if add_generation_prompt %}{{'<start_of_turn>model\n'}}{% endif %}
```

awni (approved):
Thank you!!

Also I'm wondering if we should document those prompts somewhere 🤔 or at least the process you used to determine how to build it.

PR #813, opened 2024-06-04
Su-RoPE(Rotary Position Embedding) for Phi-3

JosefAlbers:
This Pull Request introduces an implementation of the su-RoPE used by some of the Phi-3 language models. 

Key Changes:

- Supports the non-traditional RoPE variant used by Phi-3 (the traditional argument is included but unused).
- Handles query and key rotations separately, matching the current mlx-lm implementation.

Future Optimization:

- The code structure allows for easy modification to perform both query and key rotations in a single call for potential efficiency gains.

### DIFF
--- a/llms/mlx_lm/models/su_rope.py
+++ b/llms/mlx_lm/models/su_rope.py
@@ -1,6 +1,7 @@
 import math
 
 import mlx.core as mx
+from typing import List, Union
 
 
 class SuScaledRotaryEmbedding:
@@ -12,8 +13,8 @@ class SuScaledRotaryEmbedding:
         scale: float = 1.0,
         max_position_embeddings: int = 131072,
         original_max_position_embeddings: int = 4096,
-        short_factor: list[float] | float = 1.0,
-        long_factor: list[float] | float = 1.0,
+        short_factor: Union[List[float], float] = 1.0,
+        long_factor: Union[List[float], float] = 1.0,
     ):
         """
         Phi3Su Scaled Rotary Embedding layer for Phi-3 models.
--- a/llms/mlx_lm/models/su_rope.py
+++ b/llms/mlx_lm/models/su_rope.py
@@ -1,8 +1,9 @@
 import math
+
 import mlx.core as mx
 
 
-class Phi3SuScaledRotaryEmbedding:
+class SuScaledRotaryEmbedding:
     def __init__(
         self,
         dims: int,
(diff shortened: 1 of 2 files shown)


llms/mlx_lm/models/su_rope.py
```diff
@@ -0,0 +1,70 @@
+import math
+import mlx.core as mx
+
+
+class Phi3SuScaledRotaryEmbedding:
```
awni, line 5: I'm wondering if it's possible to implement this in terms of `mx.fast.rope` as it will be much much faster. It looks to me like you can pass in the appropriate base and scaling factor and achieve that?

JosefAlbers:
Thanks for the suggestion, but I'm at this moment not entirely sure if it's directly feasible since the mx.fast.rope applies a uniform scale across dimensions. I'll look into whether it is indeed possible.

awni:
>  since the mx.fast.rope applies a uniform scale across dimensions.

Indeed it may not be possible with the current API 🤔 . Have you seen this Su ROPE  used with many other models or is Phi3 a bit of an exception?

JosefAlbers:
> Have you seen this Su ROPE used with many other models or is Phi3 a bit of an exception?

Phi-3 seems to be the only one I've come across using Su ROPE, but my knowledge of the broader model landscape is yet very limited."

llms/mlx_lm/models/su_rope.py
```diff
@@ -0,0 +1,70 @@
...
+        inv_freq = (
+            self.inv_freq_long
+            if position_ids.max() + 1 > self.original_max_position_embeddings
+            else self.inv_freq_short
+        )
+        inv_freq_expanded = mx.repeat(
+            inv_freq[None, :, None], position_ids.shape[0], axis=0
+        )
+        position_ids_expanded = position_ids[:, None, :]
+        freqs = (inv_freq_expanded @ position_ids_expanded).transpose(0, 2, 1)
```
awni, line 57: This could be simpler and probably faster with just a multiply:
```suggestion
        freqs = inv_freqs[:, None] * position_ids[:, None, :]
```

JosefAlbers:
@awni Thanks for the suggestion, but I'm not sure I follow how the shapes would align for element-wise multiplication in this case. When I tried implementing it, I ran into a ValueError due to a shape mismatch (e.g., `ValueError: Shapes (1,32,3,96) and (1,1,48,6) cannot be broadcast.`). Would you mind elaborating a bit more on using multiply?

awni:
@JosefAlbers sorry my suggestion was not fully complete. Take a look at the diff I pushed to see what I meant: https://github.com/ml-explore/mlx-examples/pull/813/commits/ff569832d3a2c938ce33646ffe58bea96774a79f

JosefAlbers:
@awni Oh, I get it now. Wow.

JosefAlbers:
@awni sorry for the silly error I made in the su_rope.py (I ran the test on python 3.12 when the circleci runs the tests on python <3.9). I fixed the bug now.

JosefAlbers:
@awni sorry for the trouble, I didn't know about isort.

awni:
@JosefAlbers I'm curious about this. It looks like the su Rope is applied in the condition: `if self.rope_scaling["type"] not in ["su", "linear"]:` but the [rope type of the 128k models is `longrope`](https://huggingface.co/microsoft/Phi-3-mini-128k-instruct/blob/main/config.json#L128) so right now the su Rope is never used. Is that just an oversight or am I missing something?

JosefAlbers:
@awni I see Microsoft had made some changes to their Phi-3 models with their July update. Among them are:
- Original modeling_phi3.py before the July update: https://huggingface.co/microsoft/Phi-3-mini-128k-instruct/blob/5be6479b4bc06a081e8f4c6ece294241ccd32dec/modeling_phi3.py
- New modeling_phi3.py after the July update:
https://huggingface.co/microsoft/Phi-3-mini-128k-instruct/blob/main/modeling_phi3.py

But both the Phi3SuScaledRotaryEmbedding in their original model and the Phi3LongRoPEScaledRotaryEmbedding in the new updated model still have the same underlying implementation. Some modification to the phi3.py file I think may be needed at this point:

```python
...
            if self.rope_scaling["type"] not in ["longrope", "su", "linear"]:
                print(
                    "[WARNING] rope_scaling 'type' currently only supports 'linear' and 'su'; setting rope scaling to false."
                )
...
        if args.rope_scaling and args.rope_scaling["type"] in ["longrope", "su"]:
...
```

I will prepare the PR implementing these changes to the file if you would like.

awni:
I see.. thanks for the explanation!

PR #818, opened 2024-06-05
Add eos token to lora fine-tunes

### LINKED ISSUE #811: A simple enhancement, in dataset creation time
Hi Team, 
recently, i was using phi3 model to train model on binary classification datatset.
Despite training it for good amount of time, when i try to use it, it give me output but fails to stop after giving class name.
Problem : in current code there is no concept of adding eos_token . Hence after giving otuput it start giving prompt back.

so i made slight change in dataset creation function and after that, i trained the model again and it worked like charm.
i was not getting one word output as i want. there is no more un neccessary extra answer.
`class ChatDataset(Dataset):

    def __init__(self, path: Path, tokenizer: PreTrainedTokenizer):
        super().__init__(path)
        self._tokenizer = tokenizer

    def __getitem__(self, idx: int):
        messages = self._data[idx]["messages"]
        text = self._tokenizer.apply_chat_template(
            messages, tokenize=False, add_generation_prompt=True
        )
        text = text+self._tokenizer.eos_token   ## this is new addition of code.
        # print(text)
        return text`
        
        
 requesting github repo owner, to kindly look into this
  and update the code..thanks

awni:
Add the eos token id to sequences during fine-tuning. Closes #811

### DIFF
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -92,6 +92,9 @@ def iterate_batches(dataset, tokenizer, batch_size, max_seq_length, train=False)
         for i in indices:
             # Encode batch
             batch = [tokenizer.encode(dataset[j]) for j in batch_idx[i]]
+            for b in batch:
+                b.append(tokenizer.eos_token_id)
+
             lengths = [len(x) for x in batch]
 
             if max(lengths) > max_seq_length:
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -93,7 +93,10 @@ def iterate_batches(dataset, tokenizer, batch_size, max_seq_length, train=False)
             # Encode batch
             batch = [tokenizer.encode(dataset[j]) for j in batch_idx[i]]
             for b in batch:
-                b.append(tokenizer.eos_token_id)
+                if b[-1] == tokenizer.eos_token_id:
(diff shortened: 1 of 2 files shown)


llms/mlx_lm/tuner/trainer.py
```diff
@@ -92,6 +92,9 @@ def iterate_batches(dataset, tokenizer, batch_size, max_seq_length, train=False)
         for i in indices:
             # Encode batch
             batch = [tokenizer.encode(dataset[j]) for j in batch_idx[i]]
+            for b in batch:
+                b.append(tokenizer.eos_token_id)
+
```
barronalex, line 97: I was trying to figure out if there's some something in the tokenizer's interface that let's you control this but it seems it could be different across [all supported ones](https://huggingface.co/docs/transformers/en/model_doc/auto#transformers.AutoTokenizer)?

I guess if we wanted to be really defensive we could check if the last token of the sequence is already `eos_token_id`? Maybe overkill though.

barronalex (approved):
Looks good! Just one small question

llms/mlx_lm/tuner/trainer.py
awni: I spent a long time trying to figure that out myself and did not succeed so I gave up and wrote this hack 😅 

It's a good idea to check the end. I was wondering myself.. if it ever happens. I don't think it does because then presumably our generation code would be broken (we'd be adding eos to the prompt) but who knows.

PR #821, opened 2024-06-06
Enable distributed LoRA training

angeloskath:
The updates to `LORA.md` are missing but TL;DR we can now do

```
$ echo "m2-ultra-0 slots=1" >>hostfile
$ echo "m2-ultra-1 slots=1" >>hostfile
$ mpirun --hostfile hostfile -- python -m mlx_lm.lora --train --model mlx-community/Mistral-7B-v0.2-4bit --data /path/to/data --batch-size 16
```

to train across two nodes (or more really nothing needs to change).

### DIFF
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -254,7 +254,8 @@ def train(
                 print(
                     f"Iter {it}: "
                     f"Val loss {val_loss:.3f}, "
-                    f"Val took {val_time:.3f}s"
+                    f"Val took {val_time:.3f}s",
+                    flush=True,
                 )
 
             if training_callback is not None:
@@ -292,7 +293,8 @@ def train(
                     f"It/sec {it_sec:.3f}, "
                     f"Tokens/sec {tokens_sec:.3f}, "
                     f"Trained Tokens {trained_tokens}, "
-                    f"Peak mem {peak_mem:.3f} GB"
+                    f"Peak mem {peak_mem:.3f} GB",
+                    flush=True,
                 )
 
             if training_callback is not None:
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -10,6 +10,7 @@ from typing import Union
 import mlx.core as mx
 import mlx.nn as nn
 import numpy as np
+from mlx.nn.utils import average_gradients
 from mlx.utils import tree_flatten, tree_map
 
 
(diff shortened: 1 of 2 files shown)


mzbac:
Is that possible to do distributed inference as well?

awni:
> Is that possible to do distributed inference as well?

Possible yes, but getting a nice speedup is more challenging. That's something we're looking at, but don't have an ETA on right now.

angeloskath:
@awni feel free to review and then we can merge. I split the launcher to a different branch.

llms/tests/test_finetune.py
```diff
@@ -374,16 +383,17 @@ def test_evaluate_calls(self):
...
-            model=mock_model,
-            dataset=mock_dataset,
-            tokenizer=mock_tokenizer,
-            batch_size=2,
-            num_batches=2,
-            max_seq_length=2048,
-            loss=mock_default_loss,
-            iterate_batches=mock_iterate_batches,
-        )
+        with swapped_with_identity(mx.distributed, "all_sum"):
```
awni, line 386: Just curious, why do we need this for the test to work?

llms/mlx_lm/tuner/trainer.py
```diff
@@ -229,9 +250,13 @@ def step(batch):
...
-            )
+            if rank == 0:
+                print(
+                    f"Iter {it}: "
+                    f"Val loss {val_loss:.3f}, "
+                    f"Val took {val_time:.3f}s",
+                    flush=True,
+                )
 
             if training_callback is not None:
```
awni, line 261: Probably tha tshould go under the `rank==0` condition as well

llms/mlx_lm/tuner/trainer.py
```diff
@@ -244,30 +269,33 @@ def step(batch):
...
+                    f"Iter {it}: Train loss {train_loss:.3f}, "
+                    f"Learning Rate {learning_rate:.3e}, "
+                    f"It/sec {it_sec:.3f}, "
+                    f"Tokens/sec {tokens_sec:.3f}, "
+                    f"Trained Tokens {trained_tokens}, "
+                    f"Peak mem {peak_mem:.3f} GB",
+                    flush=True,
+                )
 
             if training_callback is not None:
```
awni, line 300: Same there.

llms/tests/test_finetune.py
awni: Nvm I see the magicmock thing is messing up the `all_sum`.

awni (approved):
Looks great!! Let's 🚢

llms/mlx_lm/tuner/trainer.py
angeloskath: Yeah ok, it makes sense. I was thinking in general callbacks should always run and if the callback is about reporting then it can choose to only run on rank=0. But our callbacks here are only about reporting so it makes sense to just run them only on node 0.

awni: It's a good point actually. It's more flexible that way. I'm on board leaving it to the user to specify the rank.

ivanfioravanti:
This works perfectly! Great job 👏

PR #822, opened 2024-06-06
Fix Qwen2 moe

Blaizzy:
This PR fixes Qwen2 MoE conversion and inference.
Twitter: @Prince_Canuma

### DIFF
--- a/llms/mlx_lm/models/qwen2_moe.py
+++ b/llms/mlx_lm/models/qwen2_moe.py
@@ -224,7 +224,7 @@ class Model(nn.Module):
         for l in range(self.args.num_hidden_layers):
             prefix = f"model.layers.{l}"
             for n in ["up_proj", "down_proj", "gate_proj"]:
-                for k in ["weight", "scales", "biases"]:
+                for k in ["weight", "biases"]:
                     if f"{prefix}.mlp.experts.0.{n}.{k}" in weights:
                         to_join = [
                             weights.pop(f"{prefix}.mlp.experts.{e}.{n}.{k}")


awni:
This doesn’t look like the correct fix to me. We need to merge scales if they are there. Could you point to the model that didn’t work so we can debug?

Blaizzy:
I could be wrong but from what I can see, both Qwen moe models don’t use scales. 

- https://huggingface.co/Qwen/Qwen1.5-MoE-A2.7B/tree/main
- https://huggingface.co/Qwen/Qwen2-57B-A14B-Instruct/tree/main

Precisely, the Qwen-2-57B-A14B model failed to convert.

awni:
I do not have a problem converting `Qwen/Qwen2-57B-A14B-Instruct` on main. Can you share the error message you are getting?

Blaizzy:
But they will be present in quantized models. I see it now. 

I will add a condition to skip scales for f16 or f32 models.

Blaizzy:
> I do not have a problem converting `Qwen/Qwen2-57B-A14B-Instruct` on main. Can you share the error message you are getting?

Alright, give me a 15min to redownload the model.

> You don't need to add that, it will not try to use scales if they are not present already. See Line 228.

I noticed, and saw the git blame to trace the changes.
From what I can tell, the line above that (line 227)  is the one what is at fault :)

Blaizzy:
That's weird, today is working fine.

Yesterday both pypi and github installation gave an error

An error exactly in line 228 where it said something about the first expert didn't have scales but I didn't save it anywhere.

PR #824, opened 2024-06-07
GPU featurization for Whisper example

awni:
As title. Timings for 1 hour of audio:

Device | Time (milliseconds) |  Speedup over real time
---- | ---- | -----
CPU | 367 | 9,800
GPU | 20 | 180,000


barronalex (approved):
🚀

PR #825, opened 2024-06-07
Add functions for input-masked loss calculation and padded batching

chimezie:
Adds support for completion-only finetuning via functions for iterating over batching that also calculates input masks along with padding and for calculating loss using the masks

### DIFF
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -60,6 +60,65 @@ class TrainingArgs:
     )
 
 
+def iterate_input_masked_batches(
+    input_text, output_text, tokenizer, max_seq_length=2048
+):
+    batch_size = len(input_text)
+
+    input_batch = [tokenizer.encode(record) for record in input_text]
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/tuner/trainer.py
```diff
@@ -63,6 +63,65 @@ class TrainingArgs:
...
+            output_lengths[j], max_length_in_batch - input_length
+        )
+        adjusted_lengths.append(full_ids_end_idx)
+        batch_arr[j, :full_ids_end_idx] = full_labels[j][:full_ids_end_idx]
+
+    batch = mx.array(batch_arr)
+    input_lengths = mx.array(input_lengths)
+    lengths = mx.array(adjusted_lengths)
+
+    return batch[:, :-1], input_lengths, lengths
```
awni, line 104: This only returns one example? Is that intentional? I assumed this is a drop-in replacement for iterate_batches but it's not clear that is the case based on how it's written..

chimezie:
Superseded by #1103

PR #827, opened 2024-06-10
Correct type annotation of llama.ModelArgs.num_key_value_heads

wangkuiyi:
mypy complains about the assigning of value `None` to the field of type `int`.


awni (approved):
🙏

Issue #829, opened 2024-06-10
Proposal: Add mypy to .pre-commit-config.yml


wangkuiyi:
So far the .pre-commit-config.yml calls only black and isort. I accidentally used https://github.pie.apple.com/aiml-oss/ml-recurrent-drafter/blob/main/.pre-commit-config.yaml and noticed the issues to be fixed in #827 and #828. Is may be a good idea to include mypy and others tools from the above pre-commit configuration.

awni:
I think its a good idea, especially if we are going to use type annotations (which we do).

wangkuiyi:
[ ] gguf_llm/generate.py
[ ] gguf_llm/models.py
[ ] llama/convert.py
[ ] mistral/convert.py
[ ] mixtral/convert.py
[ ] mixtral/mixtral.py
[ ] mlx_lm/lora.py
[ ] mlx_lm/merge.py
[ ] mlx_lm/models/cohere.py
[ ] mlx_lm/models/dbrx.py
[ ] mlx_lm/models/gemma.py
[ ] mlx_lm/models/gpt2.py
[ ] mlx_lm/models/gpt_bigcode.py
[ ] mlx_lm/models/internlm2.py
[ ] mlx_lm/models/llama.py
[ ] mlx_lm/models/minicpm.py
[ ] mlx_lm/models/mixtral.py
[ ] mlx_lm/models/olmo.py
[ ] mlx_lm/models/openelm.py
[ ] mlx_lm/models/phi3.py
[ ] mlx_lm/models/phi3small.py
[ ] mlx_lm/models/plamo.py
[ ] mlx_lm/models/qwen2.py
[ ] mlx_lm/models/qwen2_moe.py
[ ] mlx_lm/models/starcoder2.py
[ ] mlx_lm/server.py
[ ] mlx_lm/tuner/trainer.py
[ ] mlx_lm/tuner/utils.py
[ ] mlx_lm/utils.py
[ ] speculative_decoding/decoder.py
[ ] speculative_decoding/model.py
[ ] tests/test_server.py

ParamThakkar123:
@awni May I take on this issue ??

ParamThakkar123:
How many of these are done ?

awni:
Sure!  Just FYI `mlx_lm` has moved to a new repo https://github.com/ml-explore/mlx-lm

I don't know how many are done.. not too many It hink.

ParamThakkar123:
[...]

PR #830, opened 2024-06-10
Add mypy to .pre-commit-config.yml

wangkuiyi:
Fix https://github.com/ml-explore/mlx-examples/issues/829

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,3 +9,8 @@ repos:
     -   id: isort
         args:
             - --profile=black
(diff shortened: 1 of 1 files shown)


wangkuiyi:
Ah. Circle CI calls pre-commit. So we have to fix all mypy errors before enabling mypy in pre-commit.

awni:
Oh yes.. that might take some time :\. I wonder if we could point it to a list of files to start and add more over time 🤔

wangkuiyi:
Sure. I am making a list and will paste to https://github.com/ml-explore/mlx-examples/issues/829

PR #833, opened 2024-06-11
make models/phi3.py and models/phi3small.py compatible with mypy

wangkuiyi:
## Test Plan:

Run the following command:

```shell
❯ mypy \
  --explicit-package-bases \
  --ignore-missing-imports \
. | grep models/phi
```

Before this change, it should complain the following:

```
mlx_lm/models/phi3small.py:22: error: Incompatible types in assignment (expression has type "None", variable has type "int")  [assignment]
mlx_lm/models/phi3small.py:29: error: Incompatible types in assignment (expression has type "Tuple[int]", variable has type "int")  [assignment]
mlx_lm/models/phi3small.py:74: error: Name "math" is not defined  [name-defined]
mlx_lm/models/phi3small.py:176: error: "Tuple[Any, Any]" has no attribute "offset"  [attr-defined]
mlx_lm/models/phi3small.py:177: error: "Tuple[Any, Any]" has no attribute "offset"  [attr-defined]
mlx_lm/models/phi3small.py:178: error: "Tuple[Any, Any]" has no attribute "update_and_fetch"  [attr-defined]
mlx_lm/models/phi3.py:20: error: Incompatible types in assignment (expression has type "None", variable has type "int")  [assignment]
mlx_lm/models/phi3.py:68: error: Argument "short_factor" to "SuScaledRotaryEmbedding" has incompatible type "Union[float, str]"; expected "Union[List[float], float]"  [arg-type]
mlx_lm/models/phi3.py:69: error: Argument "long_factor" to "SuScaledRotaryEmbedding" has incompatible type "Union[float, str]"; expected "Union[List[float], float]"  [arg-type]
mlx_lm/models/phi3.py:73: error: Unsupported operand types for / ("int" and "str")  [operator]
mlx_lm/models/phi3.py:73: note: Right operand is of type "Union[float, str]"
mlx_lm/models/phi3.py:101: error: "Tuple[Any, Any]" has no attribute "offset"  [attr-defined]
mlx_lm/models/phi3.py:102: error: "Tuple[Any, Any]" has no attribute "offset"  [attr-defined]
mlx_lm/models/phi3.py:103: error: "Tuple[Any, Any]" has no attribute "update_and_fetch"  [attr-defined]
```

After this change, the above command should complain no more.

### DIFF
--- a/llms/mlx_lm/models/phi3.py
+++ b/llms/mlx_lm/models/phi3.py
@@ -1,10 +1,10 @@
 from dataclasses import dataclass
-from typing import Dict, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import mlx.core as mx
 import mlx.nn as nn
 
-from .base import BaseModelArgs
+from .base import BaseModelArgs, KVCache
 from .su_rope import SuScaledRotaryEmbedding
 
 
@@ -17,10 +17,10 @@ class ModelArgs(BaseModelArgs):
     num_attention_heads: int
     rms_norm_eps: float
     vocab_size: int
-    num_key_value_heads: int = None
+    num_key_value_heads: Optional[int] = None
     rope_theta: float = 10000
     rope_traditional: bool = False
-    rope_scaling: Optional[Dict[str, Union[float, str]]] = None
+    rope_scaling: Optional[Dict[str, Union[float, List[float]]]] = None
     max_position_embeddings: int = 131072
     original_max_position_embeddings: int = 4096
 
@@ -46,6 +46,7 @@ class Attention(nn.Module):
 
         dim = args.hidden_size
         self.n_heads = n_heads = args.num_attention_heads
+        assert args.num_key_value_heads is not None
         self.n_kv_heads = n_kv_heads = args.num_key_value_heads
         self.num_hidden_layers = args.num_hidden_layers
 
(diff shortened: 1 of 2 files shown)


llms/mlx_lm/models/phi3.py
```diff
@@ -17,10 +17,10 @@ class ModelArgs(BaseModelArgs):
     num_attention_heads: int
     rms_norm_eps: float
     vocab_size: int
-    num_key_value_heads: int = None
+    num_key_value_heads: Optional[int] = None
     rope_theta: float = 10000
     rope_traditional: bool = False
-    rope_scaling: Optional[Dict[str, Union[float, str]]] = None
+    rope_scaling: Optional[Dict[str, Union[float, List[float]]]] = None
```
wangkuiyi, line 23: Using Python dictionary implicitly presumed schemeless or typeless coding style, which is against the style of the rest fields in this dataclass. We may want to change this dictionary into a nested dataclass in the future.

awni: Indeed that would probably be better.

awni (approved):
🙏  thanks for the fixes. Just letting the tests clear then I will merge it.

llms/mlx_lm/models/phi3.py
wangkuiyi: I will try if that works today, then file another PR if it does.

PR #838, opened 2024-06-14
transformer_lm: add --dataset enwik8

proger:
enwik8 is great for small runs if you're into small vocabularies!


awni (approved):
Looks like a nice addition, thanks!

PR #840, opened 2024-06-16
 added docstring explanation for  bertenbeddings class in model.py in…

saul1310:
This change adds a docstring explanation for the class Bertembeddings inside of model.py inside of bert. This is my first open source commit and  I hope this helps!
Signed off by saul Ifshin <saulifshin.cs@gmail.com>

### DIFF
--- a/bert/model.py
+++ b/bert/model.py
@@ -59,6 +59,8 @@ class TransformerEncoder(nn.Module):
 
 
 class BertEmbeddings(nn.Module):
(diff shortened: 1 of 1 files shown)


awni:
Thanks for the contribution. I'm going to close this though as I don't think the docstring is adding much here to help someone understand the purpose of that class (beyond the name itself), and it's somewhat inconsistent to just have one in that class and nowhere else.

If you would like to contribute some documenation or otherwise feel free to do so or open an issue / discussion if you are unsure of what to add.

PR #851, opened 2024-06-26
Server loads the model on demand from the request

angeloskath:
As title. Basically moves the model loading logic into the `ModelProvider`. The POST handler passes the `requested_model` to the model provider to load that model instead of always using the default one.

It should be completely backwards compatible even wrt to when the default model will be loaded.


awni (approved):
Looks great and like a useful feature. Thanks for adding it!

PR #853, opened 2024-06-26
Example of response generation with optional arguments

a-wozniakowski:
The example is meant to extend the [README](https://github.com/a-wozniakowski/mlx-examples/tree/main/llms), whereby it:
- uses a [checkpoint](https://huggingface.co/mlx-community/Mixtral-8x22B-Instruct-v0.1-4bit) from the mlx-community
- loads the model and tokenizer with `load`
- transforms a prompt into a chat template with the `apply_chat_template` method of the tokenizer
- passes optional arguments into `generate`, which show how to control the generation process

### DIFF
--- a/llms/README.md
+++ b/llms/README.md
@@ -38,7 +38,7 @@ To see a description of all the arguments you can do:
 >>> help(generate)
 ```
 
-For a detailed example of response generation with optional arguments, see `mlx_lm/examples/generate_response.py`. 
+Check out the [generation example](https://github.com/ml-explore/mlx-examples/tree/main/llms/mlx_lm/examples/generate_response.py) to see how to use the API in more detail.
 
 The `mlx-lm` package also comes with functionality to quantize and optionally
 upload models to the Hugging Face Hub.
--- a/llms/README.md
+++ b/llms/README.md
@@ -38,6 +38,8 @@ To see a description of all the arguments you can do:
 >>> help(generate)
 ```
 
+For a detailed example of response generation with optional arguments, see `mlx_lm/examples/generate_response.py`. 
+
 The `mlx-lm` package also comes with functionality to quantize and optionally
 upload models to the Hugging Face Hub.
 
--- a/llms/mlx_lm/examples/generate_response.py
+++ b/llms/mlx_lm/examples/generate_response.py
@@ -1,5 +1,4 @@
-from mlx_lm import load, generate
-
+from mlx_lm import generate, load
 
 # Specify the checkpoint
 checkpoint = "mistralai/Mistral-7B-Instruct-v0.3"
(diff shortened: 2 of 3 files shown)


llms/mlx_lm/examples/generate_response.py
```diff
@@ -0,0 +1,45 @@
+from mlx_lm import load, generate
+
+
+# Specify the checkpoint
+checkpoint = "mlx-community/Mixtral-8x22B-Instruct-v0.1-4bit"
```
awni, line 5: Let's use Mistral 7B v0.3 instruct for this example, as it is a more accessible model.

llms/mlx_lm/examples/generate_response.py
```diff
@@ -0,0 +1,45 @@
...
+)
+
+# Specify the maximum number of tokens
+max_tokens = 1_000
+
+# Specify if tokens and timing information will be printed
+verbose = True
+
+# Some optional arguments for causal language model generation
+generation_details = {
```
awni, line 28: nit `generation_details` -> `generation_args`

llms/mlx_lm/examples/generate_response.py
```diff
@@ -0,0 +1,45 @@
...
+response = generate(
+    model=model,
+    tokenizer=tokenizer,
+    prompt=prompt,
+    max_tokens=max_tokens,
+    verbose=verbose,
+    **generation_details
+)
+
+print(response)
```
awni, line 45: If it's verbose I think we can remove this extra print. Will keep the output somewhat readable.

awni (commented):
Thanks for adding this, I left a few minor comments, then we can merge it. Also maybe worth linking to this example in the top-level MLX LM [README](https://github.com/ml-explore/mlx-examples/blob/main/llms/README.md#python-api)

llms/mlx_lm/examples/generate_response.py
a-wozniakowski: This choice resulted in 

    ValueError: Cannot instantiate this tokenizer from a slow version. If it's based on sentencepiece, make sure you have sentencepiece installed.
    
 Accordingly, I modified the `requirements.txt` to include transformers with sentencepiece.

llms/README.md
```diff
@@ -38,6 +38,8 @@ To see a description of all the arguments you can do:
 >>> help(generate)
 ```
 
+For a detailed example of response generation with optional arguments, see `mlx_lm/examples/generate_response.py`.
```
awni, line 41: One more thing. That would be better as a link.
```suggestion
Check-out the [generation example](https://github.com/ml-explore/mlx-examples/tree/main/llms/mlx_lm/examples/generate_response.py) to see how to use the API in more detail. 
```

a-wozniakowski:
The PR is currently pending due to a CircleCI job on hold. Could someone with the necessary permissions take a look at this?

awni:
Sorry for the delay! Tests are running and we can merge this once they are clear. Thanks for the contribution, very nice!

awni (approved):

PR #856, opened 2024-06-28
Add recurrent gemma

awni:
A port of recurrent Gemma to spice things up a little. Runs pretty well on an M1 max.

### DIFF
--- a/llms/mlx_lm/models/recurrent_gemma.py
+++ b/llms/mlx_lm/models/recurrent_gemma.py
@@ -1,0 +1,487 @@
+import math
+from dataclasses import dataclass
+from typing import List, Literal, Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .base import BaseModelArgs, KVCache, create_additive_causal_mask
+
+
+class RecurrentCache:
+
+    def __init__(self):
+        self._cache = (None, None)
+
+    def __getitem__(self, idx):
+        return self._cache[idx]
+
+    def update(self, conv_state, recurrent_state):
+        self._cache = (conv_state, recurrent_state)
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    model_type: str
+    hidden_size: int
+    attention_bias: bool
+    conv1d_width: int
+    embeddings_scale_by_sqrt_dim: bool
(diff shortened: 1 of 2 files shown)


llms/mlx_lm/models/recurrent_gemma.py
```diff
@@ -0,0 +1,505 @@
...
+        self.keys = None
+        self.values = None
+        self.offset = 0
+        self.window_size = window_size
+
+    def update_and_fetch(self, keys, values):
+        # TODO consider using rotating buffer here
+        # especially for very long generations
+        def _update(x, v):
+            t = x.shape[2] - self.window_size
```
angeloskath, line 63: Should that also have `- v.shape[2]` ie `x.shape[2] - self.window_size - v.shape[2]` such that the returned value is of size `self.window_size` at axis 2?

angeloskath (approved):
This is very nice and refreshing!

llms/mlx_lm/models/recurrent_gemma.py
awni: A couple things, the context size is actually `window_size + 1`. That's why there is no `+1` in there. Also `v.shape[2]` is always 1 so it doesn't matter too much. But if `v.shape[2] > 1` then my understanding is we don't want to truncate the cache which would drop some tokens that we are allowed to attend to, but rather rely on the attention mask. That said, I haven't tested this code with v.shape[2] > 1 and I think its probably broken elsewhere (likely in the conv cache).

lin72h:
Awesome work! Just curious does the newly released 9b recurrent gemma also work?
[recurrentgemma-9b](https://huggingface.co/google/recurrentgemma-9b)

awni:
Yes! It should work, but I don't think I tried it so it would be good to confirm.

awni:
Indeed you're not missing anything. The 9B config is slightly different :\, so we'll need a patch for that.

awni:
Ok I added some fixes to https://github.com/ml-explore/mlx-examples/pull/877

lin72h:
Nice job! 🍻

chahn:
Thanks a lot for the fix @awni!

PR #867, opened 2024-07-04
Adding the get_embeddings method

Goekdeniz-Guelmez:
Added the `get_input_embeddings` method to the LLM codebases. The purpose of this method is to provide direct access to the input embeddings, offering several key benefits:

### Custom Processing:
- Allows for custom operations or transformations on embeddings before they proceed to subsequent layers, enhancing flexibility.
 
### Feature Extraction:
- Enables the extraction of embeddings for downstream tasks like clustering or classification without retraining the entire model, particularly useful with pre-trained models.

### (maybe future plans) Transfer Learning:
- For selective fine-tuning of certain layers or embeddings during transfer learning.

### Facilitates Experimentation:
- Supports research and development efforts by enabling straightforward experimentation with different types of embeddings and embedding spaces.

if the `--generate-embeddings` argument is not added, it will then generate normally.

```sh
python -m mlx_lm.generate \
    --model model_name \
    --generate-embeddings \
    --prompt "This is a input"
```

will return the embeddings for future processing and prints them out too.

```text
Generated Embeddings:
array([[0, 0, -0.116699, ..., 0.0473633, 0.0952148, -0.0473633],
       [0.109375, -0.0218506, 0.0218506, ..., -0.0449219, 0, -0.0672607],
       [-0.0712891, -0.0712891, 0, ..., 0, -0.0996094, 0.0500488],
       [0.0615234, 0.0615234, 0, ..., -0.0529785, 0.0529785, 0],
       [-0.0288696, 0.0481567, -0.0673828, ..., 0, -0.0219116, 0.0109863]], dtype=float16)
```

### DIFF
--- a/llms/mlx_lm/generate.py
+++ b/llms/mlx_lm/generate.py
@@ -4,7 +4,7 @@ import argparse
 
 import mlx.core as mx
 
-from .utils import generate, load
+from .utils import generate, generate_embeddings, load
 
 DEFAULT_MODEL_PATH = "mlx_model"
 DEFAULT_PROMPT = "hello"
@@ -32,6 +32,11 @@ def setup_arg_parser():
         "--trust-remote-code",
         action="store_true",
         help="Enable trusting remote code for tokenizer",
+    )
+    parser.add_argument(
+        "--generate-embeddings",
+        action="store_true",
+        help="Generate the Embeddings from the model",
     )
     parser.add_argument(
         "--eos-token",
@@ -145,16 +150,19 @@ def main():
 
     formatter = colorprint_by_t0 if args.colorize else None
 
-    generate(
-        model,
-        tokenizer,
-        prompt,
-        args.max_tokens,
-        verbose=True,
-        formatter=formatter,
-        temp=args.temp,
-        top_p=args.top_p,
-    )
+    if generate_embeddings:
+        generate_embeddings(model, tokenizer, prompt)
+    else:
+        generate(
+            model,
+            tokenizer,
+            prompt,
+            args.max_tokens,
+            verbose=True,
+            formatter=formatter,
+            temp=args.temp,
+            top_p=args.top_p,
+        )
 
 
 if __name__ == "__main__":
--- a/llms/mlx_lm/LORA.md
+++ b/llms/mlx_lm/LORA.md
@@ -96,6 +96,16 @@ mlx_lm.generate \
     --prompt "<your_model_prompt>"
 ```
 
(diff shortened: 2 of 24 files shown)


llms/mlx_lm/generate.py
```diff
@@ -145,16 +150,19 @@ def main():
...
-        tokenizer,
-        prompt,
-        args.max_tokens,
-        verbose=True,
-        formatter=formatter,
-        temp=args.temp,
-        top_p=args.top_p,
-    )
+    if generate_embeddings:
+        generate_embeddings(model, tokenizer, prompt)
```
awni, line 154: Seems pretty unlikely to me that someone would use this in the CLI, so maybe we don't add it here? Wdyt?

awni:
Can you say a bit more about what you would use this for? 

> Enables the extraction of embeddings for downstream tasks like clustering or classification without retraining the entire model, particularly useful with pre-trained models.

I don't think it's super common to use the input embeddings for downstream tasks. Are you using them for anything right now?

llms/mlx_lm/generate.py
Goekdeniz-Guelmez: Yea that does seem unlikely, I can update the server file, that makes the most sense.

Goekdeniz-Guelmez:
Thanks for your input! While its not the most common practice of feature extraction in the downstream tasks like classification and clustering, but using input embeddings can be quite powerful for certain applications. Currently, I'm using them for embedding search to improve the precision of similarity-based retrieval. If you have any more questions or suggestions, I'd love to discuss them further!

Goekdeniz-Guelmez:
> Can you say a bit more about what you would use this for?
> 
> > Enables the extraction of embeddings for downstream tasks like clustering or classification without retraining the entire model, particularly useful with pre-trained models.
> 
> I don't think it's super common to use the input embeddings for downstream tasks. Are you using them for anything right now?

Thanks for your input! While its not the most common practice of feature extraction in the downstream tasks like classification and clustering, but using input embeddings can be quite powerful for certain applications. Currently, I'm using them for embedding search to improve the precision of similarity-based retrieval. If you have any more questions or suggestions, I'd love to discuss them further!

awni:
@Goekdeniz-Guelmez I think this is too niche to merit adding a new API for it and a constraint on models. Most of this can be built outside of MLX LM without too much difficulty (e.g. the same `generate_embeddings` API and then using something like switch statement for the models to select the appropriate layer.

If we fine that more people are looking for this kind of utility we can re-evaluate (feel free to open an issue or discussion about it if you think it is worth following up on).

PR #871, opened 2024-07-04
Add support for InternLM-2.5

Blaizzy:
This PR adds support for InternLM-2.5
Twitter: @Prince_Canuma

### DIFF
--- a/llms/mlx_lm/models/base.py
+++ b/llms/mlx_lm/models/base.py
@@ -2,6 +2,47 @@ import inspect
 from dataclasses import dataclass
 
 import mlx.core as mx
+import mlx.nn as nn
+
+
+class DynamicNTKScalingRoPE(nn.Module):
+    """Implements the rotary positional encoding with Dynamic NTK scaling."""
+
+    def __init__(
+        self,
+        dims: int,
+        max_position_embeddings: int = 2048,
+        traditional: bool = False,
+        base: float = 10000,
+        scale: float = 1.0,
+    ):
+        super().__init__()
+        self.max_position_embeddings = max_position_embeddings
+        self.original_base = base
+        self.dims = dims
+        self.traditional = traditional
+        self.scale = scale
+
+    def extra_repr(self):
+        return f"{self.dims}, traditional={self.traditional}, max_position_embeddings={self.max_position_embeddings}, scaling_factor={self.scaling_factor}"
+
+    def __call__(self, x, offset: int = 0):
+        seq_len = x.shape[1] + offset
+        if seq_len > self.max_position_embeddings:
+            self.base = self.original_base * (
+                (self.scale * seq_len / self.max_position_embeddings) - (self.scale - 1)
+            ) ** (self.dims / (self.dims - 2))
+        else:
+            self.base = self.original_base
+
+        return mx.fast.rope(
+            x,
+            self.dims,
+            traditional=self.traditional,
+            base=self.base,
+            scale=self.scale,
+            offset=offset,
+        )
 
 
 def create_additive_causal_mask(N: int, offset: int = 0):
--- a/llms/mlx_lm/models/base.py
+++ b/llms/mlx_lm/models/base.py
@@ -29,17 +29,17 @@ class DynamicNTKScalingRoPE(nn.Module):
     def __call__(self, x, offset: int = 0):
         seq_len = x.shape[1] + offset
         if seq_len > self.max_position_embeddings:
-            self.base = self.original_base * (
+            base = self.original_base * (
                 (self.scale * seq_len / self.max_position_embeddings) - (self.scale - 1)
             ) ** (self.dims / (self.dims - 2))
         else:
-            self.base = self.original_base
+            base = self.original_base
 
         return mx.fast.rope(
             x,
             self.dims,
             traditional=self.traditional,
-            base=self.base,
+            base=base,
             scale=self.scale,
             offset=offset,
         )
--- a/llms/mlx_lm/models/base.py
+++ b/llms/mlx_lm/models/base.py
@@ -2,47 +2,6 @@ import inspect
 from dataclasses import dataclass
 
 import mlx.core as mx
-import mlx.nn as nn
-
-
-class DynamicNTKScalingRoPE(nn.Module):
-    """Implements the rotary positional encoding with Dynamic NTK scaling."""
-
-    def __init__(
-        self,
-        dims: int,
-        max_position_embeddings: int = 2048,
-        traditional: bool = False,
-        base: float = 10000,
-        scale: float = 1.0,
-    ):
-        super().__init__()
-        self.max_position_embeddings = max_position_embeddings
-        self.original_base = base
-        self.dims = dims
-        self.traditional = traditional
-        self.scale = scale
-
-    def extra_repr(self):
-        return f"{self.dims}, traditional={self.traditional}, max_position_embeddings={self.max_position_embeddings}, scaling_factor={self.scaling_factor}"
-
-    def __call__(self, x, offset: int = 0):
-        seq_len = x.shape[1] + offset
-        if seq_len > self.max_position_embeddings:
-            base = self.original_base * (
-                (self.scale * seq_len / self.max_position_embeddings) - (self.scale - 1)
-            ) ** (self.dims / (self.dims - 2))
-        else:
-            base = self.original_base
-
-        return mx.fast.rope(
-            x,
-            self.dims,
-            traditional=self.traditional,
-            base=base,
-            scale=self.scale,
-            offset=offset,
-        )
 
 
 def create_additive_causal_mask(N: int, offset: int = 0):
--- a/llms/mlx_lm/models/internlm2.py
+++ b/llms/mlx_lm/models/internlm2.py
@@ -27,14 +27,13 @@ class ModelArgs(BaseModelArgs):
         if self.num_key_value_heads is None:
             self.num_key_value_heads = self.num_attention_heads
 
+
         if self.rope_scaling:
-            required_keys = {"factor", "type"}
-            if not all(key in self.rope_scaling for key in required_keys):
-                raise ValueError(f"rope_scaling must contain keys {required_keys}")
-
             if self.rope_scaling["type"] != "linear":
-                raise ValueError("rope_scaling 'type' currently only supports 'linear'")
-
+                print(
+                    "[WARNING] rope_scaling 'type' currently only supports 'linear' and 'su'; setting rope scaling to false."
+                )
+                self.rope_scaling = None
 
 class Attention(nn.Module):
     def __init__(self, args: ModelArgs):
--- a/llms/mlx_lm/models/internlm2.py
+++ b/llms/mlx_lm/models/internlm2.py
@@ -14,10 +14,10 @@ class ModelArgs(BaseModelArgs):
     num_hidden_layers: int
     intermediate_size: int
     num_attention_heads: int
-    max_position_embeddings: int
     rms_norm_eps: float
     vocab_size: int
     bias: bool = True
+    max_position_embeddings: int = 32768
     num_key_value_heads: int = None
     rope_theta: float = 10000
     rope_traditional: bool = False
--- a/llms/mlx_lm/models/internlm2.py
+++ b/llms/mlx_lm/models/internlm2.py
@@ -4,7 +4,7 @@ from typing import Dict, Optional, Tuple, Union
 import mlx.core as mx
 import mlx.nn as nn
 
-from .base import BaseModelArgs
+from .base import BaseModelArgs, DynamicNTKScalingRoPE
 
 
 @dataclass
@@ -14,6 +14,7 @@ class ModelArgs(BaseModelArgs):
     num_hidden_layers: int
     intermediate_size: int
     num_attention_heads: int
+    max_position_embeddings: int
     rms_norm_eps: float
     vocab_size: int
     bias: bool = True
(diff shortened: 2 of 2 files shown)


awni:
What si this fix for @Blaizzy? It looks a bit odd to me to turn off RoPE entirely when you get a type other than `linear`? Which model was causing the issue? Maybe we need to support additional RoPE scaling types?

Blaizzy:
Yes, we need to support an additional rope scaling.

It's called `dynamic scaling`.

The model is:

https://huggingface.co/internlm/internlm2_5-7b-chat


I'm planning on implementing it during the weekend. But noticed that the model works without it so I choose that as an initial patch.

Blaizzy:
Adding dynamic scaling is a small change. Just need to sit down when I get some time.

awni:
Cool [I see it here](https://huggingface.co/internlm/internlm2_5-7b-chat/blob/main/modeling_internlm2.py#L140-L155).  Looks like it only applies after `max_position_embeddings` which is [pretty large for this model](https://huggingface.co/internlm/internlm2_5-7b-chat/blob/main/config.json#L18) so it makes sense that it works without it.

Blaizzy:
> Cool [I see it here](https://huggingface.co/internlm/internlm2_5-7b-chat/blob/main/modeling_internlm2.py#L140-L155). Looks like it only applies after `max_position_embeddings` which is [pretty large for this model](https://huggingface.co/internlm/internlm2_5-7b-chat/blob/main/config.json#L18) so it makes sense that it works without it.

Yes, exactly. This scaling is going to be import to support the 1M context version of this model. 

https://huggingface.co/internlm/internlm2_5-7b-chat-1m

By default this model has a context of 262K but can reach 1M with dynamic scaling.

Blaizzy:
@awni, I have question:
Is there any Mac with apple silicon that can support 1M context?

Or in other words, is realistic to support LLMs with +1M context on a Mac, at the moment?

awni:
I don't think 1M is too relevant here. But I see in the config that the scaling starts to apply after `max_position_embeddings` which is `32768`. See here https://huggingface.co/internlm/internlm2_5-7b-chat/blob/main/config.json#L18

Blaizzy:
> I don't think 1M is too relevant here. But I see in the config that the scaling starts to apply after `max_position_embeddings` which is `32768`. See here https://huggingface.co/internlm/internlm2_5-7b-chat/blob/main/config.json#L18

Thanks, that I understand.

Since it works without dynamic scaling. 

How do we proceed?

Blaizzy:
@awni how about this:

```python
import mlx.nn as nn

class DynamicNTKScalingRoPE(nn.Module):
    """Implements the rotary positional encoding with Dynamic NTK scaling."""
    def __init__(
        self,
        dims: int,
        max_position_embeddings: int = 2048,
        traditional: bool = False,
        base: float = 10000,
        scale: float = 1.0,
    ):
        super().__init__()
        self.max_position_embeddings = max_position_embeddings
        self.original_base = base
        self.dims = dims
        self.traditional = traditional
        self.scale = scale

    def extra_repr(self):
        return f"{self.dims}, traditional={self.traditional}, max_position_embeddings={self.max_position_embeddings}, scaling_factor={self.scaling_factor}"

    def __call__(self, x, offset: int = 0):
        seq_len = x.shape[1] + offset
        if seq_len > self.max_position_embeddings:
            self.base = self.original_base * (
                (self.scale * seq_len / self.max_position_embeddings) - (self.scale - 1)
            ) ** (self.dims / (self.dims - 2))
        else:
            self.base = self.original_base

        return mx.fast.rope(
            x,
            self.dims,
            traditional=self.traditional,
            base=self.base,
            scale=self.scale,
            offset=offset,
        )

```

awni:
@Blaizzy sorry for the delay. That looks great to me! Do you mind adding it to this PR?

Blaizzy:
Sure thing!

Blaizzy:
@awni done ✅

awni:
> I have tested it, and it works!

Just curious, how did you test that?

llms/mlx_lm/models/base.py
```diff
@@ -2,6 +2,47 @@
...
+        self.traditional = traditional
+        self.scale = scale
+
+    def extra_repr(self):
+        return f"{self.dims}, traditional={self.traditional}, max_position_embeddings={self.max_position_embeddings}, scaling_factor={self.scaling_factor}"
+
+    def __call__(self, x, offset: int = 0):
+        seq_len = x.shape[1] + offset
+        if seq_len > self.max_position_embeddings:
+            self.base = self.original_base * (
```
awni, line 32: No need to make this an instance variable. Just use `base` instead of `self.base`

Blaizzy:
> > I have tested it, and it works!
> 
> Just curious, how did you test that?

I ran it and also analysed how it changes the rope_theta.

llms/mlx_lm/models/base.py
Blaizzy: Done ✅

Blaizzy:
@awni btw, whenever you want to merge Dynamic NTK Scaling Rope to core MLX. 

Please ping me, I would like to do it :)

awni:
Here is a test which suggests the implementation is incorrect.

```python
import mlx.core as mx
import mlx.nn as nn
import numpy as np
import torch

class InternLM2RotaryEmbedding(torch.nn.Module):
    """Rotary Position Embedding for the InternLM2 model. Credits to the Reddit user /u/lucidrains."""

    def __init__(self, dim, max_position_embeddings=2048, base=10000, device=None, scaling_factor=1.0):
        super().__init__()
        self.scaling_factor = scaling_factor
        self.dim = dim
        self.max_position_embeddings = max_position_embeddings
        self.base = base
        inv_freq = 1.0 / (self.base ** (torch.arange(0, self.dim, 2, dtype=torch.int64).float().to(device) / self.dim))
        self.register_buffer("inv_freq", inv_freq, persistent=False)
        # For BC we register cos and sin cached
        self.max_seq_len_cached = max_position_embeddings

    @torch.no_grad()
    def forward(self, x, position_ids):
        # x: [bs, num_attention_heads, seq_len, head_size]
        inv_freq_expanded = self.inv_freq[None, :, None].float().expand(position_ids.shape[0], -1, 1)
        position_ids_expanded = position_ids[:, None, :].float()
        # Force float32 since bfloat16 loses precision on long contexts
        # See https://github.com/huggingface/transformers/pull/29285
        device_type = x.device.type
        device_type = device_type if isinstance(device_type, str) and device_type != "mps" else "cpu"
        with torch.autocast(device_type=device_type, enabled=False):
            freqs = (inv_freq_expanded.float() @ position_ids_expanded.float()).transpose(1, 2)
            emb = torch.cat((freqs, freqs), dim=-1)
            cos = emb.cos()
            sin = emb.sin()
        return cos.to(dtype=x.dtype), sin.to(dtype=x.dtype)

class InternLM2DynamicNTKScalingRotaryEmbedding(InternLM2RotaryEmbedding):
    """InternLM2RotaryEmbedding extended with Dynamic NTK scaling.
    Credits to the Reddit users /u/bloc97 and /u/emozilla"""

    def forward(self, x, position_ids):
        # difference to the original RoPE: inv_freq is recomputed when the sequence length > original length
        seq_len = torch.max(position_ids) + 1
        if seq_len > self.max_position_embeddings:
            base = self.base * (
                (self.scaling_factor * seq_len / self.max_position_embeddings) - (self.scaling_factor - 1)
            ) ** (self.dim / (self.dim - 2))
            inv_freq = 1.0 / (base ** (torch.arange(0, self.dim, 2, dtype=torch.int64).float().to(x.device) / self.dim))
            self.register_buffer("inv_freq", inv_freq, persistent=False)  # TODO joao: this may break with compilation

        cos, sin = super().forward(x, position_ids)
        return cos, sin

def rotate_half(x):
    """Rotates half the hidden dims of the input."""
    x1 = x[..., : x.shape[-1] // 2]
    x2 = x[..., x.shape[-1] // 2 :]
    return torch.cat((-x2, x1), dim=-1)

def apply_rotary_pos_emb(x, pos_ids, unsqueeze_dim=1):
    cos, sin = torch_rope(x, pos_ids)
    cos = cos.unsqueeze(unsqueeze_dim)
    sin = sin.unsqueeze(unsqueeze_dim)
    return (x * cos) + (rotate_half(x) * sin)


class DynamicNTKScalingRoPE(nn.Module):
    """Implements the rotary positional encoding with Dynamic NTK scaling."""
    def __init__(
        self,
        dims: int,
        max_position_embeddings: int = 2048,
        traditional: bool = False,
        base: float = 10000,
        scale: float = 1.0,
    ):
        super().__init__()
        self.max_position_embeddings = max_position_embeddings
        self.original_base = base
        self.dims = dims
        self.traditional = traditional
        self.scale = scale

    def extra_repr(self):
        return f"{self.dims}, traditional={self.traditional}, max_position_embeddings={self.max_position_embeddings}, scaling_factor={self.scaling_factor}"

    def __call__(self, x, offset: int = 0):
        seq_len = x.shape[1] + offset
        if seq_len > self.max_position_embeddings:
            self.base = self.original_base * (
                (self.scale * seq_len / self.max_position_embeddings) - (self.scale - 1)
            ) ** (self.dims / (self.dims - 2))
        else:
            self.base = self.original_base

        return mx.fast.rope(
            x,
            self.dims,
            traditional=self.traditional,
            base=self.base,
            scale=self.scale,
            offset=offset,
        )

x = np.random.uniform(size=(1, 1, 8, 128))
offset = 2049
pos_ids = np.arange(offset, offset+8)[None]

torch_out = apply_rotary_pos_emb(torch.tensor(x), torch.tensor(pos_ids))
mlx_rope = DynamicNTKScalingRoPE(128)
mlx_out = mlx_rope(mx.array(x), offset)

print(np.allclose(np.array(torch_out), mlx_out, rtol=1e-3, atol=1e-3))
```

Blaizzy:
@awni 

The MLX implementation of `DynamicNTKScalingRoPE` is correct, but it's sensitive to the input tensor's shape. The issue arises when the provided tensor shape doesn't match the expected shape we use during inference.

- Expected shape: `(1, seq_len, 1, dim)` ✅
- Provided shape: `(1, 1, seq_len, dim)` ❌

## Solution

We need to modify the `call` method in the `DynamicNTKScalingRoPE` MLX Module. Specifically, we need to change how we extract the sequence length.

### Current Implementation (Incorrect for the provided shape)

```python
class DynamicNTKScalingRoPE(nn.Module):
    def call(self, x, offset: int = 0):
        seq_len = x.shape[1] + offset
        # ... rest of the method
```

### Modified Implementation (Correct for the provided shape)

```python
class DynamicNTKScalingRoPE(nn.Module):
    def call(self, x, offset: int = 0):
        seq_len = x.shape[2] + offset
        # ... rest of the method
```

## Explanation

The change from `x.shape[1]` to `x.shape[2]` accounts for the difference in tensor shapes:

- In the expected shape `(1, seq_len, 1, dim)`, the sequence length is at index 1.
- In the provided shape `(1, 1, seq_len, dim)`, the sequence length is at index 2.

## Verification ✅

After making this change, we get the correct result:

```shell
True
Relative Distance (using norms): 1.9498426076714643e-05
Max Absolute Relative Difference: 0.3780472381919428
```

awni:
Thanks for double checking that!

llms/mlx_lm/models/base.py
```diff
@@ -2,6 +2,47 @@
 from dataclasses import dataclass
 
 import mlx.core as mx
+import mlx.nn as nn
+
+
+class DynamicNTKScalingRoPE(nn.Module):
```
awni, line 8: Can you move this to the internlm2 file? I don't think it's used anywhere else so let's keep it there for now to keep the base file clean.

awni (commented):
Sorry, one last comment, could you please address and then we can merge this?

llms/mlx_lm/models/base.py
Blaizzy: Done ✅

Blaizzy: I had put it there because a lot transformers models including llama use it to scale context window.

- Llama 2 and 3: https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L269
- Gemma: https://github.com/huggingface/transformers/blob/a1a34657d41627b21dddf2bf9cc55941329a60b6/src/transformers/models/gemma/modeling_gemma.py#L137
- StableLM: https://github.com/huggingface/transformers/blob/a1a34657d41627b21dddf2bf9cc55941329a60b6/src/transformers/models/stablelm/modeling_stablelm.py#L120
- Falcon: https://github.com/huggingface/transformers/blob/a1a34657d41627b21dddf2bf9cc55941329a60b6/src/transformers/models/falcon/modeling_falcon.py#L172
- Olmo: https://github.com/huggingface/transformers/blob/a1a34657d41627b21dddf2bf9cc55941329a60b6/src/transformers/models/olmo/modeling_olmo.py#L119
- gpt neo: https://github.com/huggingface/transformers/blob/a1a34657d41627b21dddf2bf9cc55941329a60b6/src/transformers/models/gpt_neox/modeling_gpt_neox.py#L541
- Phi: https://github.com/huggingface/transformers/blob/a1a34657d41627b21dddf2bf9cc55941329a60b6/src/transformers/models/phi/modeling_phi.py#L120

Blaizzy:
@awni just fixed the test.

I didn't notice it was failling

awni:
Thanks for the fix!

Blaizzy:
My pleasure! 😊

awni (approved):

PR #879, opened 2024-07-09
Validation with full data set, results in NaN validation score

jac-jim:
## Problem

The CLI arguments allow you to validate with the entire dataset by passing a negative one value, but this quickly results in a division by zero `NaN` to appear as the validation score!

## Expected Behavior

Using `-1` for the `--val-batches` argument should run the validation process using all available mini batches from the dataset.

## Observed Behavior

Upon starting a `--train` with `--val-batches -1`, the process quickly kicks back a `NaN` for the validation score!

## Proposed Remedy

Check if the `--val-batches` got a `-1` and if so, re-assign it to be the complete size of the data (validation) set.

## Testing

I tested this two-line change locally and it works.

## Why

* I have learned that `mlx_lm` offers more (in the sibling directory of this repo), although I still found the study of this file helpful.
* I was doing a "memorization" training, such that `train.jsonl` had the full dataset, `valid.jsonl` had the full dataset, and the "test" phase (with `test.jsonl`) is discarded. This way, the chat model is completely memorizing the source data.

### DIFF
--- a/lora/lora.py
+++ b/lora/lora.py
@@ -223,7 +223,7 @@ def evaluate(model, dataset, loss, tokenizer, batch_size, num_batches):
 
     # num_batches can be -1 to indicate the entire set
     index_iterator = iter(range(num_batches)) if num_batches != -1 else iter(int, 1)
-        
+
     for it, batch in zip(
         index_iterator,
         iterate_batches(dataset, tokenizer, batch_size),
--- a/lora/lora.py
+++ b/lora/lora.py
@@ -221,12 +221,11 @@ def evaluate(model, dataset, loss, tokenizer, batch_size, num_batches):
     all_losses = []
     ntokens = 0
 
-    # CLI arguments may set num_batches to -1; if so, find the true count
-    if num_batches == -1:
-        num_batches = len(dataset)
+    # num_batches can be -1 to indicate the entire set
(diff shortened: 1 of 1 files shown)


lora/lora.py
```diff
@@ -220,6 +220,11 @@ def iterate_batches(dset, tokenizer, batch_size, train=False):
 def evaluate(model, dataset, loss, tokenizer, batch_size, num_batches):
     all_losses = []
     ntokens = 0
+
+    # CLI arguments may set num_batches to -1; if so, find the true count
+    if num_batches == -1:
+        num_batches = len(dataset)
```
awni, line 226: This is not quite right. `len(dataset) != num_batches` since the dataset is not batched yet.

You can do something like what we do in MLX LM:

```python
    index_iterator = iter(range(num_batches)) if num_batches != -1 else iter(int, 1)
```

And then use that in place of `range(num_batches)`.

awni (changes requested):
Thanks for catching that. See inline comment

awni:
@jac-jim do you mind running the [formatting hooks](https://github.com/ml-explore/mlx-examples/blob/main/CONTRIBUTING.md)? Looks like the format test failed.

jac-jim:
Thank you for the awesome help with this @awni !

awni (approved):

PR #882, opened 2024-07-10
Add support for deepseek coder v2 lite

mzbac:
@awni I think this is ready for review and I tested on `deepseek-ai/DeepSeek-Coder-V2-Lite-Instruct` and it seem work as expected.
image

Note:
- The yarn rope may be suboptimal, but I'm not very experienced with it so I pretty much copied exactly the PyTorch implementation.
- The MOE layer still doesn't return the exact output as the PyTorch implementation, but it seems like it can generate correct responses.
- `group_limited_greedy` is not implemented because I am unable to run it locally.

### DIFF
--- a/llms/mlx_lm/models/deepseek_v2.py
+++ b/llms/mlx_lm/models/deepseek_v2.py
@@ -83,7 +83,6 @@ class DeepseekV2YarnRotaryEmbedding(nn.Module):
         max_position_embeddings=2048,
         base=10000,
         scaling_factor=1.0,
-        original_max_position_embeddings=4096,
         beta_fast=32,
         beta_slow=1,
         mscale=1,
@@ -94,7 +93,6 @@ class DeepseekV2YarnRotaryEmbedding(nn.Module):
         self.max_position_embeddings = max_position_embeddings
         self.base = base
         self.scaling_factor = scaling_factor
-        self.original_max_position_embeddings = original_max_position_embeddings
         self.beta_fast = beta_fast
         self.beta_slow = beta_slow
         self.mscale = mscale
@@ -133,46 +131,30 @@ class DeepseekV2YarnRotaryEmbedding(nn.Module):
             self.scaling_factor, self.mscale_all_dim
         )
 
-        emb = mx.concatenate((freqs, freqs), axis=-1)
-        self._cos_cached = (mx.cos(emb) * mscale).astype(mx.float32)
-        self._sin_cached = (mx.sin(emb) * mscale).astype(mx.float32)
+        self._cos_cached = mx.cos(freqs) * mscale
+        self._sin_cached = mx.sin(freqs) * mscale
 
-    def __call__(self, x, seq_len: None):
-        # x: [bs, num_attention_heads, seq_len, head_size]
+    def apply_rotary_pos_emb(self, x, cos, sin):
+        x1 = x[..., ::2]
+        x2 = x[..., 1::2]
+        rx1 = x1 * cos - x2 * sin
+        rx2 = x1 * sin + x2 * cos
+        return mx.concatenate([rx1, rx2], axis=-1)
+
+    def __call__(self, x, offset=0):
+        seq_len = offset + x.shape[2]
         if self.max_seq_len_cached is None or seq_len > self.max_seq_len_cached:
             self.set_cos_sin_cache(seq_len=seq_len)
 
-        return (
-            self._cos_cached[:seq_len],
-            self._sin_cached[:seq_len],
+        if self._cos_cached.dtype != x.dtype:
+            self._cos_cached = self._cos_cached.astype(x.dtype)
+            self._sin_cached = self._sin_cached.astype(x.dtype)
+
+        return self.apply_rotary_pos_emb(
+            x,
+            self._cos_cached[offset:seq_len],
+            self._sin_cached[offset:seq_len],
         )
-
-
-def mlx_rotate_half(x):
-    """Rotates half the hidden dims of the input."""
-    x1 = x[..., : x.shape[-1] // 2]
-    x2 = x[..., x.shape[-1] // 2 :]
-    return mx.concatenate((-x2, x1), axis=-1)
-
-
-def mlx_apply_rotary_pos_emb(q, k, cos, sin, position_ids, unsqueeze_dim=0):
-    cos = mx.expand_dims(cos[position_ids], axis=unsqueeze_dim)
-    sin = mx.expand_dims(sin[position_ids], axis=unsqueeze_dim)
-
-    b, h, s, d = q.shape
-    q = mx.reshape(
-        mx.transpose(mx.reshape(q, (b, h, s, d // 2, 2)), (0, 1, 2, 4, 3)), (b, h, s, d)
-    )
-
-    b, h, s, d = k.shape
-    k = mx.reshape(
-        mx.transpose(mx.reshape(k, (b, h, s, d // 2, 2)), (0, 1, 2, 4, 3)), (b, h, s, d)
-    )
-
-    q_embed = (q * cos) + (mlx_rotate_half(q) * sin)
-    k_embed = (k * cos) + (mlx_rotate_half(k) * sin)
-
-    return q_embed, k_embed
 
 
 class DeepseekV2Attention(nn.Module):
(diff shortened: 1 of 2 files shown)


llms/mlx_lm/models/deepseek_v2.py
```diff
@@ -0,0 +1,532 @@
...
+        compressed_kv = self.kv_a_proj_with_mqa(x)
+        compressed_kv, k_pe = mx.split(compressed_kv, [self.kv_lora_rank], axis=-1)
+        k_pe = k_pe.reshape(B, L, 1, self.qk_rope_head_dim).transpose(0, 2, 1, 3)
+        kv = self.kv_b_proj(self.kv_a_layernorm(compressed_kv))
+        kv = kv.reshape(B, L, self.num_heads, -1).transpose(0, 2, 1, 3)
+
+        k_nope, values = mx.split(kv, [self.qk_nope_head_dim], axis=-1)
+
+        k_pe = mx.concatenate([k_pe] * self.num_heads, axis=1)
+        cache = None
```
mzbac, line 322: cache doesn't work at the moment, I haven't have chance to look at the new kv cache

llms/mlx_lm/models/deepseek_v2.py
```diff
@@ -0,0 +1,532 @@
...
+            )
+
+    def __call__(self, x):
+        inds, scores = self.gate(x)
+        y = self.switch_mlp(x, inds)
+        y = (y * scores[..., None]).sum(axis=-2)
+        if self.config.n_shared_experts is not None:
+            y = y + self.shared_experts(x)
+
+        return y
```
mzbac, line 427: The Deepseek's greedy selection looks similar to Mixtral Moe, but I couldn't get it to work -> https://huggingface.co/deepseek-ai/DeepSeek-Coder-V2-Lite-Instruct/blob/main/modeling_deepseek.py#L393

lin72h:
OMG! It's happening @mzbac you rock!

llms/mlx_lm/models/deepseek_v2.py
mzbac: @awni , the DeepSeek's compressed_kv results in different shapes between k and v states. I think we may need to update the current KVCache implementation to allow it to work with the k-v cache in different shares.
image

awni: Seems ok to me. Is it working with the cache fix or is the MOE layer still an issue?

vovw:
Screenshot 2024-07-15 at 12 03 16 AM


it works, but the model is really slow on my M3 pro with 18gb ram

vovw:
how many tok/sec did it give you @mzbac ?

mzbac:
> how many tok/sec did it give you @mzbac ?

@vovw, the FP16 precision of the DeepSeek Coder V2 Lite will require approximately 25GB of VRAM. On your machine with only 18GB, this will result in slow swapping. You could try quantizing it to 8bit to avoid the swapping

mzbac:
@awni, I think the model's Moe layer is still not working properly. The model doesn't work with DeepSeek v2 chat, and I noticed that the current MLX model implementation works with 2 experts per token. However, when it works with a larger number of experts per token, the selected expert doesn't seem to work as in PyTorch implementation. I have spent the weekend trying to figure out why but unfortunately, I am unable to solve the puzzle. This may need some help from you.

mzbac:
I have tested on `deepseek-ai/DeepSeek-Coder-V2-Lite-Instruct` and `deepseek-ai/DeepSeek-V2-Lite-Chat`. `deepseek-ai/DeepSeek-Coder-V2-Lite-Instruct` seems to work, but still outputs some weird sentences sometimes. The way I test the expert per token is by directly checking the tensor output between PyTorch implementation and MLX implementation.

```import unittest
import torch
import mlx.core as mx
import mlx.nn as nn
import numpy as np
from pathlib import Path
from safetensors.torch import save_file
import glob

# Import the original PyTorch implementation
from modeling_deepseek import DeepseekV2MoE as PyTorchDeepseekV2MoE, DeepseekV2Config as PyTorchDeepseekV2Config

# Import your MLX implementation here
from model import DeepseekV2MoE as MLXDeepseekV2MoE, DeepseekV2Config as MLXDeepseekV2Config


def transfer_weights(pytorch_model, model_path: Path):
    # Save PyTorch weights to safetensor file
    state_dict = pytorch_model.state_dict()
    save_file(state_dict, str(model_path / "model.safetensors"))


def load_mlx_model(model_path: Path, mlx_config: dict) -> nn.Module:
    """
    Load weights from safetensor file and initialize the MLX model.
    """
    weight_files = glob.glob(str(model_path / "model*.safetensors"))
    if not weight_files:
        raise FileNotFoundError(f"No safetensors found in {model_path}")

    # Load weights from safetensor file(s)
    weights = {}
    for wf in weight_files:
        weights.update(mx.load(wf))

    # Initialize MLX model
    mlx_model = MLXDeepseekV2MoE(MLXDeepseekV2Config(**mlx_config))

    def sanitize(weights):
        for n, m in [("w1", "gate_proj"), ("w2", "down_proj"), ("w3", "up_proj")]:
            for k in ["weight", "scales", "biases"]:
                if f"experts.0.{m}.{k}" in weights:
                    to_join = [
                        weights.pop(
                            f"experts.{e}.{m}.{k}"
                        )
                        for e in range(mlx_config["n_routed_experts"])
                    ]
                    weights[f"switch_mlp.{m}.{k}"] = (
                        mx.stack(to_join)
                    )
        return weights
    # Load weights into MLX model
    # print(weights.keys())
    weights = sanitize(weights)
    mlx_model.load_weights(list(weights.items()))

    return mlx_model


class TestMoE(unittest.TestCase):
    def setUp(self):
        self.config = {
            "hidden_size": 512,
            "intermediate_size": 10944,
            "hidden_act": "silu",
            "num_experts_per_tok": 2,
            "n_routed_experts": 8,
            "n_shared_experts": 2,
            "routed_scaling_factor": 1.0,
            "scoring_func": "softmax",
            "aux_loss_alpha": 0.001,
            "seq_aux": True,
            "topk_method": "greedy",
            "n_group": 1,
            "topk_group": 1,
            "norm_topk_prob": False,
            "moe_intermediate_size": 1408,
            "num_hidden_layers": 27,
            "num_attention_heads": 16,
            "num_key_value_heads": 16,
            "max_position_embeddings": 163840,
            "moe_layer_freq": 1,
            "ep_size": 1
        }

        # Initialize PyTorch model
        pytorch_config = PyTorchDeepseekV2Config(**self.config)
        self.pytorch_moe = PyTorchDeepseekV2MoE(pytorch_config)
        self.pytorch_moe.eval()

        # Create a temporary directory to save and load the model
        self.temp_dir = Path("temp_model")
        self.temp_dir.mkdir(exist_ok=True)

        # Transfer weights from PyTorch to safetensor file
        transfer_weights(self.pytorch_moe, self.temp_dir)

        # Load MLX model using the load_mlx_model function
        self.mlx_moe = load_mlx_model(self.temp_dir, self.config)
        self.mlx_moe.eval()

        # Set random seed for reproducibility
        torch.manual_seed(42)
        np.random.seed(42)
        mx.random.seed(42)

    def tearDown(self):
        # Clean up temporary directory
        for file in self.temp_dir.glob("*"):
            file.unlink()
        self.temp_dir.rmdir()

    def test_moe_forward(self):
        # Create input tensor
        batch_size, seq_len = 1, 6
        input_shape = (batch_size, seq_len, self.config["hidden_size"])
        pytorch_input = torch.rand(*input_shape)
        mlx_input = mx.array(pytorch_input.numpy())

        # Get PyTorch output
        with torch.no_grad():
            pytorch_output = self.pytorch_moe(pytorch_input)

        # Get MLX output
        mlx_output = self.mlx_moe(mlx_input)

        # Convert outputs to numpy for comparison
        pytorch_output_np = pytorch_output.numpy()
        mlx_output_np = mlx_output

        # Compare outputs
        np.testing.assert_allclose(
            pytorch_output_np, mlx_output_np, rtol=1e-2, atol=1e-2)

        # Print PyTorch output shape and some values
        print(f"PyTorch output shape: {pytorch_output.shape}")
        print(
            f"PyTorch output first few values: {pytorch_output.float().flatten()[:5]}")

        # Print MLX output shape and some values
        print(f"MLX output shape: {mlx_output.shape}")
        print(f"MLX output first few values: {mlx_output.flatten()[:5]}")


if __name__ == '__main__':
    unittest.main()```

**Comment by awni** [comment]
I sent a fix to your branch. I think the only difference was the order of the softmax. They do softmax -> topk whereas you were doing topk -> softmax.

**Comment by mzbac** [comment]
@awni, the fix works 🚀. I was thinking that softmax just normalization didn't know to do it before topK and after would cause the issue.

**Comment by awni** [comment]
Great! Shall we move this out of draft and get it merged?

**Comment by awni** [comment]
I updated the Rope a bit. There was a problem with the type causing the whole inference to be done in fp32 which was very slow. It runs much faster now in fp16.

**Comment by mzbac** [comment]
> I updated the Rope a bit. There was a problem with the type causing the whole inference to be done in fp32 which was very slow. It runs much faster now in fp16.

Thanks @awni, the rope makes more sense to me now!

File: llms/mlx_lm/models/deepseek_v2.py [hunk 3]
```diff
@@ -94,7 +93,6 @@ def __init__(
         self.max_position_embeddings = max_position_embeddings
         self.base = base
         self.scaling_factor = scaling_factor
-        self.original_max_position_embeddings = original_max_position_embeddings
```
mzbac: @awni I think the original_max_position_embeddings is still needed, it's used by `yarn_find_correction_range`.

awni: Oops, I will add it back.

mzbac:
With the rope fixed, it boosted the tps from 58 t/s to 96 t/s! 🚀 
image

awni:
I added the group greedy method and tested it with the bigger model. It generates good output but is quite slow even quantized so we need to do some perf optimization for that.

awni (approved):
Thanks for the addition @mzbac

llms/mlx_lm/models/deepseek_v2.py
```diff
@@ -309,25 +308,27 @@ def __init__(self, config: ModelArgs):
...
+        if self.topk_method == "group_limited_greedy":
+            bsz, seq_len = x.shape[:2]
+            scores = scores.reshape(bsz, seq_len, self.n_group, -1)
+            group_scores = scores.max(axis=-1)
+            k = self.n_group - self.topk_group
+            group_idx = mx.argpartition(group_scores, kth=k - 1, axis=-1)[..., :k]
+            batch_idx = mx.expand_dims(mx.arange(bsz), (1, 2))
+            seq_idx = mx.expand_dims(mx.arange(seq_len), (0, 2))
+            scores[batch_idx, seq_idx, group_idx] = 0.0
+            scores = scores.reshape(bsz, seq_len, -1)
```
mzbac, line 327: I really learned something from this. :)

PR #885, opened 2024-07-11
Enable Dora training with use_dora parameter

### LINKED ISSUE #880: DoRa training is never activated
See:

https://github.com/ml-explore/mlx-examples/blob/63800c8feb81710989c934a2af7677a78164256f/llms/mlx_lm/lora.py#L169

Even when the use_dora is specified as True, DoRA layers are not used by the mlx_lm trainer because the **use_dora** keyword argument to **linear_to_lora_layers** is never specified, and it defaults to False.

chimezie:
Fixes #880

### DIFF
--- a/llms/mlx_lm/lora.py
+++ b/llms/mlx_lm/lora.py
@@ -166,7 +166,7 @@ def train_model(
     model.freeze()
 
     # Convert linear layers to lora layers and unfreeze in the process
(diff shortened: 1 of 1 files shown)


awni:
Good catch! I wonder if this got dropped during a merge or something 😬

awni (approved):

PR #889, opened 2024-07-15
Keep the LM server in a valid state

awni:
Fixes #887 

If a model is loaded, then you load an invalid model and then reload the old model, the key is not updated.. so it thinks it has the old model correctly loaded even though the model and tokenizer were cleared.


angeloskath (approved):
Thanks for the fix!

PR #898, opened 2024-07-21
Introducing `--resume-training` to preserve adapter files from being overwritten

jac-jim:
Instead of trying to illustrate some undesirable behavior from `--resume-adapter-file` as a bug, I've decided to share `--resume-training` to meet my needs.

## New Feature Synopsis
This enabled `--resume-training` argument is an alternative to `--resume-adapter-file` and allows the program to continue where it left off, by selecting the highest integer valued checkpoint found in the `--adapters-path`.

After the program hits `9999999`, it should halt, and give the user some instruction on how to resume training if they so desire.

The program will refuse to start if both `--resume-adapter-file` **and** `--resume-training` are supplied, since `--resume-training` fills in the spot utilized by `--resume-adapter-file`.  The feature is accomplished by adding an `adapters_found_count` to the `TrainingArguments` class, and this offset is summed with the `it` iterations count during the training loop to determine the next filename.

### Backwards Compatibility
This feature is completely compatible, it should not change the existing function of the program.  When `--resume-adapter-file` is used this new feature should remain unnoticed.

## Example

`python -m mlx_lm.lora --train --adapter-path /My/Adapter/Path --resume-training ...` 

**EXCLUSIVE OR**

`python -m mlx_lm.lora --train --adapter-path /My/Adapter/Path --resume-adapter-file /My/Favorite/adapters.safetensors ...`

## Why
During my learning I am trying to establish metrics (objective and subjective) between checkpoints: otherwise one has to bother with managing directories, else, the checkpoints are overwritten when the training process is restarted.

## Testing
I tested locally, with `--save-every` set to `25`, and restarting a few times works: `--resume-training` finds the correct last `XXXXXXX_adapters.safetensors` file, and continues writing higher-numbered files until complete.

## Todo
* Style? I'm new here, I tried to emulate the style you guys were using (`is not None`, for example); feel free to re-organize the logic (or ask me to).
* Perceived impact: am I an edge case?
* Wiki documentation? Examples? Will this new feature add **confusion** to any docs or examples?
* Are there unit tests? I have not ran (or looked for) any tests here. Does this break any of the tests if they exist?

## Remarks
Thank you for keeping this project inclusive! This original code, documentation and PR was done by hand (without generative AI). 😎

### DIFF
--- a/llms/mlx_lm/lora.py
+++ b/llms/mlx_lm/lora.py
@@ -51,6 +51,7 @@ CONFIG_DEFAULTS = {
     "steps_per_report": 10,
     "steps_per_eval": 200,
     "resume_adapter_file": None,
+    "resume_training": False,
     "adapter_path": "adapters",
     "save_every": 100,
     "test": False,
@@ -108,6 +109,12 @@ def build_parser():
         "--resume-adapter-file",
         type=str,
         help="Load path to resume training with the given adapters.",
+    )
+    parser.add_argument(
+        "--resume-training",
+        action="store_true",
+        help="Resume training from the last saved adapter file found in the adapter path, without overwriting older adapters.",
+        default=None,
     )
     parser.add_argument(
         "--adapter-path",
@@ -169,9 +176,28 @@ def train_model(
     linear_to_lora_layers(model, args.lora_layers, args.lora_parameters, args.use_dora)
 
     # Resume training the given adapters.
-    if args.resume_adapter_file is not None:
+    if (args.resume_adapter_file and args.resume_training) is not None:
+        raise ValueError(
+            "Invalid Arguments: Please select either `--resume-adapter-file` or `--resume-training`, avoid both."
+        )
+    elif args.resume_adapter_file is not None:
         print(f"Loading pretrained adapters from {args.resume_adapter_file}")
         model.load_weights(args.resume_adapter_file, strict=False)
+    elif args.resume_training is not None:
+        if not args.adapter_path:
+            raise ValueError(
+                "Missing Argument: Please include an --adapter-path when using --resume-training, yeah?"
+            )
+        # Determine the highest integer adapter file in --adapter-path
+        search_adapter_path = Path(args.adapter_path)
+        # Cute puzzle math right here: the result is a six digit integer
+        adapters_found_count = int(
+            str(list(search_adapter_path.glob("**/*.safetensors"))[-1])[
+                : -1 * len("_adapters.safetensors")
+            ][-6:]
+        )
+    else:
+        adapters_found_count = 0
 
     print_trainable_parameters(model)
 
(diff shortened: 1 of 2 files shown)


llms/mlx_lm/tuner/trainer.py
```diff
@@ -285,13 +291,33 @@ def step(batch):
...
-            )
+
+            # Preserve all original adapter files if the count is available
+            if args.adapters_found_count == 0:
+                next_adapter_number = it
+            else:
+                next_adapter_number = args.adapters_found_count + it
+
+            # Handle the scenario, if there are ten million adapter filenames in `--resume-training` mode...
+            if next_adapter_number > 9999999 and (args.resume_training is not None):
```
awni, line 306: I don't think we need to check this case. It's so extremely unlikely..

llms/mlx_lm/lora.py
```diff
@@ -169,9 +176,28 @@ def train_model(
     linear_to_lora_layers(model, args.lora_layers, args.lora_parameters, args.use_dora)
 
     # Resume training the given adapters.
-    if args.resume_adapter_file is not None:
+    if args.resume_adapter_file is not None and args.resume_training:
+        raise ValueError(
+            "Invalid Arguments: Please select either `--resume-adapter-file` or `--resume-training`, avoid both."
```
awni, line 181:
```suggestion
            "Invalid Arguments: Use only one of `--resume-adapter-file` or `--resume-training`."
```

llms/mlx_lm/lora.py
```diff
@@ -169,9 +176,28 @@ def train_model(
...
+            raise ValueError(
+                "Missing Argument: Please include an --adapter-path when using --resume-training, yeah?"
+            )
+        # Determine the highest integer adapter file in --adapter-path
+        search_adapter_path = Path(args.adapter_path)
+        # Cute puzzle math right here: the result is a six digit integer
+        adapters_found_count = int(
+            str(list(search_adapter_path.glob("**/*.safetensors"))[-1])[
+                : -1 * len("_adapters.safetensors")
+            ][-6:]
```
awni, line 197: This is a bit confusing and I'm not sure it's right. I don't think they will be sorted numerically..

Instead can you just do:

```
adapters_found_count = len(list(search_adapter_path.glob("**/*.safetensors"))
```

awni:
I'm wondering if it makes more sense to handle this outside of MLX LM? You could write a wrapper script which finds the count and then starts from that adapter no? I worry a bit about CLI argument bloat..

awni:
I think a better way to handle this rather than introduce complexity in MLX LM is to write a wrapper script which picks up the last adapters and resumes from there. Different people prefer different work flows so that will keep things pretty flexible and simple. Closing for now, feel free to comment if you have any more thoughts on this. We can revisit if it seems like a common need.

jac-jim:
Indeed, sorry for the noise.On Aug 1, 2024, at 18:14, Awni Hannun ***@***.***> wrote:﻿
I think a better way to handle this rather than introduce complexity in MLX LM is to write a wrapper script which picks up the last adapters and resumes from there. Different people prefer different work flows so that will keep things pretty flexible and simple. Closing for now, feel free to comment if you have any more thoughts on this. We can revisit if it seems like a common need.

PR #899, opened 2024-07-21
support load model by custom get_model_classes

mzbac:
This will enable the community to utilize most of the utility functions but implement their own variant of the model, e.g. the sharding model https://github.com/mzbac/mlx_sharding/blob/main/server/server.py#L25-L66 , without having to copy the source code from mlx_lm.

@awni, please feel free to close this if you don't think it fits the purpose.


awni (approved):
Thanks for adding it! Looks good to me.

PR #902, opened 2024-07-23
Adapters loading

khushgx:
Right now to add in adapters once the server has started, we need to restart the server. I just minimally changed the ```load``` function and the ```do_POST``` function in server.py, adding in an optional argument to load so that the user can add in an adapter during runtime. I tested it with by starting a server with Mistral and then changing the model to Phi-3B w/ LoRa adapters with a POST request.

**I did not add the restriction that adapters need to be in the current working directory **


khushgx:
Hi @awni , I think the commit you added caused some failed tests?

awni (approved):
Thanks for the addition! I will merge this once the tests clear.

PR #911, opened 2024-07-25
Unify attention mask in LLMs

otriscon:
This PR introduces minor changes that help use mlx-lm's models with
advanced decoding techniques such as speculative decoding and
prompt cache reuse. They are the based on my experience building
https://github.com/otriscon/llm-structured-output on top of mlx-lm.

Currently, each model implementation in `mlx-examples/llms/models` has ad-hoc
code to create a mask for the attention mechanism. This usually takes the form:

```
    mask = None
    if h.shape[1] > 1:
        mask = nn.MultiHeadAttention.create_additive_causal_mask(h.shape[1])
        mask = mask.astype(h.dtype)
```

This creates a mask when the input consists of more than one token, but this
code assumes the multi-token input is at the beginning of inference.
If, for example, we are evaluating multiple tokens because of speculative
decoding or prompt cache reuse, this mask will not have the correct shape and
and will cause the raising of an exception in the attention computation.

Some of the models correctly implement the mask creation with code like this:

```
    mask = None
    if h.shape[1] > 1:
        mask = create_additive_causal_mask(
            h.shape[1], cache[0].offset if cache is not None else 0
        )
        mask = mask.astype(h.dtype)
```

This PR unifies the attention mask creation for all models with a new function
`create_attention_mask`, reducing code duplication and helping all models 
support inference performance enhancements like those mentioned above.

Additionally, this PR makes a small change to KVCache to allow batching.
The current implementation of the LLM key-value cache assumes that
the input batch is of size 1. Input batching (evaluating multiple
alterative inputs at the same time) can be a valuable tool for
speculative sampling and other techniques. This PR removes the hard-coded
batch size from the code that resizes the key-value cache.

While I haven't been able to test with every single model supported by mlx-lm,
I've performed an evaluation run with each of the following LLMs:

- mlx-community/Meta-Llama-3-8B-Instruct-8bit
- mlx-community/Meta-Llama-3-70B-Instruct-4bit
- mistralai/Mistral-7B-Instruct-v0.2
- mlx-community/Mixtral-8x22B-4bit
- mlx-community/Phi-3-mini-128k-instruct-8bit
- deepseek-ai/DeepSeek-Coder-V2-Lite-Instruct

### DIFF
--- a/llms/mlx_lm/models/base.py
+++ b/llms/mlx_lm/models/base.py
@@ -23,9 +23,10 @@ class KVCache:
     def update_and_fetch(self, keys, values):
         prev = self.offset
         if self.keys is None or (prev + keys.shape[2]) > self.keys.shape[2]:
+            B = keys.shape[0]
             n_steps = (self.step + keys.shape[2] - 1) // self.step
-            k_shape = (1, self.n_kv_heads, n_steps * self.step, self.k_head_dim)
-            v_shape = (1, self.n_kv_heads, n_steps * self.step, self.v_head_dim)
+            k_shape = (B, self.n_kv_heads, n_steps * self.step, self.k_head_dim)
+            v_shape = (B, self.n_kv_heads, n_steps * self.step, self.v_head_dim)
             new_k = mx.zeros(k_shape, keys.dtype)
             new_v = mx.zeros(v_shape, values.dtype)
             if self.keys is not None:
--- a/llms/mlx_lm/models/base.py
+++ b/llms/mlx_lm/models/base.py
@@ -1,5 +1,6 @@
 import inspect
 from dataclasses import dataclass
+from typing import List, Optional
 
 import mlx.core as mx
 import mlx.nn as nn
@@ -64,7 +65,7 @@ def create_additive_causal_mask(N: int, offset: int = 0):
     return mask * -1e9
 
 
-def create_attention_mask(h: mx.array, cache: list[KVCache] = None):
+def create_attention_mask(h: mx.array, cache: Optional[List[KVCache]] = None):
     T = h.shape[1]
     if T > 1:
         # Input consists of multiple tokens, create a causal mask so that prior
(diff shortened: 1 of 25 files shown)


llms/mlx_lm/models/base.py
```diff
@@ -60,3 +55,26 @@ def from_dict(cls, params):
                 if k in inspect.signature(cls).parameters
             }
         )
+
+
+def create_additive_causal_mask(N: int, offset: int = 0):
+    if offset > 0:
```
otriscon, line 61: Here I'm keeping the existing code for creating an offset mask in order to minimize change. An alternative would be to slice the output of `nn.MultiHeadAttention.create_additive_causal_mask`:

```
def create_additive_causal_mask(N: int, offset: int = 0):
    return nn.MultiHeadAttention.create_additive_causal_mask(N)[offset:,]
```

llms/mlx_lm/models/base.py
```diff
@@ -60,3 +55,26 @@ def from_dict(cls, params):
...
+        rinds = mx.arange(offset + N)
+        linds = mx.arange(offset, offset + N) if offset else rinds
+        mask = linds[:, None] < rinds[None]
+        return mask * -1e9
+    return nn.MultiHeadAttention.create_additive_causal_mask(N)
```
otriscon, line 66: A simpler alternative for this function would be to slice the output of `nn.MultiHeadAttention.create_additive_causal_mask`:

```
def create_additive_causal_mask(N: int, offset: int = 0):
    return nn.MultiHeadAttention.create_additive_causal_mask(N)[offset:,]
```

I'm not sure why the original author chose to create the offset mask from `aranges`, so to minimize change/risk I'm keeping the original code when there's an offset.

awni (approved):
Looks awesome, thanks!

otriscon:
Thank you @awni 

I've simplified the mask creation to use just one code path as per your comment.
I've also addressed the CI linter error which was caused by using new-style type annotations.

awni:
Awesome, very nice addition!

fblissjr:
This is awesome @otriscon ! Looking at your https://github.com/otriscon/llm-structured-output repo and really like it.

PR #914, opened 2024-07-25
Allow the entire model to be targed for LoRA and DoRA fine tuning: LoRA and DoRA embeddings with small DoRALinear bug fix

zaithottakath:
- Added in `LoRAEmbedding` and `DoRAEmbedding` with tests so that embeddings can be targeted for  fine tuning.
- Added in the ability to target all `Linear` and `Embedding` modules regardless of if they are in `model.layers` allowing both the embeddings and the `lm_head` to be targeted for fine tuning, allowing a nearly full LoRA or DoRA fine tune of the model.
- Fixed a bug with `DoRALinear` that sets the wrong `self.m` value due to it not being recalculated when the `Linear` layer is changed in `DoRALinear.from_linear`

I checked huggingface's PEFT library for how they handle DoRA for embeddings and there is [still an open ticket for it](https://github.com/huggingface/peft/issues/1677). I wasn't able to find any reference implementations, so this could be the first example of that.


llms/mlx_lm/tuner/dora.py
```diff
@@ -93,3 +100,107 @@ def __call__(self, x):
...
+
+        lora_a = (self.scale * self.lora_a).astype(dtype)
+        lora_b = self.lora_b.astype(dtype)
+        weight = weight + lora_a @ lora_b
+        norm_scale = self.m / mx.linalg.norm(weight, axis=1)
+        fused_embedding.weight = norm_scale[:, None] * weight
+
+        return fused_embedding
+
+    fuse = to_embedding
```
awni, line 147: Why don't we just call it `fuse` and get rid of the `to_embedding` and `to_linear` etc. That seems cleaner, no?

zaithottakath: Agreed!

zaithottakath: I also changed `from_embedding` and `from_linear` to `from_base`

llms/mlx_lm/tuner/utils.py
```diff
@@ -130,7 +132,15 @@ def to_lora(layer):
 
     for l in model.layers[num_layers - num_lora_layers :]:
         lora_layers = [(k, to_lora(m)) for k, m in l.named_modules() if k in keys]
-        l.update_modules(tree_unflatten(lora_layers))
+        if lora_layers:
+            l.update_modules(tree_unflatten(lora_layers))
+
+    non_layer_modules = [
+        (k, m) for k, m in model.named_modules() if "model.layers" not in k
+    ]
```
awni, line 140: This is a bug I think. The model keys are not typically `model.layers.*`. `model.layers` is a property for easy access to the models hidden layers / blocks.

Maybe the keys for the embedding and linear layers should be the full path to make this simpler? Not sure.. maybe there is a cleaner way.

zaithottakath: Sorry, I'm not quite sure I understand. Here I'm trying to gather all `named_modules()` that are not in in the hidden layers`"model.layers" not in k`. So `non_layer_modules` should be all modules - hidden layer modules. So now the `lm_head` and `embeddings` can be targeted from the set of modules that are not the hidden layer modules.

llms/tests/test_lora.py
```diff
@@ -60,6 +65,22 @@ def check_config(params):
         params["keys"] = ["self_attn.k_proj"]
         check_config(params)
 
+        params["keys"] = ["lm_head"]
```
zaithottakath, line 68: The full path the layer that is being targeted for both the LM head (if the model has one) and the embeddings. @awni

llms/tests/test_lora.py
```diff
@@ -60,6 +65,22 @@ def check_config(params):
...
+        check_config(
+            params,
+            expected_trainable_parameters=(
+                params["rank"] * (args.hidden_size + args.vocab_size)
+            ),
+        )
+
+        params["keys"] = ["model.embed_tokens"]
```
zaithottakath, line 76: @awni same here. This is the full path to the embeddings.

llms/mlx_lm/tuner/utils.py
awni: Sorry for the delay. What I mean to say is `model.layers` is not always the path to the layers. Sometimes it's like `model.blocks` or something else. So `non_layer_modules` will not always be "all modules - hidden layer modules".

awni: If the keys for the non layer modules are always the full path then the above line is not really necessary is it?

zaithottakath: Great point about it being unnecessary. Will update.

zaithottakath: @awni updated and test still passing.

llms/mlx_lm/tuner/dora.py
```diff
@@ -25,10 +25,10 @@ def from_linear(
             dropout=dropout,
             scale=scale,
         )
-        dora_lin.linear = linear
+        dora_lin.set_linear(linear)
```
awni, line 28: Good catch with that bug!

awni (approved):
Very nice addition, thanks!

PR #915, opened 2024-08-01
Adding simple GAN network 

Shubbair:
add simple GAN network with the implementation step by step in jupyter notebook and arrange the code in main file


Shubbair:
hello ,
i made here simple GAN network just accept it

PR #932, opened 2024-08-13
Enable fine tuning for deepseek

awni:
Simple fix, as title.


barronalex (approved):
👍

mzbac:
@awni you are so fast! I was thinking of adding this for the weekend. I have a question about this change: from my understanding, DeepSeek v2 has conditions regarding whether to include `q_proj` or `q_a_proj` for the model. If we enable all those linear layers without checking the configuration, could it cause any exceptions for LoRA?

awni:
It shouldn't cause any exceptions. It just adds a layer if it matches any of the supplied keys, you can see [here](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/tuner/utils.py#L145). So I listed all the possible keys since we want to cover both cases.

PR #936, opened 2024-08-15
output the tokens/second performance for the llama example

byjlw:
People want to know the performance of MLX. This gives them the Tokens/Second metric they can use to compare with other platforms.

### DIFF
--- a/llms/llama/llama.py
+++ b/llms/llama/llama.py
@@ -212,6 +212,8 @@ def generate(args):
     skip = 0
     prompt_processing = None
     tokens = []
+    start_time = time.time()
+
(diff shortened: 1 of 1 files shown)


awni:
You should use mlx-lm for that:

```
pip install mlx-lm
```

Then run it with:

```
mlx-lm.generate --model mlx-community/Meta-Llama-3.1-8B-Instruct-4bit --prompt "Hello!" --temp 0.0 --max-tokens 512
```

That will give you lot's of details about the time / memory use etc.

We're not planning to update the self-contained minimal examples unless there is a bug.

PR #945, opened 2024-08-19
Use fast rope

awni:
- Fixes the llama3.1 NTK RoPE implementation which was not correct
- Uses fast rope where possible

Will post some benchmarks shortly

### DIFF
--- a/llms/mlx_lm/models/deepseek_v2.py
+++ b/llms/mlx_lm/models/deepseek_v2.py
@@ -68,13 +68,12 @@ def yarn_get_mscale(scale=1, mscale=1):
     return 0.1 * mscale * math.log(scale) + 1.0
 
 
-def yarn_linear_ramp_mask(min, max, dim):
-    if min == max:
-        max += 0.001  # Prevent singularity
+def yarn_linear_ramp_mask(mn, mx, dim):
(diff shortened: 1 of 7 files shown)


awni:
### Benchmarks

Timings on M2 Ultra, all models are 4-bit.

Model       | Slow RoPE | Fast RoPE
----- | ----- | ------
Llama 3.1 8B | 86.6 | 105.2
Phi-3 128k | 102.9 | 173.0
DeepSeekV2 Chat Lite | 100.3 | 111.5

awni:
This is ready for review. The tests are failing because it needs a pre-released MLX.

The only sticking point is what to do about the frequency selection in the SuRope. Right now I hardcoded it to the long frequencies. It works well for both short and long generations whereas the short frequencies don't work for long generations. 

It's possible if we use the short frequencies we could get better results for short generations. Setting the frequencies dynamically is tricky because it requires knowing in advance the length that will be generated and that's both cumbersome to implement and only possible to approximate (since one can set `--max-tokens large_number` but only generate a small number of tokens).

angeloskath (approved):
Pretty nice cleanup!

And thanks for the SD update we forgot it when you changed group norm to use `mx.fast`.

PR #948, opened 2024-08-21
feat(mlx_lm)!: batch_generate

llllvvuu:
This is based on @willccbb's implementation at https://github.com/willccbb/mlx_parallm.

BREAKING CHANGE: `generate_step` takes `(bs, seq_len)` instead of `(seq_len,)`. In particular, `sampler` and `logits_processors` will need to handle logits of shape `(bs, vocab_size)` instead of `(vocab_size,)`.


llllvvuu:
Kind of interesting: for quantized models, the throughput is doesn't go up a lot between bs=1,2,3,4, but then it starts to go up a lot, which is the opposite of what I expected intuitively. I observe the same on @willccbb's original repo.

llms/mlx_lm/utils.py
```diff
@@ -334,56 +341,85 @@ def generate(
...
         tokenizer = TokenizerWrapper(tokenizer)
 
-    if verbose:
-        print("=" * 10)
-        print("Prompt:", prompt)
-
-    prompt_tokens = mx.array(tokenizer.encode(prompt))
-    detokenizer = tokenizer.detokenizer
+    if is_batch:
+        tokenizer._tokenizer.padding_side = "left"
```
awni, line 349: I see that we left pad shorter prompts here which makes sense. But one thing that I'm wondering is how this is handled in the causal models if at all? Shouldn't the causal mask take into account the padding?

awni:
I think it makes sense to minimize the complexity to the `generate` function (which is becoming a bit spaghetti) to split out the batched generation into a separate function called `batch_generate`. I would simplify that function to have fewer arguments (like no `formatter`, no printing during generation, `verbose` only prints the timings (e.g. as you have it now).

Also maybe more tricky is the fact that I think for this to be correct, the causal masks need to consider the left padding in the input (please correct me if I'm wrong about that). This has two implications:
1. Probably we'd need to add a mask parameter to the model `__call__` functions and provide an appropriately constructed mask for the batch case.
2. The Rotating KV cache will be broken in this case (it keeps the initial tokens which would be the padded tokens) and when rotates the mask would need to be updated to consider the padding (which is a bit complicated/tedious). In this case I may suggest disabling this option entirely..

Let me know what you think about the above.

llms/mlx_lm/utils.py
llllvvuu: I didn't handle it, generation seems OK without it but indeed to be correct I should consume the `tokenizer._tokenizer(prompt, padding=True)["attention_mask"]`. To do this I would need to update our model APIs to have `attention_mask` as an input similar to how transformers has `model.generate` taking `attention_mask`. Probably this is involves hitting every file in `models/`. But shouldn't mostly be copy/paste. I can look into it.

llllvvuu:
> I think it makes sense to minimize the complexity to the `generate` function (which is becoming a bit spaghetti) to split out the batched generation into a separate function called `batch_generate`. I would simplify that function to have fewer arguments (like no `formatter`, no printing during generation, `verbose` only prints the timings (e.g. as you have it now).

Makes sense to me, will implement.

> 
> Also maybe more tricky is the fact that I think for this to be correct, the causal masks need to consider the left padding in the input (please correct me if I'm wrong about that). This has two implications:
> 
> 1. Probably we'd need to add a mask parameter to the model `__call__` functions and provide an appropriately constructed mask for the batch case.

Yes, this sounds straightforward enough.

> 2. The Rotating KV cache will be broken in this case (it keeps the initial tokens which would be the padded tokens) and when rotates the mask would need to be updated to consider the padding (which is a bit complicated/tedious). In this case I may suggest disabling this option entirely..

I'll do a bit of thinking if there's an easy way to handle this, otherwise I'll remove that parameter in `batch_generate`.

Will update when these changes are ready!

awni:
@llllvvuu are you coming back to this?

llllvvuu:
hey @awni , sorry for the delay, I'd been job hunting this month. I should be able to get back to this in ~a week

awni:
No worries, just checking. I'll follow up in a week or so.

nath1295:
Just realised the attention mask has been mentioned in this PR, which is the reason I raised this issue #1044

llllvvuu:
Sorry for the delay @awni . I took advantage of https://github.com/ml-explore/mlx-examples/pull/1173 to update this PR. It is pending https://github.com/ml-explore/mlx/pull/1726 for the mask dtype.

I noticed one other potential issue: For absolute/rotary positional encodings, the position IDs of padded prompts won't start from 0. I'm not sure what the priority of this is or if it requires any change to `mx.fast.rope`

qinxuye:
Any update? we do need this to support parallel generation.

awni:
Will get to this soon. Sorry for the delay.

qinxuye:
Sorry to follow up again, but we really need this capability. May I ask about the progress?

awni:
@qinxuye could you say a bit more about what you are looking for?

qinxuye:
> @qinxuye could you say a bit more about what you are looking for?

Oh, now we want to support continuous batching for MLX in Xinference, but if MLX cannot support batch inference, we don't know how to move forward.

willccbb:
the code to implement batch inference is pretty simple, you can just copy the code from mlx_parallm (or probably better: https://github.com/N8python/gsm-mlx) and add it to your codebase. the hard part is orchestrating the continuous batching, not the batched KV caching itself. this is largely why I stopped development on mlx_parallm, that would have been the natural next step but I didn't have an immediate need and it seemed like it'd take a lot of time to get it working properly.

qinxuye:
> the code to implement batch inference is pretty simple, you can just copy the code from mlx_parallm (or probably better: https://github.com/N8python/gsm-mlx) and add it to your codebase. the hard part is orchestrating the continuous batching, not the batched KV caching itself. this is largely why I stopped development on mlx_parallm, that would have been the natural next step but I didn't have an immediate need and it seemed like it'd take a lot of time to get it working properly.

Thanks for your advice, that makes sense, we could just implement the batch logic first, and about the continuous batching, we already have some code base in our project, I will work on it to see how to integrate them together.

llllvvuu:
@qinxuye it looks like there is a new repo for mlx-lm, I'll take a look sometime and see if I can open something over there. I also never did resolve the position ids thing (although it didn't seem to cause any issues in my tests - maybe the LLMs I used)

PR #949, opened 2024-08-21
feat(mlx_lm): Nemotron

llllvvuu:
https://huggingface.co/nvidia/Minitron-4B-Base

Copy of llama.py with partial RoPE, ReLU squared instead of SwiGLU, and LayerNorm instead of BatchNorm. Also they add 1 to the LayerNorm weight for some reason.


llms/mlx_lm/models/nemotron.py
```diff
@@ -0,0 +1,239 @@
+# Copyright © 2024 Apple Inc.
+
+from dataclasses import dataclass
+from typing import Dict, Optional, Tuple, Union
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from mlx_lm.activations import get_activation
```
awni, line 9: Do you mind to hard code the activation and remove this. It makes the code a little easier to follow if the implementation is self-contained to a file. As far as I know the different models tend to fix the activation so having this flexibility from the config param isn't super useful.

llllvvuu: Done, thanks for the suggestion!

awni:
I wasn't able to test the model because it doesn't have safetensors. Is there a version which has safetensors?

llllvvuu:
I'm waiting for my safetensors PR there to be merged: https://huggingface.co/nvidia/Minitron-4B-Base/discussions/4

I think this works:
```sh
huggingface-cli download nvidia/Minitron-4B-Base --revision refs/pr/4
```

awni:
Another option is to simply upload new repos to the mlx-community with safetensors? Wdyt about doing that? Then we can land this since there will be usable models.

llllvvuu:
Uploaded: `mlx_lm.generate --model mlx-community/Minitron-4B-Base-4bit --prompt "The answer to life is " --ignore-chat-template`

awni:
Do you mind giving me access to push to this branch? I have a few nits that are more convenient to push directly.

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

llllvvuu:
Hmm, should be enabled already. Let me know if it still doesn't work, then I'll just add as collaborator

awni:
Indeed usually the default is to allow access .. but for your fork it doesn't work .. it seems like an authentication issue:

```
error: Authentication error: Authentication required: You must have push access to verify locks
error: failed to push some refs to 'github.com:llllvvuu/mlx-examples.git'
```

awni:
Huh..  the following seems to fix it 🤷‍♂️ 

```
git config --local lfs.https://github.com/.locksverify false
```

awni (approved):
Thanks for the addition!!

Issue #917, opened 2024-08-02
[Feature Request] Store KV cache of computed prompts to disk to avoid re-compute in follow-up runs


mark-lord:
It would be great if MLX_lm supported a --cache_prompt flag like in llama.cpp's integration (discussed in [this issue](https://github.com/ml-explore/mlx-examples/issues/new)). 

This would be a big benefit in reducing latency / start up time for repeated runs that include the same prompts - e.g. chatbot applications with a running chatlog, or when using long multishot examples that take up a lot of tokens (very useful for production environments).

I'm no expert so likely won't be able to assist with a PR on this :'( But looking at the discussion for the initial llama.cpp integration, I have a couple of observations:
- It seems that the cache should store the model's memory_k and memory_v data, along with other necessary state information incl. (but not limited to?) n_past, RNG state, logits, and embedding vectors
- I think the llama.cpp team may have implemented some form of lightweight compression (zstd?) for the cached data, as apparently the memory tensors are often compressible, especially when the full context is not used

The default behaviour of --cache-prompt might be to just save to the model folder like trained adapters; though it'd be useful to then have additional flags to maybe clear or delete the cache afterwards

Jckwind:
take a look at this open source library / repo: https://github.com/otriscon/llm-structured-output they have an implementation of a reusable KV cache for mlx. i've gotten it working - works surprisingly well!

awni:
I think we can close this! 🚀

PR #956, opened 2024-08-27
Add the ability to load the KV cache from a file.

angeloskath:
It adds a command called `mlx_lm.cache_prompt` that processes a prompt and saves the KV cache to a file. It also adds the option to load a KV cache from a file in `mlx_lm.generate`.

Combined with long context models it makes for some awesome applications. The following is an example using it to ask question about MLX to Phi-3.5 .

https://github.com/user-attachments/assets/ad386003-0236-4318-b25d-4fa9d7fa8146

### DIFF
--- a/llms/mlx_lm/generate.py
+++ b/llms/mlx_lm/generate.py
@@ -83,6 +83,12 @@ def setup_arg_parser():
         default=1024,
         help="Set the maximum key-value cache size",
     )
+    parser.add_argument(
+        "--kv-cache-file",
+        type=str,
+        default=None,
+        help="A file containing saved KV caches to avoid recomputing them",
+    )
     return parser
 
 
@@ -111,6 +117,24 @@ def colorprint_by_t0(s, t0):
     else:
         color = "red"
     colorprint(color, s)
+
+
+def load_kv_cache_from_file(kv_cache_file):
+    if kv_cache_file is None:
+        return None
+
+    kv_cache = mx.load(kv_cache_file)
+    cache_per_layer = {}
+    for k, x in kv_cache.items():
+        layer, kv_type = k.split("_")
+        if layer not in cache_per_layer:
+            cache_per_layer[layer] = {}
+        cache_per_layer[layer][kv_type] = x
+
+    cache_history = [None] * len(cache_per_layer)
+    for layer, c in cache_per_layer.items():
+        cache_history[int(layer)] = (c["keys"], c["values"])
+    return cache_history
 
 
 def main():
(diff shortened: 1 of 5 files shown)


fblissjr:
Is this essentially what Anthropic just released with prompt caching? It looks like it - which is amazing.

awni:
Yea it's the same idea!

awni:
Closes #917

angeloskath:
Added the metadata in the kv cache file. Now we can omit almost all arguments from generate when loading a kv cache.

For instance, below we don't need to pass the model or `max-kv-size` to the generate function, it just grabs them from the kv cache file.
```
cat /tmp/mlx_docs_prompt.txt | mlx_lm.cache_prompt --model mlx-community/Phi-3.5-mini-instruct-8bit --max-kv-size 48000 --prompt "-" --kv-cache-file /tmp/mlx_docs.safetensors
mlx_lm.generate --max-tokens 1024 --kv-cache-file /tmp/mlx_docs.safetensors --prompt "Create a uniform random array with shape 10 by 10 in MLX."
```

awni (approved):
Awesome! Let's 🚢  it.

angeloskath:
@altaic do you mean at the same time? The way it is set up the `generate` function simply expects a `cache_history` so multiple histories can be loaded in memory and selected on demand (not with the `generate.py` script though).

Loading a new one from a different file is of course trivial.

Also I am not promising anything but @barronalex is looking into quantizing the cache history which will make it a) faster and b) much more sensible to have multiple ones loaded at the same time.

ibehnam:
It looks like this is implemented mainly as a CLI goody, and it's not available in the Python API? 🤔

awni:
Please file an issue if that's something that would be useful for you and we can look into adding it to the Python API.

llms/mlx_lm/generate.py
```diff
@@ -113,6 +118,24 @@ def colorprint_by_t0(s, t0):
...
+    cache_per_layer = {}
+    for k, x in kv_cache.items():
+        layer, kv_type = k.split("_")
+        if layer not in cache_per_layer:
+            cache_per_layer[layer] = {}
+        cache_per_layer[layer][kv_type] = x
+
+    cache_history = [None] * len(cache_per_layer)
+    for layer, c in cache_per_layer.items():
+        cache_history[int(layer)] = (c["keys"], c["values"])
```
Jonathan-Dobson, line 135: Can you fix this type error?

No overloads for "__setitem__" match the provided argumentsPylance[reportCallIssue](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportCallIssue)
builtins.pyi(1028, 9): Overload 1 is the closest match
Argument of type "tuple[Unknown, Unknown]" cannot be assigned to parameter "value" of type "None" in function "__setitem__"
  "tuple[Unknown, Unknown]" is incompatible with "None"Pylance[reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType)


(variable) cache_history: list[None]

Jonathan-Dobson (commented):
Can you fix the metadata type error?

Object of type "None" is not subscriptablePylance[reportOptionalSubscript](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportOptionalSubscript)

(variable) metadata: Unknown | None

llms/mlx_lm/generate.py
```diff
@@ -113,6 +118,24 @@ def colorprint_by_t0(s, t0):
     colorprint(color, s)
 
 
+def load_kv_cache_from_file(kv_cache_file):
```
Jonathan-Dobson, line 121: Could you add compatible type definitions for this function.

PR #957, opened 2024-08-27
Support HuggingFace model tree

madroidmaq:
| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -649,7 +649,7 @@ def quantize_model(
     nn.quantize(model, q_group_size, q_bits)
     quantized_config["quantization"] = {"group_size": q_group_size, "bits": q_bits}
     # support hf model tree #957
-    quantized_config["quantization_config"] = {"group_size": q_group_size, "bits": q_bits}
+    quantized_config["quantization_config"] = quantized_config["quantization"]
     quantized_weights = dict(tree_flatten(model.parameters()))
 
     return quantized_weights, quantized_config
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/utils.py
```diff
@@ -646,7 +647,7 @@ def quantize_model(
     """
     quantized_config = copy.deepcopy(config)
     nn.quantize(model, q_group_size, q_bits)
-    quantized_config["quantization"] = {"group_size": q_group_size, "bits": q_bits}
+    quantized_config["quantization_config"] = {"group_size": q_group_size, "bits": q_bits}
```
awni, line 650: why change that?

madroidmaq: @awni 

I'm not quite sure about the specific use of `quantization`. When I don't adjust this, the quantized model is displayed as `Finetuned` in the Model Tree section. However, I think the quantized model after mlx conversion would be more appropriate, so I compared the models currently displayed as `Quantized` and found there are basically two situations:
1. Repositories with GGUF files;
2. Repositories with the `quantization_config` field in `config.json` (such as [fsaudm/Meta-Llama-3.1-8B-Instruct-INT8](https://huggingface.co/fsaudm/Meta-Llama-3.1-8B-Instruct-INT8/blob/main/config.json)), the `quantization` field does not have this effect.

Of course, if it is confirmed that the `quantization` field has a specific use, I think both can be retained.

awni: I think this change breaks a lot of stuff on our end which relies on that field being names `quantization`. So it would be good to revert it for now. It would be good to have HF parse the field correctly.. or we could consider supporting both `quantization` (for back-compat) + `quantization_config`. But for now could you please revert this?

awni (commented):
Thanks!

awni (approved):
Thanks @madroidmaq !

Issue #959, opened 2024-08-28
CLIP: preprocess


SauravMaheshkar:
Hi 👋🏽 I wanted to replicate the [Linear-probe evaluation script from the official CLIP repository](https://github.com/openai/CLIP?tab=readme-ov-file#linear-probe-evaluation). However, I can't figure out how to properly pre-process the images. I've attached my script and the corresponding error trace.

<details>

<summary>Evaluation script</summary>

```python
import numpy as np
import mlx.core as mx
from mlx.data.datasets import load_cifar10

from model import CLIPModel
from image_processor import CLIPImageProcessor


def get_cifar10(batch_size, root=None):
    mean = np.array([0.485, 0.456, 0.406]).reshape((1, 1, 3))
    std = np.array([0.229, 0.224, 0.225]).reshape((1, 1, 3))

    def normalize(x):
        x = x.astype(np.float32) / 255.0
        return (x - mean) / std

    tr = load_cifar10(root=root)
    tr_iter = tr.to_stream().key_transform("image", normalize).batch(batch_size)

    test = load_cifar10(root=root, train=False)
    test_iter = test.to_stream().key_transform("image", normalize).batch(batch_size)

    return tr_iter, test_iter


def get_features(model, iter):
    all_features = []
    all_labels = []
    image_proc = CLIPImageProcessor.from_pretrained("mlx_model")

    for _, batch in enumerate(iter):
        image, label = batch["image"], batch["label"]
        x = image_proc([mx.array(image)])
        y = mx.array(label)

        image_embeds = model.get_image_features(x)
        all_features.append(image_embeds)
        all_labels.append(y)

    return mx.concatenate(all_features), mx.concatenate(all_labels)


if __name__ == "__main__":
    model = CLIPModel.from_pretrained("mlx_model")

    train_iter, test_iter = get_cifar10(batch_size=256)
    train_features, train_labels = get_features(model, train_iter)
    test_features, test_labels = get_features(model, test_iter)
```

</details>

<details>
<summary>Error</summary>

```console
Traceback (most recent call last):
  File "/Users/saurav/dev/SauravMaheshkar/mlx-examples/clip/eval.py", line 47, in <module>
    train_features, train_labels = get_features(model, train_iter)
  File "/Users/saurav/dev/SauravMaheshkar/mlx-examples/clip/eval.py", line 33, in get_features
    x = image_proc([mx.array(image)])
[...]

PR #960, opened 2024-08-28
feat(clip): add linear probe evaluation script

SauravMaheshkar:
Adds a script to perform linear probe evaluation using the `mlx.data` module for data loading. Mostly a mirror of the [Linear-probe evaluation script from the official CLIP repository](https://github.com/openai/CLIP?tab=readme-ov-file#linear-probe-evaluation).

## References

* https://github.com/ml-explore/mlx-examples/issues/959

### DIFF
--- a/clip/eval.py
+++ b/clip/eval.py
@@ -1,0 +1,55 @@
+# Mirror of the Linear Probe Evaluation Script
+# from the official CLIP Repository.
+import mlx.core as mx
+import numpy as np
+from image_processor import CLIPImageProcessor
+from mlx.data.datasets import load_cifar10
+from model import CLIPModel
+from PIL import Image
+from sklearn.linear_model import LogisticRegression
+
+
+def get_cifar10(batch_size, root=None):
+    tr = load_cifar10(root=root)
+    tr_iter = tr.to_stream().batch(batch_size)
+
+    test = load_cifar10(root=root, train=False)
+    test_iter = test.to_stream().batch(batch_size)
+
+    return tr_iter, test_iter
+
+
+def get_features(model, image_proc, iter):
+    all_features = []
+    all_labels = []
+
+    for _, batch in enumerate(iter):
+        image, label = batch["image"], batch["label"]
+        x = image_proc([Image.fromarray(im) for im in image])
+        y = mx.array(label)
+
+        image_embeds = model.get_image_features(x)
+        all_features.append(image_embeds)
+        all_labels.append(y)
+
+    return mx.concatenate(all_features), mx.concatenate(all_labels)
+
+
+if __name__ == "__main__":
+    model = CLIPModel.from_pretrained("mlx_model")
+    image_proc = CLIPImageProcessor.from_pretrained("mlx_model")
+
+    train_iter, test_iter = get_cifar10(batch_size=256)
+    train_features, train_labels = get_features(model, image_proc, train_iter)
+    test_features, test_labels = get_features(model, image_proc, test_iter)
+
+    # Perform logistic regression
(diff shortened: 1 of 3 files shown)


clip/eval.py
```diff
@@ -0,0 +1,55 @@
...
+    test_iter = test.to_stream().batch(batch_size)
+
+    return tr_iter, test_iter
+
+
+def get_features(model, image_proc, iter):
+    all_features = []
+    all_labels = []
+
+    for _, batch in enumerate(iter):
```
angeloskath, line 26: Why enumerate? Possibly tqdm would be nice.

clip/eval.py
```diff
@@ -0,0 +1,55 @@
...
+from PIL import Image
+from sklearn.linear_model import LogisticRegression
+
+
+def get_cifar10(batch_size, root=None):
+    tr = load_cifar10(root=root)
+    tr_iter = tr.to_stream().batch(batch_size)
+
+    test = load_cifar10(root=root, train=False)
+    test_iter = test.to_stream().batch(batch_size)
```
angeloskath, line 17: Since it is a very small in memory dataset the stream is unnecessary here. I 'd keep it a buffer so that it is nicer and we have a `len` as well.

Namely,

```python
train_iter = load_cifar10(root=root).batch(batch_size)
```

clip/eval.py
```diff
@@ -0,0 +1,55 @@
...
+    all_labels = []
+
+    for _, batch in enumerate(iter):
+        image, label = batch["image"], batch["label"]
+        x = image_proc([Image.fromarray(im) for im in image])
+        y = mx.array(label)
+
+        image_embeds = model.get_image_features(x)
+        all_features.append(image_embeds)
+        all_labels.append(y)
```
angeloskath, line 33: You need an `mx.eval(image_embeds)` at some point otherwise you just create a huge graph for the GPU to compute at the same time which leads to memory problems.

angeloskath (changes requested):
Thanks for the addition. I would rename the script to `linear_probe.py` and add the eval. I am wondering if it would be nicer (since it is an example after all) to train a logistic regression model in MLX instead of using scikit-learn.

angeloskath:
Another thing to consider would be to have two commands in the `linear_probe.py` script. One that extracts features and saves them in a safe tensors file and another that does trains the logistic regression classifier given that file. The first part might be generally useful for extracting clip features for a dataset for instance.

clip/eval.py
SauravMaheshkar: Fixed in https://github.com/ml-explore/mlx-examples/pull/960/commits/dee703eacf2c8ae1a5c917426b2d38aa1db77ddc

clip/eval.py
SauravMaheshkar: Fixed in https://github.com/ml-explore/mlx-examples/pull/960/commits/dee703eacf2c8ae1a5c917426b2d38aa1db77ddc

SauravMaheshkar:
> Thanks for the addition. I would rename the script to `linear_probe.py` and add the eval. I am wondering if it would be nicer (since it is an example after all) to train a logistic regression model in MLX instead of using scikit-learn.

What I had in mind when submitting this PR was to showcase similar performance between the official implementation and the mlx port. Adding a mlx implementation of logistic regression seems like a nice idea but IMO it should reside in a different directory. Maybe another `misc/` or `core/` directory that contains implementations of various fundamental models.

SauravMaheshkar:
Genlte ping @angeloskath

angeloskath (approved):

PR #961, opened 2024-08-28
Fix `setattr` for the TokenizerWrapper

angeloskath:
Using the `--use-default-chat-template` seems to fail with the new `transformers` library and looking at how we handle `setattr` in the tokenizer wrapper seemed wrong anyway.

New behaviour: any `setattr` call is actually moved to the wrapped tokenizer except if it starts with an underscore.


awni (approved):
Thanks!!

PR #962, opened 2024-08-29
Add prompt piping 

chimezie:
Allows mlx_lm.generate to generate just the output (suppress verbosity) and to use STDIN as the prompt (to facilitate prompt piping) via the following options:

* --prompt-only (set generation verbosity to False and only print the prompt)
* Allow --prompt to be "-", which indicates the prompt should be pulled from STDIN (motivated by similar option in #956)

Below is an example

```bash
$ mlx_lm.generate --model mistralai/Mistral-Nemo-Instruct-2407 \
  --prompt "Write a request to create a quicksort function in Python, making sure the request is clear about the function being reursive" \
  --max-tokens 512 --temp 0.0 --prompt-only \
  --use-default-chat-template | mlx_lm.generate --model mlx-community/DeepSeek-Coder-V2-Lite-Instruct-4bit-mlx \
                                                --max-tokens 512 --temp 0.0 --prompt - --use-default-chat-template --colorize




Prompt: <｜begin▁of▁sentence｜>User: Here's a clear and concise request to create a recursive quicksort function in Python:

```plaintext
**Request:**

I need you to write a Python function named `quicksort` that implements the quicksort algorithm using recursion. The function should take a list of comparable items as input and return a new sorted list using the quicksort algorithm.

Here's a brief description of the quicksort algorithm for your reference:

1. Choose a 'pivot' element from the array and partition the other elements into two sub-arrays, according to whether they are less than or greater than the pivot.
2. Recursively apply the above steps to the sub-arrays.
3. Combine the sorted sub-arrays and the pivot to produce the final sorted array.

Please ensure that the function handles empty lists and lists with a single element correctly, as they are already sorted. Also, make sure to include proper input validation to handle non-list inputs or lists containing non-comparable items.

Here's a sample usage of the function:

```python
print(quicksort([3,6,8,10,1,2,1]))  # Output: [1, 1, 2, 3, 6, 8, 10]
```

Please provide the function definition along with a brief explanation of how it works.
```python
def quicksort(arr):
    # Base case: if the input is an empty list or a list with a single element, it's already sorted
    if len(arr) <= 1:
        return arr
    else:
        # Choose the pivot element (in this case, the first element of the list)
        pivot = arr[0]

        # Partition the other elements into two lists: less_than_pivot and greater_than_pivot
        less_than_pivot = [x for x in arr[1:] if x <= pivot]
        greater_than_pivot = [x for x in arr[1:] if x > pivot]

        # Recursively apply quicksort to the sub-lists and combine the results with the pivot
        return quicksort(less_than_pivot) + [pivot] + quicksort(greater_than_pivot)

# Test the function
print(quicksort([3,6,8,10,1,2,1]))  # Output: [1, 1, 2, 3, 6, 8, 10]
```

Assistant:
 Here's the Python function `quicksort` that implements the quicksort algorithm using recursion, as per your request:

```python
def quicksort(arr):
    # Base case: if the input is an empty list or a list with a single element, it's already sorted
    if len(arr) <= 1:
        return arr
    else:
        # Choose the pivot element (in this case, the first element of the list)
        pivot = arr[0]

        # Partition the other elements into two lists: less_than_pivot and greater_than_pivot
        less_than_pivot = [x for x in arr[1:] if x <= pivot]
        greater_than_pivot = [x for x in arr[1:] if x > pivot]

        # Recursively apply quicksort to the sub-lists and combine the results with the pivot
        return quicksort(less_than_pivot) + [pivot] + quicksort(greater_than_pivot)

# Test the function
print(quicksort([3,6,8,10,1,2,1]))  # Output: [1, 1, 2, 3, 6, 8, 10]
```

### Explanation:

1. **Base Case**: If the input list `arr` has a length of 0 or 1, it is already sorted, so the function returns the list as is.
2. **Pivot Selection**: The first element of the list is chosen as the pivot.
3. **Partitioning**: The rest of the elements are partitioned into two lists:
   - `less_than_pivot` contains elements that are less than or equal to the pivot.
   - `greater_than_pivot` contains elements that are greater than the pivot.
4. **Recursive Sorting**: The `quicksort` function is called recursively on the `less_than_pivot` and `greater_than_pivot` lists.
5. **Combining Results**: The sorted `less_than_pivot` list, the pivot, and the sorted `greater_than_pivot` list are combined to produce the final sorted list.

This implementation ensures that the function handles empty lists and lists with a single element correctly, as well as properly handles non-list inputs or lists containing non
==========
Prompt: 554 tokens, 507.215 tokens-per-sec
Generation: 512 tokens, 69.750 tokens-per-sec
Peak memory: 8.519 GB
```

### DIFF
--- a/llms/mlx_lm/generate.py
+++ b/llms/mlx_lm/generate.py
@@ -1,17 +1,18 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import argparse
+import json
 
 import mlx.core as mx
 
 from .utils import generate, load
 
-DEFAULT_MODEL_PATH = "mlx_model"
 DEFAULT_PROMPT = "hello"
 DEFAULT_MAX_TOKENS = 100
 DEFAULT_TEMP = 0.6
 DEFAULT_TOP_P = 1.0
 DEFAULT_SEED = 0
+DEFAULT_MAX_KV_SIZE = 1024
 
 
 def setup_arg_parser():
@@ -20,7 +21,6 @@ def setup_arg_parser():
     parser.add_argument(
         "--model",
         type=str,
-        default="mlx_model",
         help="The path to the local model directory or Hugging Face repo.",
     )
     parser.add_argument(
@@ -80,8 +80,13 @@ def setup_arg_parser():
     parser.add_argument(
         "--max-kv-size",
         type=int,
-        default=1024,
         help="Set the maximum key-value cache size",
+    )
+    parser.add_argument(
+        "--kv-cache-file",
+        type=str,
+        default=None,
+        help="A file containing saved KV caches to avoid recomputing them",
     )
     return parser
 
@@ -113,6 +118,24 @@ def colorprint_by_t0(s, t0):
     colorprint(color, s)
 
 
+def load_kv_cache_from_file(kv_cache_file):
+    if kv_cache_file is None:
+        return None, None
+
+    kv_cache, metadata = mx.load(kv_cache_file, return_metadata=True)
+    cache_per_layer = {}
+    for k, x in kv_cache.items():
+        layer, kv_type = k.split("_")
+        if layer not in cache_per_layer:
+            cache_per_layer[layer] = {}
+        cache_per_layer[layer][kv_type] = x
+
+    cache_history = [None] * len(cache_per_layer)
+    for layer, c in cache_per_layer.items():
+        cache_history[int(layer)] = (c["keys"], c["values"])
+    return cache_history, metadata
+
+
 def main():
     parser = setup_arg_parser()
     args = parser.parse_args()
@@ -122,13 +145,25 @@ def main():
     if args.cache_limit_gb is not None:
         mx.metal.set_cache_limit(args.cache_limit_gb * 1024 * 1024 * 1024)
 
+    # Load the kv cache and metadata if a kv cache file is provided
+    cache_history, metadata = load_kv_cache_from_file(args.kv_cache_file)
+
     # Building tokenizer_config
-    tokenizer_config = {"trust_remote_code": True if args.trust_remote_code else None}
+    tokenizer_config = (
+        {} if cache_history is None else json.loads(metadata["tokenizer_config"])
+    )
+    if args.trust_remote_code:
+        tokenizer_config["trust_remote_code"] = True
     if args.eos_token is not None:
         tokenizer_config["eos_token"] = args.eos_token
 
+    # If no model path is provided then use the one in the kv cache history
+    model_path = args.model
+    if cache_history is not None and model_path is None:
+        model_path = metadata["model"]
+
     model, tokenizer = load(
-        args.model,
+        model_path,
         adapter_path=args.adapter_path,
         tokenizer_config=tokenizer_config,
     )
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: llms/mlx_lm/generate.py [hunk 1]
```diff
@@ -66,6 +69,13 @@ def setup_arg_parser():
         action="store_true",
         help="Use the default chat template",
     )
+    parser.add_argument(
+        "--prompt-only",
```
awni, line 73: The name of this argument is a bit confusing to me. I think it might be preferrable to make it a `--verbose` flag which defaults to `True`. If `False` then only the generation is printed with no streaming.

chimezie: @awni  Ok.  It was the fact that we were calling generate with verbose hardcoded to True that made me consider another name.  If we want to keep the defaults as True, what about having --verbose/--no-verbose options . I.e, (Python 3.9+):

parser.add_argument('--verbose', action=argparse.BooleanOptionalAction, default=True)

awni: That's really nice, I didn't know about it. But we'd like to support Python 3.8 so it won't work.

You could do something like this:

```
    def str2bool(string):
        return string != "False"

    parser.add_argument(
        "--verbose",
        type=str2bool,
        default=True,
        help="Log verbose output when 'True' or only print the response when 'False'",
    )

Comment (chimezie) [reply llms/mlx_lm/generate.py]: @awni I have made this change

File: llms/mlx_lm/generate.py [hunk 2]
```diff
@@ -14,6 +15,10 @@
 DEFAULT_SEED = 0
 
 
+def str2bool(string):
+    return string.lower() not in ["False", "F"]
```
awni, line 19: Looks like that will never return `True` since you lower case the string before the check:
```suggestion
    return string not in ["False", "F"]
```

chimezie: @awni Thanks for pointing out that typo.  I wanted the option to work for 'T', 'True', 'F', and 'False' without regard to case, so I changed it to:

```python
def str2bool(string):
    return string.lower() not in ["false", "f"]
```

llms/mlx_lm/generate.py
```diff
@@ -195,6 +213,8 @@ def main():
     else:
         prompt = args.prompt
 
+    if args.colorize and args.prompt_only:
```
awni, line 216: @chimezie this argument is gone now. Please fix.

llms/mlx_lm/generate.py
```diff
@@ -203,18 +223,20 @@ def main():
...
         args.max_tokens,
-        verbose=True,
+        verbose=args.verbose,
         formatter=formatter,
         temp=args.temp,
         top_p=args.top_p,
         max_kv_size=max_kv_size,
         cache_history=cache_history,
     )
+    if args.prompt_only:
```
awni, line 238: Same here.

llms/mlx_lm/generate.py
```diff
@@ -195,6 +213,8 @@ def main():
     else:
         prompt = args.prompt
 
+    if args.colorize and args.prompt_only:
+        raise ValueError("Cannot use --colorize with --prompt-only")
```
awni, line 217:
```suggestion
        raise ValueError("Cannot use --colorize with --verbose=False")
```

chimezie: Fixed

llms/mlx_lm/generate.py
chimezie: Fixed

awni (approved):

PR #971, opened 2024-09-02
Fix attention layers map  for SD-2-1-Base

pranav4501:
The  [safetensors](https://huggingface.co/stabilityai/stable-diffusion-2-1-base/tree/main/vae?show_file_info=vae%2Fdiffusion_pytorch_model.safetensors) are not mapped correctly for attention layers of SD-2-1-base.

### DIFF
--- a/stable_diffusion/stable_diffusion/model_io.py
+++ b/stable_diffusion/stable_diffusion/model_io.py
@@ -140,6 +140,17 @@ def map_vae_weights(key, value):
     if "to_v" in key:
         key = key.replace("to_v", "value_proj")
 
+
+    # Map attention layers in SD-2-1-base:VAE
+    if "key" in key:
+        key = key.replace("key", "to_k")
+    if "proj_attn" in key:
+        key = key.replace("proj_attn", "out_proj")
+    if "query" in key:
+        key = key.replace("query", "query_proj")
+    if "value" in key:
+        key = key.replace("value", "value_proj")
+
     # Map the mid block
     if "mid_block.resnets.0" in key:
         key = key.replace("mid_block.resnets.0", "mid_blocks.0")


stable_diffusion/stable_diffusion/model_io.py
```diff
@@ -140,6 +140,17 @@ def map_vae_weights(key, value):
...
+
+    # Map attention layers in SD-2-1-base:VAE
+    if "key" in key:
+        key = key.replace("key", "to_k")
+    if "proj_attn" in key:
+        key = key.replace("proj_attn", "out_proj")
+    if "query" in key:
+        key = key.replace("query", "query_proj")
+    if "value" in key:
+        key = key.replace("value", "value_proj")
```
angeloskath, line 152: These don't look correct given that they will result in `value_proj_proj` and `query_proj_proj` and `to_k`. Am I missing sth?

angeloskath:
Hi @pranav4501 could you give an example of what wasn't working and what this fixes? I will take a closer look but I would appreciate sth like `image2image.py` is not working for this model or some weights and the corresponding model file sth that clearly shows the problem.

Thanks and sorry if I am putting extra work on you.

pranav4501:
Hi @angeloskath , Sorry for the late response. You will have to make a couple changes to [model_io.py](https://github.com/ml-explore/mlx-examples/blob/main/stable_diffusion/stable_diffusion/model_io.py) of stable diffusion
1. [Update this line to L-175](https://github.com/ml-explore/mlx-examples/blob/main/stable_diffusion/stable_diffusion/model_io.py#L175)
```py
model.load_weights(list(weights),strict=True)
```
2. [Comment out this line L-262](https://github.com/ml-explore/mlx-examples/blob/main/stable_diffusion/stable_diffusion/model_io.py#L262), Positions ids in CLIP is not implemented, I will try to look into it
3. Run The image generation
```bash
python3 txt2image.py "A photo of an astronaut riding a horse on Mars." --n_images 1 --steps 1 --model sd
```
4. You should encounter something like
```bash
ValueError: Received parameters not in model: encoder.mid_blocks.1.proj_attn.weight encoder.mid_blocks.1.value.bias decoder.mid_blocks.1.proj_attn.bias decoder.mid_blocks.1.value.bias
```

PR #981, opened 2024-09-07
Fix bug in upload + docs nit

awni:
- I broke the uploading with some bad text formatting so this resolves that.
- Also made some clarifications in the LoRA docs about the dataset format.


angeloskath (approved):
Thanks for fixing it!

PR #984, opened 2024-09-12
Add /v1/models endpoint to mlx_lm.server

jamesm131:
This pull request introduces a new /v1/models endpoint to mlx_lm.server. The purpose of this addition is to improve compatibility with clients such as Open WebUI, which expect to retrieve a list of available models before use.

Changes:
1. Implemented GET request handling for the /v1/models route
2. Added a method to return information about the loaded model
3. Added tests for the /v1/models route

I've tested Open WebUI compatibility and this now works.

### DIFF
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -11,6 +11,7 @@ from pathlib import Path
 from typing import Dict, List, Literal, NamedTuple, Optional, Sequence, Union
 
 import mlx.core as mx
+from huggingface_hub import scan_cache_dir
 
 from .utils import generate_step, load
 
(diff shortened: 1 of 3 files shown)


llms/mlx_lm/server.py
```diff
@@ -618,6 +618,38 @@ def handle_text_completions(self) -> mx.array:
...
+        """
+        self._set_completion_headers(200)
+        self.end_headers()
+        # Create a list of available models
+        models = [
+            {
+                "id": self.model_provider.cli_args.model,
+                "object": "model",
+                "created": self.created,
+            }
```
awni, line 644: This looks like it just retrieves the active model. Would it be good to list all the already downloaded models or something like that?

awni: > It could check the huggingface/hub directory for mlx-community models.
 
We have some code in MLX LM already for [querying the HF cache](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/manage.py). We could repurpose some of that for this. Ideally it would list al models (not just MLX Community) in the HF cache that have a `model_type`  which MLX LM supports. That wouldn't include models not in the cache.. but I think that's ok.

> The bigger change would be on how the server should react when a v1/chat/completions call is made to a model that hasn't been pre-loaded.

The server already handles this.

Are you interested in adding that? In this PR or a follow on?

awni:
I think we can do this in stages. So starting with this PR and then possibly adding caching / support for model_type if as a second step.

I will review this as a starting point.

Regarding the next step, to me the main question is if it should support not-yet-downloaded models. If the answer is yes then maybe we want to use the HF Hub API to query for model_type. If the answer is no, then we should find a way to just get model types for already downloaded models. As for caching that's an optimization we may use in either case depending on the speed.

awni (approved):
Thanks for the addition!!

PR #991, opened 2024-09-18
Encodec

awni:
Should we add this to MLX examples?

- Commonly used vocoder in TTS models
- It's a pretty nice example using conv transpose, `mx.fast.metal_kernel` / LSTMs

Also the benchmarks are looking pretty good:

On an M2 Ultra with the latest MLX:

Bench | Milliseconds
----- | -----
MLX | 171.721 
Transformers PyTorch MPS | 405.021


fblissjr:
Was late to see this and since no replies and I see you putting the models up on HF - definitely. :)

Focusing on other modalities, along with parallel inference, would be great examples.

Given the focus on inference-optimization now, it'd be awesome (if possible on the hardware?) to do dynamic generation like this: https://github.com/turboderp/exllamav2

angeloskath (approved):
Looks awesome!

PR #1013, opened 2024-10-04
Make T5 work with official models without conversions

zcbenz:
Restructure the T5 model code so it can load the official models without weights conversions.

The new code outputs exactly the same with the old code, however I got some questions when trying to improve the code:

1. I tried to use `fast.scaled_dot_product_attention` in the attention code but the output is wrong, can you check what went wrong? The fast version is commented in the code.
2. The original models have 2 different position biases for self attention and cross attention in the decoder, but this code uses the position bias of self attention for both kinds of attentions, and the output is correct, why?


t5/t5.py
```diff
@@ -108,30 +111,67 @@ def __call__(
...
 
-        # Dimensions are [batch x num heads x sequence x hidden dim]
-        scores = queries @ keys
-        if mask is not None:
-            scores = scores + mask.astype(scores.dtype)
+        # scale = math.sqrt(1 / queries.shape[-1])
+        # output = mx.fast.scaled_dot_product_attention(
+        #     queries, keys, values, scale=scale, mask=mask
+        # )
+        # output = output.transpose(0, 2, 1, 3).reshape(B, L, -1)
```
angeloskath, line 134: Why not use the above and instead write the computation explicitly?

t5/t5.py
```diff
@@ -217,78 +267,70 @@ def __call__(
...
-            cache = [None] * len(self.layers)
+            cache = [None] * len(self.block)
 
         T = offset + x.shape[1]
-        pos_bias = self.relative_attention_bias(T, T, offset=offset)
+        pos_bias = self.relative_attention_bias(
+            self.block[0].layer[0].SelfAttention.relative_attention_bias,
+            T,
+            T,
+            offset)
```
angeloskath, line 301: This is a bit convoluted. I would much rather it remained as it were.

angeloskath (commented):
Generally I like the point of avoiding the conversion step. It is what we are doing in `mlx-lm` and the diffusion examples. However, I think there is no need to make everything exactly compatible.

A good example of an overkill is the positional bias where the weights "need" to be in a specific layer and then the whole code changes such that it accommodates this requirement. It would be much simpler imho to make a small sanitize function that would move the positional bias wherever we want it.

I suggest that we make this using the `sanitize(...)` pattern that we use in mlx-lm as well. Wdyt?

t5/t5.py
zcbenz: When using `fast.scaled_dot_product_attention` the output of the model becomes wrong, I left the code there so someone familiar with the fast version can check what went wrong.

zcbenz:
> A good example of an overkill is the positional bias where the weights "need" to be in a specific layer and then the whole code changes such that it accommodates this requirement. It would be much simpler imho to make a small sanitize function that would move the positional bias wherever we want it.
> 
> I suggest that we make this using the `sanitize(...)` pattern that we use in mlx-lm as well. Wdyt?

I agree `sanitize(...)` pattern would make the position bias code better much. But before making the change, I do have a question on the position bias though.

The original models have 2 different position biases for self attention and cross attention in the decoder, but this code uses the position bias of self attention for both kinds of attentions, and the output is correct, do you know why?

zcbenz:
Closed in favor of #1020.

PR #1014, opened 2024-10-04
Fix rotating KV cache for chat use case

### LINKED ISSUE #1000: RotatingKVCache: Problem when reusing cache between multiple generations
The design of `RotatingKVCache` assumes that, text generation starts with a long prompt first, and then continues token by token, and then end.

But in chat apps the program usually work like this:

```
cache = create_cache();
while (true) {
  new_messages = wait_input_from_user();
  generate(new_messages, model, cache);
}
```

which does multiple generations with the same cache, and when using `RotatingKVCache` in such apps it would return wrong results after 1st `generate` call.

awni:
The rotating KV cache didn't work if one alternates cache filling with generation. This fixes that + some tests.

Closes #1000

### DIFF
--- a/llms/mlx_lm/_version.py
+++ b/llms/mlx_lm/_version.py
@@ -1,3 +1,3 @@
 # Copyright © 2023-2024 Apple Inc.
 
-__version__ = "0.18.2"
+__version__ = "0.19.1"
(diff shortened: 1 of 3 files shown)


awni:
Closing in favor of #1015 which I think has a better fix.

PR #1015, opened 2024-10-05
More cache improvements

### LINKED ISSUE #1000: RotatingKVCache: Problem when reusing cache between multiple generations

awni:
Sorry for the large diff. There's a lot of boiler plate / moving stuff around which accounts for most of it.

The main bits are:
- Fix RotatingKVCache for alternating chat, response use case
- Enable prompt caching for all types (not just `KVCache`)
- Unify APIs and cache types in a single file for ease of use / consistency.
- Chat mode allows prompt caching for efficiency. [Example here](https://github.com/ml-explore/mlx-examples/pull/1015/files#diff-4c37898018d7155df3ed91b43daf9976328abfcb123ce81fc96516f53e505305).
- Add a bunch of tests.

Closes #1000

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@ __pycache__/
 # C extensions
 *.so
 
+# Vim
+*.swp
+
 # Distribution / packaging
 .Python
 build/
--- a/llms/mlx_lm/models/stablelm.py
+++ b/llms/mlx_lm/models/stablelm.py
@@ -2,7 +2,6 @@
 
 import math
 from dataclasses import dataclass
-from typing import Tuple
 
 import mlx.core as mx
 import mlx.nn as nn
@@ -198,8 +197,8 @@ class Model(nn.Module):
         self,
         x: mx.array,
         mask: mx.array = None,
-        cache: mx.array = None,
-    ) -> Tuple[mx.array, mx.array]:
+        cache=None,
+    ) -> mx.array:
         mask = create_attention_mask(x, cache)
         y = self.model(x, mask, cache)
         return self.lm_head(y)
@@ -207,11 +206,3 @@ class Model(nn.Module):
     @property
     def layers(self):
         return self.model.layers
-
-    @property
-    def head_dim(self):
-        return self.args.hidden_size // self.args.num_attention_heads
-
-    @property
-    def n_kv_heads(self):
-        return self.args.num_key_value_heads
--- a/llms/README.md
+++ b/llms/README.md
@@ -155,14 +155,14 @@ different queries. To cache a prompt use `mlx_lm.cache_prompt`. For example:
 cat prompt.txt | mlx_lm.cache_prompt \
   --model mistralai/Mistral-7B-Instruct-v0.3 \
   --prompt - \
-  --kv-cache-file mistral_prompt.safetensors
+  --prompt-cache-file mistral_prompt.safetensors
 ``` 
 
 Then use the cached prompt with `mlx_lm.generate`:
 
 ```
 mlx_lm.generate \
-    --kv-cache-file mistral_prompt.safetensors \
+    --prompt-cache-file mistral_prompt.safetensors \
     --prompt "\nSummarize the above text."
 ```
 
--- a/llms/README.md
+++ b/llms/README.md
@@ -20,6 +20,31 @@ The `mlx-lm` package also has:
 - [Merging models](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/MERGE.md)
 - [HTTP model serving](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/SERVER.md)
 
+### Quick Start
+
+To generate text with an LLM use:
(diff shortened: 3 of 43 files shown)


awni:
I also added a chat command to MLX which is a good use case for the prompt cach re-use. The example is kind of fun to play with:

```
mlx_lm.chat
```

Then you can just chat with the model and it preserves the history and doesn't do any prompt recomputations..

```text
[INFO] Starting chat sessiong with mlx-community/Llama-3.2-3B-Instruct-4bit. To exit, enter 'q'.
>> Hi, my name is Awni!
Hi Awni! It's nice to meet you. Is there something I can help you with or would you like to chat?
>> What's the tallest mountain in the world?
The tallest mountain in the world is Mount Everest, which is located in the Himalayas on the border between Nepal and Tibet, China. It stands at an elevation of 8,848 meters (29,029 feet) above sea level.
>> Do you remember my name?
Yes, your name is Awni.
>> Nice talking with you!
It's great to chat with you too, Awni! Is there anything else you'd like to talk about or ask about?
>> 
```

llms/mlx_lm/models/stablelm.py
```diff
@@ -198,20 +197,12 @@ def __call__(
         self,
         x: mx.array,
         mask: mx.array = None,
-        cache: mx.array = None,
-    ) -> Tuple[mx.array, mx.array]:
+        cache=None,
```
angeloskath, line 200: I don't personally care but in most cases this is written as `cache: Optional[Any] = None`

angeloskath (approved):
This is a fantastic refactoring! Especially love the tests.

I am wondering what is ht point of the extra state in the KV cace? Is anybody using it now? Is there any reason it is set to the empty string instead of `None`?

I may have missed a discussion but I couldn't find its use in the code as well.

awni:
> I am wondering what is ht point of the extra state in the KV cace? Is anybody using it now? Is there any reason it is set to the empty string instead of None?

It's my least favorite thing in this diff, but I didn't think of a cleaner solution yet (if you have ideas I'm all 👂 )

It is used only for the `RotatingKVCache` so that we can save the `cache.offset` and `cache._idx`. Otherwise we don't get the right/same behavior when serializing and deserializing that cache.

The reason I made it a string and not `None` is because that simplified saving it in safetensors metadata.  So downstream code just does something like `dict(tree_flatten([c for c in cache.state[1]]))`.

.gitignore
```diff
@@ -6,6 +6,9 @@ __pycache__/
 # C extensions
 *.so
 
+# Vim
+*.swp
```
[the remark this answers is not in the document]
awni: No no, I just got tired of seeing them when I do `git status` etc.

llms/mlx_lm/models/stablelm.py
awni: Good catch, will fix.

angeloskath:
Huh, not sure how I managed to miss it in the `RotatingKVCache`...

I think this should be changed because the `state` is what we evaluate from the caches so this is why I was confused with the string. The most minor change that would be imho significantly better would be to split it into `state` and `serialization_state`. It is a bit verbose but at least it separates the two types of information cleanly.

Nothing changes much. Line 47 would do sth like 

```python
cache_data = [c.state for c in cache]
cache_info = [c.serialization_state for c in cache] # do we also want type(self).__name__ here?
```

and lines 75 would change to

```python
for c, state, serialization_state in zip(cache, arrays, info):
    c.state = state
    c.serialization_state = serialization_state
```

The rest remains the same... Wdyt?

awni:
Yea I thought about a separate property.. and/or overriding `__getstate__` and `__setstate__`. The main downside I didn't like is that all the caches needed to implement it.. but maybe the right call is to check if the attribute exists to avoid that. I think you're right it could be cleaner even if a little more verbose.

angeloskath:
I added a small base class that implements the empty meta state and makes the load/save code a tad bit cleaner? Should I push it on top or we are avoiding base classes for some reason?

Also just played with the chat command it is an absolute joy to use :-)

awni:
> I added a small base class that implements the empty meta state and makes the load/save code a tad bit cleaner? Should I push it on top or we are avoiding base classes for some reason?

No reason at this point, please send the diff!

awni:
Ok, I tested prompt caching with a few different models / cache types and it seems to work well. I'm going to merge this.

As a follow up we should consider:
- Making a way to serialize the chat context
- Adding a chat endpoint to `mlx_lm.server` with prompt caching

mark-lord:
So excited this got merged 😄

Issue #894, opened 2024-07-18
Classification Example


vjagannath786:
Looking for some examples where LLM models like llama or Gemma is used for classification. And, also if it can return the logits.

awni:
I don't know about examples where they are used for classification but if any come across our path we will add them here.

You can get the logprobs by using `generate_step` directly:

```python
def generate_with_logits(
    model: nn.Module,
    tokenizer: Union[PreTrainedTokenizer, TokenizerWrapper],
    prompt: str,
    max_tokens: int = 100,
    **kwargs,
):
    if not isinstance(tokenizer, TokenizerWrapper):
        tokenizer = TokenizerWrapper(tokenizer)

    prompt_tokens = mx.array(tokenizer.encode(prompt))
    detokenizer = tokenizer.detokenizer

    detokenizer.reset()
    all_logprobs = []

    for (token, logprobs), n in zip(
        generate_step(prompt_tokens, model, **kwargs),
        range(max_tokens),
    ):
        if token == tokenizer.eos_token_id:
            break
        detokenizer.add_token(token)
        all_logprobs.append(logprobs)

    detokenizer.finalize()

    return detokenizer.text, all_logprobs
```

farris:
Could this be added as default to the library?

PR #1016, opened 2024-10-06
Return logprobs from generate api

farris:
Continuation of https://github.com/ml-explore/mlx-examples/issues/894#issuecomment-2236464052

Could be incorporated into new function altogether but proposing to build into existing `generate` as the logic is minimal...

Thanks 😄

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -314,10 +314,11 @@ def generate(
     tokenizer: Union[PreTrainedTokenizer, TokenizerWrapper],
     prompt: str,
     max_tokens: int = 100,
+    logprobs_seq: bool = False,
     verbose: bool = False,
     formatter: Optional[Callable] = None,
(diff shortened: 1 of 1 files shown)


fblissjr:
I actually did this myself when it first came up in the PR to introduce logprobs. Comment here: https://github.com/ml-explore/mlx-examples/pull/806#issuecomment-2211931951

farris:
> I actually did this myself when it first came up in the PR to introduce logprobs. Comment here: [#806 (comment)](https://github.com/ml-explore/mlx-examples/pull/806#issuecomment-2211931951)
> 
> Yours makes more sense from a library standpoint though. I would find it way more useful in utils, personally.
> 
> I use logprobs and the associated tokens/vocab quite often.

Nice, maybe the functionality should be shared between the server implementation which you did and the regular `generate` interface 

@awni , thoughts?

PR #1020, opened 2024-10-08
MusicGen

barronalex:
Adds a [MusicGen](https://arxiv.org/abs/2306.05284) example.

Uses the existing T5 and Encodec examples for the text encoding / audio decoding.

I ended up removing the conversion step for T5 as in #1013 so we'll want to sequence that carefully if we merge this.

On M1 Max:

| Bench | Time (ms) |
| ------------- | ------------- |
| MLX  | 81.0 |
| Transformers PyTorch MPS  | 190.5 |


awni:
Whoa cool!! Should we try an land 1013 ahead of this or is it simpler to close that and just land it here?

angeloskath:
Music, images, text ... 🥳🥳🥳

I think this is more inline with the rest of the examples that have the `sanitize` pattern and I think @zcbenz agreed on the other PR.

I haven't done a thorough read but the only thing that may require some though is using the paths to import t5 and encodec. It is probably better than simply copying them into every project but I am not 100% on the paths. Otoh I don't know what other options we have:
- symlinks?
- top-level mlx-examples install?

zcbenz:
Let's close #1013 in favor of this, it is easier rebasing some changes on this PR.

barronalex:
Thanks @zcbenz, sorry for the concurrent PR!

Agreed on the ugliness of the paths. Symlinks sound pretty reasonable to me.

awni:
Symlinks seem like a good call for now.

Though for the sake of discussion, I wonder  if we should consider packaging them up in some way. It's kind of common to use say a T5 in various places. It would be nice if I could do:

`pip install mlx_t5` and get the API to use it.

Or maybe we should package up a bunch of models here as like `mlx_models` and then you can do `from mlx_models import T5` or something like that.. not sure just brainstorming.

It's also not that bad to copy the code (which is one reason to keep it pretty minimal in the first place).

encodec/README.md
```diff
@@ -33,13 +33,14 @@ An example using the model:
 
 ```python
 import mlx.core as mx
-from utils import load, load_audio, save_audio
+from encodec import EncodecModel
+from utils import load_audio, save_audio
 
 # Load the 48 KHz model and preprocessor.
-model, processor = load("mlx-community/encodec-48khz-float32")
+model, processor = EncodecModel.from_pretrained("mlx-community/encodec-48khz-float32")
```
awni, line 40: That's nice. Very 🤗

encodec/README.md
```diff
@@ -33,13 +33,14 @@ An example using the model:
...
 
 # Load an audio file
-audio = load_audio("path/to/aduio", model.sampling_rate, model.channels)
+audio = load_audio("path/to/audio", model.sampling_rate, model.channels)
```
awni, line 43: ❤️

musicgen/generate.py
```diff
@@ -0,0 +1,23 @@
...
+
+from utils import save_audio
+
+from musicgen import MusicGen
+
+
+def main(text: str, output_path: str, model_name: str, max_steps: int):
+    model = MusicGen.from_pretrained(model_name)
+    audio = model.generate(text, max_steps=max_steps)
+    save_audio(output_path, audio, model.sampling_rate)
```
awni, line 13: This might be the simplest generate script I've ever seen. Awesome!

awni (approved):
This is very nice! Good improvements all around.

Two requests:

1. Could you add an example-level README for this example? Can be pretty basic, just like a line about what the model is with some attribution and a line or two about how to run it.
2. Could you add a `requirements.txt` for this as well?

Can be pretty simple, just like

musicgen/musicgen.py
```diff
@@ -0,0 +1,359 @@
...
+@partial(mx.compile, inputs=mx.random.state, outputs=mx.random.state)
+def top_k_sampling(
+    logits: mx.array, top_k: float, temperature: float, axis: int = -1
+) -> mx.array:
+    """
+    Apply top-k sampling to logits.
+
+    Args:
+        logits: The logits from the model's output.
+        top_k: The cumulative probability threshold for top-p filtering.
```
awni, line 157:
```suggestion
        top_k: The cumulative probability threshold for top-k filtering.
```

musicgen/musicgen.py
```diff
@@ -0,0 +1,359 @@
...
+        temperature: Temperature parameter for softmax distribution reshaping.
+        axis: Axis along which to sample
+    Returns:
+        token selected based on the top-p criterion.
```
awni, line 161:
```suggestion
        token selected based on the top-k criterion.
```

musicgen/musicgen.py
awni: Also I think that description of the `k` parameter is inaccurate.

musicgen/musicgen.py
```diff
@@ -0,0 +1,359 @@
...
+            )
+
+        with open(path / "config.json", "r") as f:
+            config = SimpleNamespace(**json.load(f))
+            config.text_encoder = SimpleNamespace(**config.text_encoder)
+            config.audio_encoder = SimpleNamespace(**config.audio_encoder)
+            config.decoder = SimpleNamespace(**config.decoder)
+
+        weights = torch.load(path / "state_dict.bin", weights_only=True)["best_state"]
+        weights = {k: mx.array(v.numpy()) for k, v in weights.items()}
```
awni, line 354: I think it's better to go directly to `mx.array` here and not step through `numpy`. Avoids a copy and also won't crash for `bfloat16` (not sure if that's a thing)

musicgen/musicgen.py
```diff
@@ -0,0 +1,359 @@
...
+
+        if cache is None:
+            cache = [None] * len(self.layers)
+
+        x = sum([self.emb[k](audio_tokens[..., k]) for k in range(self.num_codebooks)])
+
+        offset = cache[0].offset if cache[0] is not None else 0
+        offset = mx.full((audio_tokens.shape[0], 1, 1), offset)
+        pos_emb = create_sin_embedding(offset, self.hidden_size)
+        x += pos_emb
```
awni, line 241: II'm not sure if the models are ever in lower precision. If so a cast here will be important: `x += pos_emb.astype(x.dtype)`

musicgen/musicgen.py
```diff
@@ -0,0 +1,359 @@
...
+        cache: list[KVCache] = None,
+    ):
+
+        if cache is None:
+            cache = [None] * len(self.layers)
+
+        x = sum([self.emb[k](audio_tokens[..., k]) for k in range(self.num_codebooks)])
+
+        offset = cache[0].offset if cache[0] is not None else 0
+        offset = mx.full((audio_tokens.shape[0], 1, 1), offset)
```
awni, line 239: I don't think you need to expand to the batch size here `mx.array(offset)` probably works ... though won't matter from a perf standpoint.

musicgen/musicgen.py
barronalex: Really nice that that works!

barronalex:
Thanks for the comments! I'd forgotten to `git add` the README and requirements they're in there now :)

musicgen/README.md
```diff
@@ -0,0 +1,47 @@
...
+```
+# on macOS using Homebrew (https://brew.sh/)
+brew install ffmpeg
+```
+
+Install SciPy:
+
+```
+pip install scipy
+```
```
awni, line 28: Nit: scipy is in the requirements.txt so you can probably remove the stuff here about installing scipy.

awni (approved):
Looks great! Let's 🚢

t5/convert.py
```diff
@@ -1,75 +0,0 @@
-import numpy as np
```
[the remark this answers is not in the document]
barronalex: I think the 23 downloads are from your nice [open_sora](https://github.com/awni/mlx-examples/tree/open_sora/open_sora) example. That should still work OK since it has it's own T5 loading so we're probably good for the moment.

PR #1024, opened 2024-10-09
Tokenizer updates + tests

awni:
- Fixes issue with the naive streaming decoder where it would drop tokens when `tokenizer.clean_up_tokenization_spaces` is set.
- Updates BPE matching condition to be more permissive and also mostly fixes it for the case `tokenizer.clean_up_tokenization_spaces`
- Adds some tests


angeloskath (approved):
👍🙏

PR #1026, opened 2024-10-09
Prompt caching in `mlx_lm.server`

awni:
Added a basic prompt cache in `mlx_lm.server` for chat mode. But it does support chatting with cache reuse.

### DIFF
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -3,16 +3,30 @@
 import argparse
 import json
 import logging
+import platform
 import time
 import uuid
 import warnings
+from dataclasses import dataclass, field
 from http.server import BaseHTTPRequestHandler, HTTPServer
 from pathlib import Path
-from typing import Dict, List, Literal, NamedTuple, Optional, Sequence, Union
+from typing import (
+    Any,
+    Dict,
+    List,
+    Literal,
+    NamedTuple,
+    Optional,
+    Sequence,
+    Tuple,
+    Union,
+)
 
 import mlx.core as mx
 from huggingface_hub import scan_cache_dir
 
+from ._version import __version__
+from .models.cache import make_prompt_cache
 from .utils import generate_step, load
 
 
@@ -94,6 +108,13 @@ def convert_chat(messages: List[dict], role_mapping: Optional[dict] = None):
     return prompt.rstrip()
 
 
+@dataclass
+class PromptCache:
+    cache: List[Any] = field(default_factory=list)
+    model_key: Tuple[str, Optional[str]] = ("", None)
+    tokens: List[int] = field(default_factory=list)
+
+
 class ModelProvider:
     def __init__(self, cli_args: argparse.Namespace):
         """Load models on demand and persist them across the whole process."""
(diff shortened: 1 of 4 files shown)


chimezie:
This would not be back wards compatible with any later incorporation of batched input for generate (i.e., #948)

llms/mlx_lm/server.py
```diff
@@ -474,15 +531,15 @@ def handle_completion(
 
     def handle_stream(
         self,
-        prompt: mx.array,
+        prompt: List[int],
```
chimezie, line 534: This in particular.  Can't we handle a single or multiple (batched) prompt that falls back to the behavior for a single prompt by default?

chimezie (changes requested):
Leave door for  open for (#948)

llms/mlx_lm/server.py
awni: We can always update this type to `List[List[int]]` when the time comes.

llms/mlx_lm/server.py
```diff
@@ -496,16 +553,20 @@ def handle_stream(
         stop_sequence_suffix = None
         logging.debug(f"Starting stream:")
 
-        for (token, _), _ in zip(
+        prompt = self.get_prompt_cache(prompt)
+        prompt = mx.array(prompt)
```
angeloskath, line 557: I think this is a small efficiency bug for which you added the `mx.array` on line 562.

Basically, line 600 below casts this back to a list immediately so no need to cast to array here. We can cast only when passed to `generate_step` as you do below.

llms/mlx_lm/server.py
```diff
@@ -559,7 +623,7 @@ def completion_usage_response(
     ):
         response = {
             "id": self.request_id,
-            "system_fingerprint": f"fp_{uuid.uuid4()}",
+            "system_fingerprint": self.system_fingerprint,
```
angeloskath, line 626: ❤️

llms/mlx_lm/server.py
```diff
@@ -602,7 +665,7 @@ def handle_chat_completions(self) -> mx.array:
             prompt = convert_chat(body["messages"], body.get("role_mapping"))
             prompt = self.tokenizer.encode(prompt)
 
-        return mx.array(prompt)
+        return prompt
```
angeloskath, line 668: So this returns a list now? I am not sure why although there is no need to cast to mlx too early I guess. At the very least we probably need to change the return type annotation and docstring.

llms/mlx_lm/server.py
```diff
@@ -614,11 +677,8 @@ def handle_text_completions(self) -> mx.array:
         # Determine response type
         self.request_id = f"cmpl-{uuid.uuid4()}"
         self.object_type = "text_completion"
-
         assert "prompt" in self.body, "Request did not contain a prompt"
-        prompt_text = self.body["prompt"]
-        prompt = self.tokenizer.encode(prompt_text)
-        return mx.array(prompt)
+        return self.tokenizer.encode(self.body["prompt"])
```
angeloskath, line 681: Same as above.

angeloskath (approved):
Looks fantastic! I left a few comments that may or may not need addressing.

PR #1027, opened 2024-10-09
Clear cache during prompt processing

### LINKED ISSUE #1025: MLX_LM: Memory usage much higher than expected
Running the following:

`mlx_lm.generate --model 'mlx-community/meta-llama-3.1-8b-instruct-4bit' --prompt '{10k tokens long prompt}' --max-kv-size-12000`

I would expect to get a memory usage somewhere around ~5gb as reported in this [tweet](https://x.com/awnihannun/status/1824541607468966289) - however, I'm getting approximately 26gb of app memory used and 6gb of wired memory compared to idle

Also doesn't seem to be an isolated issue, seen a similar report [here](https://www.reddit.com/r/LocalLLaMA/comments/1fz6z79/comment/lr3z369/?context=3) where Llama 3.2 3B MLX 4Bit with 16k context is using 15.22gb of memory, compared to 3.67gb used for the same model in 4bit GGUF.

Unfortunately not been able to figure out what might be causing this...

Steps to reproduce:

1. Install MLX_LM
2. Create a 10k token-long prompt (I just copy/pasted some random text until it reached 10k tokens in https://platform.openai.com/tokenizer, obviously slightly inaccurate but it's a ballpark)
3. Run `mlx_lm.generate --model 'mlx-community/meta-llama-3.1-8b-instruct-4bit' --prompt '{10k tokens long prompt}' --max-kv-size-12000` and copy/paste the 10k token prompt into the prompt field
4. Watch activity monitor soar to ~30gb used

Also getting the same behaviour in LMStudio - steps to reproduce are to install it, load up the Llama-3.1-8b-4bit model, then paste the same prompt there

awni:
Closes #1025, see that for discussion / improvement.

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -242,6 +242,7 @@ def generate_step(
         model(y[:prefill_step_size][None], cache=prompt_cache)
         mx.eval([c.state for c in prompt_cache])
         y = y[prefill_step_size:]
+        mx.metal.clear_cache()
 
     y, logprobs = _step(y)
 


angeloskath (approved):
👍

I assume it doesn't really affect the speed since we weren't hitting the cache anyway...

awni:
Yea I didn't see any difference.  (And of course faster for very long prompts since you don't have nearly as much memory pressure).

PR #1028, opened 2024-10-10
Add FLUX finetuning

angeloskath:
This adds a dreambooth example for FLUX-dev finetuning. It still needs a README and possibly some more cleaning but it works well for the dev model. The following is a list of things to do in the future

- Prior preservation
- Ensure it works for schnell (maybe find hyper parameters)
- QLoRA for machines with smaller memory
- Check if we can avoid T5 padding for some extra speed

Distributed training scales almost perfectly, most of my training was with 8 nodes.


flux/dreambooth.py
```diff
@@ -0,0 +1,346 @@
...
+                    (
+                        (width - side) / 2,
+                        (height - side) / 2,
+                        (width + side) / 2,
+                        (height + side) / 2,
+                    )
+                )
+            img = img.resize(self.args.resolution, Image.LANCZOS)
+            img = mx.array(np.array(img))
+            img = (img[:, :, :3].astype(flux.dtype) / 255) * 2 - 1
```
awni, line 65: Should that be `self.flux.dtype`?

angeloskath: Of course it should :-)

flux/dreambooth.py
```diff
@@ -0,0 +1,346 @@
...
+        "--guidance", type=float, default=4.0, help="The guidance factor to use."
+    )
+    parser.add_argument(
+        "--iterations",
+        type=int,
+        default=600,
+        help="How many iterations to train for",
+    )
+    parser.add_argument(
+        "--batch_size",
```
awni, line 159: nit: `batch-size`

flux/dreambooth.py
```diff
@@ -0,0 +1,346 @@
...
+    # setting.
+    mx.random.seed(0xF0F0F0F0 + mx.distributed.init().rank())
+
+    # Load the model and set it up for LoRA training. We use the same random
+    # state when creating the LoRA layers so all workers will have the same
+    # initial weights.
+    flux = FluxPipeline("flux-" + args.model)
+    flux.flow.freeze()
+    with random_state(0x0F0F0F0F):
+        flux.linear_to_lora_layers(args.lora_rank, args.lora_blocks)
```
awni, line 232: I'm not sure if you are using it elsewhere. But if not you could probably get rid of this context manager by doing:

```python
mx.random.seed(0xF0...)
...
flux.linear_to_lora_layers(...)

mx.random.seed(0XF0 .. + rank())
```

flux/dreambooth.py
```diff
@@ -0,0 +1,346 @@
...
+            else:
+                return compute_loss_and_accumulate_grads(
+                    x, t5_feat, clip_feat, guidance, prev_grads
+                )
+
+    print("Create the training dataset.", flush=True)
+    dataset = FinetuningDataset(flux, args)
+    dataset.encode_images()
+    dataset.encode_prompts()
+    guidance = mx.full((args.batch_size,), args.guidance, dtype=flux.dtype)
```
awni, line 315: possible nit:
```suggestion
    guidance = mx.array(args.guidance, dtype=flux.dtype)
```

flux/txt2image.py
```diff
@@ -0,0 +1,142 @@
...
+        flux.fuse_lora_layers()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Generate images from a textual prompt using stable diffusion"
+    )
+    parser.add_argument("prompt")
+    parser.add_argument("--model", choices=["schnell", "dev"], default="schnell")
+    parser.add_argument("--n_images", type=int, default=4)
```
awni, line 46: nit `--n-images`

flux/txt2image.py
```diff
@@ -0,0 +1,142 @@
...
+    parser.add_argument(
+        "--image_size", type=lambda x: tuple(map(int, x.split("x"))), default=(512, 512)
+    )
+    parser.add_argument("--steps", type=int)
+    parser.add_argument("--guidance", type=float, default=4.0)
+    parser.add_argument("--n_rows", type=int, default=1)
```
awni, line 52: nit `--n-rows`

flux/txt2image.py
```diff
@@ -0,0 +1,142 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Generate images from a textual prompt using stable diffusion"
+    )
+    parser.add_argument("prompt")
+    parser.add_argument("--model", choices=["schnell", "dev"], default="schnell")
+    parser.add_argument("--n_images", type=int, default=4)
+    parser.add_argument(
+        "--image_size", type=lambda x: tuple(map(int, x.split("x"))), default=(512, 512)
```
awni, line 48: nit `--image-size`

flux/txt2image.py
```diff
@@ -0,0 +1,142 @@
...
+    )
+    parser.add_argument("--steps", type=int)
+    parser.add_argument("--guidance", type=float, default=4.0)
+    parser.add_argument("--n_rows", type=int, default=1)
+    parser.add_argument("--decoding_batch_size", type=int, default=1)
```
awni, line 53: nit `--decoding-batch-size`

flux/txt2image.py
```diff
@@ -0,0 +1,142 @@
...
+    del flux.clip
+
+    # Actual denoising loop
+    for x_t in tqdm(latents, total=args.steps):
+        mx.eval(x_t)
+
+    # The following is not necessary but it may help in memory constrained
+    # systems by reusing the memory kept by the flow transformer.
+    del flux.flow
+    peak_mem_generation = mx.metal.get_peak_memory() / 1024**3
```
awni, line 108: This would be max(generation RAM , text encoding RAM)... its very minor, but one way to avoid that is to use `mx.metal.reset_peak_memory` before the generation loop.

awni (approved):
This is beautifully done! I didn't read the model code just the training and generation loops. I will peruse it but let's not let that slow us down from landing it.

I think all that's missing is a readme and maybe some descriptions in the txt2image arguments.

Also probably should put the copyright stamp in the files.

flux/dreambooth.py
```diff
@@ -0,0 +1,346 @@
...
+            self.t5_features.append(t5_feat)
+            self.clip_features.append(clip_feat)
+
+    def iterate(self, batch_size):
+        xs = mx.concatenate(self.latents)
+        t5 = mx.concatenate(self.t5_features)
+        clip = mx.concatenate(self.clip_features)
+        mx.eval(xs, t5, clip)
+        while True:
+            indices = mx.random.randint(0, len(self.latents), (batch_size,))
```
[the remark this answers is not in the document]
angeloskath: Done, I also added some mild input augmentation with random crop/pan.

flux/dreambooth.py
angeloskath: I did change it like that, but in general I kinda like these context managers. I think it might not be a bad idea to add such a context manager in C++ if it can be nicely implemented...

awni: Yea I can see that. I think it should be pretty easy to do it in the Python bindings the same way we do for `mx.stream`.

akkikumar72:
Awesome work 🎊 
Will be nice to have a readme for this, so we can try/test it in weekend

angeloskath:
@akkikumar72 I totally agree. I am AFK for a few hours but I will write the README and get it merged tonight (PST).

flux/dreambooth.py
angeloskath: Yeah unfortunately this wouldn't work but `mx.array([args.guidance], dtype=flux.dtype)` probably would 🤷‍♂️.

flux/txt2image.py
angeloskath: Done, and fixed a type that slipped in there as well (generation memory was reported twice).

nattaylor:
This is amazing.  Any idea when you might have QLoRA implemented? :)

PR #1036, opened 2024-10-12
FLUX: update README.md

madroidmaq:


madroidmaq:
Just discovered that there will be some overlapping content with #1035.

angeloskath (approved):
Thanks for the additions to the readme!

PR #603, opened 2024-03-21
Switch to fast RMS/LN Norm

awni:
- Switch RMSNorm to `nn.RMSNorm` to get fast rms
- Switch LayerNorm to `nn.LayerNorm` to get fast LN
- Update a few more models to use SDPA
- Clean up old lora to reduce some surface area and point to MLX LM for more features.

NB tests won't clear until 0.8 is out, but they clear locally for me with latest MLX

PR #1037, opened 2024-10-12
Fix PLaMo model to support Grouped Query Attention

mitmul:
The model definition of PLaMo was updated in https://github.com/ml-explore/mlx-examples/pull/603 and it seems to have removed the shared key/value heads so that the output of PLaMo is broken for now.

This PR fixes the issue by reverting the removed part of shared kv heads (as like in llama models)

### DIFF
--- a/llms/mlx_lm/models/plamo.py
+++ b/llms/mlx_lm/models/plamo.py
@@ -89,6 +89,9 @@ class Attention(nn.Module):
             queries = self.rotary_emb(queries)
             keys = self.rotary_emb(keys)
 
+        keys = mx.tile(keys, [1, self.config.n_shared_head, 1, 1])
+        values = mx.tile(values, [1, self.config.n_shared_head, 1, 1])
+
         output = mx.fast.scaled_dot_product_attention(
             queries,
             keys,


llms/mlx_lm/models/plamo.py
```diff
@@ -81,6 +81,9 @@ def __call__(
             0, 2, 1, 3
         )
 
+        keys = mx.tile(keys, [1, self.config.n_shared_head, 1, 1])
+        values = mx.tile(values, [1, self.config.n_shared_head, 1, 1])
```
awni, line 85: This is not be necessary. The `scaled_dot_product_attention` handles GQA. Was it crashing for you?

mitmul: Hm, I got the broken output from pfnet/plamo-13b before applying this suggested change.

I would like to confirm whether I correctly understood the meaning of your comment. When you say that there is no need to apply `mx.tile()` in advance, does that mean the following code should pass the assertion? I actually tried running it, but the following code failed with an error.

```python
import mlx.core as mx

keys = mx.random.uniform(low=-1.0, high=1.0, shape=(1, 5, 43, 128))
values = mx.random.uniform(low=-1.0, high=1.0, shape=(1, 5, 43, 128))
queries = mx.random.uniform(low=-1.0, high=1.0, shape=(1, 40, 43, 128))

no_tiled = mx.fast.scaled_dot_product_attention(queries, keys, values, scale=1.0)

keys = mx.tile(keys, [1, 8, 1, 1])
values = mx.tile(values, [1, 8, 1, 1])

tiled = mx.fast.scaled_dot_product_attention(queries, keys, values, scale=1.0)

assert mx.allclose(no_tiled, tiled).item()
```

awni: I thought that test should pass but either I'm mistaken or there is a bug because it doesn't pass for me either.

awni: Actually I misunderstood. That test should not pass.  Our GQA assumes the keys are repeated in the head dimension not tiled. We'd have to rearrange the queries in order to fix that.

So something like:

```
queries = queries.reshape(bsz, q_len, self.config.n_shared_head, self.k_num_heads, self.qk_dim)
queries = queries.transpose(0, 3, 2, 1, 4).flatten(1, 2)
```

This is better because it let's you avoid blowing up the KV cache size when you explicitly tile the keys and values. 

Another alternative which is probably just as good is to tile *after* updating the KV cache and applying RoPE.

awni: So concretely could you please move the `tile` below line 94 then we can merge this. Sorry for the misunderstanding.

mitmul: Thank you very much for the clarification and not I understand what's happening with the current code. I moved that part after KV cache updating by 
55a0282.

awni (approved):
Thanks!

PR #1040, opened 2024-10-12
Make llm async eval less brittle

awni:
This is a partial fix for inadvertently slowing `async_eval` by putting in synchronization points:

- Everything that comes out of `generate_step` is already in the `async_eval`
- Everything outside of `generate_step` should be in a separate stream

After https://github.com/ml-explore/mlx/pull/1478 lands we should consider putting the whole async eval in it's own unique stream to avoid making this brittle to downstream code that processes `logprobs`.


angeloskath (approved):
🚀

PR #1049, opened 2024-10-15
FLUX: save train config

madroidmaq:
Save some parameters during training in the `output_dir/adapter_config.json` file for easy inspection and review.
At the same time, save the `output_dir/final_adapter.safetensors` file at the end of training to facilitate script execution and avoid the problem of non-integer training steps.


angeloskath (approved):
This looks good, thanks!

I will do a training to make sure I didn't break anything with my small changes and then merge it.

PR #1050, opened 2024-10-15
fix: typo on flux model preloading

zakame:
Still referring to the sd version of the pipeline class.


zakame:
Thanks for the example, learning quite a bit here so far! 🎉 👍

PR #1062, opened 2024-10-22
Override dtype even when quantizing

### LINKED ISSUE #1057: Possible Gemma bugs
I came across https://github.com/huggingface/transformers/pull/29402 and I think it applies to MLX's gemma implementation too.

It has 3 possible problems:
* MLP's activation function should be `nn.gelu_approx` instead of `nn.gelu`.
* The `h = h * (self.args.hidden_size**0.5)` normalization needs casting to `h.dtype`.
* The `1.0 + self.weight` in `RMSNorm` needs casting to float32.

I think MLX casts scalar to array's dtype in binary ops so they may not all apply.

awni:
This change let's you pick the non-quantized weight dtype even when using quantization. There are some models (namely Gemma 2 27B) which can have very large values in the activations that overflow float16.

Also closes #1057

### DIFF
--- a/llms/mlx_lm/convert.py
+++ b/llms/mlx_lm/convert.py
@@ -31,7 +31,7 @@ def configure_parser() -> argparse.ArgumentParser:
     )
     parser.add_argument(
         "--dtype",
-        help="Type to save the parameters, ignored if -q is given.",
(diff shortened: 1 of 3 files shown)


barronalex (approved):
Nice!

PR #1072, opened 2024-10-27
Fix detokenizer space match for quote

### LINKED ISSUE #1073: [BUG] missing spaces in response with mlx-lm 0.19.1 and 0.19.2
**Describe the bug**
I have fine-tuned the meta-llama/Meta-Llama-3.1-8B-Instruct model in different variants with mlx-lm and Lora to generate Dart-Unit-Tests. The tuning of the models was done with different versions of mlx-lm (0.19.0, 0.19.1 and 0.19.2). I also used mlx-lm to fuse the adapters to the LLM again. Now I have noticed that during inference/generation with 0.19.1 and 0.19.2 spaces are missing in the generated import-statements  when they are followed by a ', for example: `import'package:flutter/material.dart';` vs `import 'package:flutter/material.dart';`. Regardless of which version the model was fine-tuned with, if I downgrade to 0.19.0 the generation works correctly. In the training data the import are ok. Maybe a problem with tokenising?

For example

Generation-Response with 0.19.0
```
Here are the widget tests for the `ListTile`:

```dart
import 'package:flutter/material.dart';
import 'package:flutter_bloc/flutter_bloc.dart';
import 'package:flutter_test/flutter_test.dart';
```

Generation-Response with 0.19.1

```
Here are the tests for the `ListTile` class:

```dart
import'package:flutter/material.dart';
import'package:flutter_bloc/flutter_bloc.dart';
import'package:flutter_test/flutter_test.dart';
```

Generation-Response with 0.19.1
```
body:'body',
timestamp: DateTime.now(),
```

**To Reproduce**

code snippet used to genrate code
```python
model_path = "results/llama3_1_8B_instruct_lora/tuning_09/lora_fused_model"
model, tokenizer = load(model_path)

generation_config = GenerationConfig.from_pretrained("meta-llama/Meta-Llama-3.1-8B-Instruct")
generation_args = {
    "temp": generation_config.temperature,
    "top_p": generation_config.top_p,
}

prompt_template = "Generate {test_type}s in Dart for the following {code_type}.\n" \
                  "### Path: {code_file_path}\n" \
                  "### Code:\n" \
                  "{code}\n" \
                  "### Test:\n"

def build_prompt(
        code: str, 
        code_file_path: str, 
        test_type: str, 
        code_type: str
) -> str:
    prompt = prompt_template.format(
        test_type=test_type,
        code_type=code_type,
        code_file_path=code_file_path,
        code=code
    )
    messages = [{"role": "user", "content": prompt}]
    return tokenizer.apply_chat_template(
        messages, tokenize=False, add_generation_prompt=True
    )

def generate_test(code, code_file_path, code_type, test_type) -> tuple[str, str]:
    logging.info(f"Generating test for {os.path.basename(code_file_path)}")
    prompt =  build_prompt(
        code=code, 
        code_file_path=code_file_path, 
        code_type=code_type, 
        test_type=test_type
    )
    print(prompt)
```

**Expected behavior**
No missing spaces before '.

**Desktop (please complete the following information):**
 - OS Version: [MacOS 15.0.1]
 - mlx-lm [0.19.1 und 0.19.2]
 - mlx [0.19.1]

awni:
I introduces a bug when trying to port the `clean_up_tokenization_spaces` from HF.

Their match is actually to find and replace instances of ` ' ` with `'`. That one specifically requires a special case to post-process which is tedious, so I just removed that case as I have never seen it come up. But if it does we can revisit.

For reference, the [transformers matching conditions](https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_utils_base.py#L4052).

Closes #1073

### DIFF
--- a/llms/mlx_lm/__init__.py
+++ b/llms/mlx_lm/__init__.py
@@ -1,4 +1,10 @@
 # Copyright © 2023-2024 Apple Inc.
 
 from ._version import __version__
+
+import os
+os.environ["TRANSFORMERS_NO_ADVISORY_WARNINGS"] = "1"
+
 from .utils import convert, generate, load, stream_generate
+
+
--- a/llms/mlx_lm/__init__.py
+++ b/llms/mlx_lm/__init__.py
@@ -1,10 +1,9 @@
 # Copyright © 2023-2024 Apple Inc.
 
+import os
+
 from ._version import __version__
 
-import os
 os.environ["TRANSFORMERS_NO_ADVISORY_WARNINGS"] = "1"
 
 from .utils import convert, generate, load, stream_generate
-
-
--- a/llms/mlx_lm/tokenizer_utils.py
+++ b/llms/mlx_lm/tokenizer_utils.py
@@ -169,7 +169,7 @@ class BPEStreamingDetokenizer(StreamingDetokenizer):
     """
 
     _byte_decoder = None
-    _space_matches = (".", "?", "!", ",", "'", "n't", "'m", "'s", "'ve", "'re")
+    _space_matches = (".", "?", "!", ",", "n't", "'m", "'s", "'ve", "'re")
 
     def __init__(self, tokenizer):
 
--- a/llms/tests/test_tokenizers.py
+++ b/llms/tests/test_tokenizers.py
@@ -51,6 +51,9 @@ class TestTokenizers(unittest.TestCase):
         tokens = tokenizer.encode("3 3")
         check(tokens)
 
+        tokens = tokenizer.encode("import 'package:flutter/material.dart';")
+        check(tokens)
+
     def test_tokenizers(self):
         tokenizer_repos = [
             ("mlx-community/Qwen1.5-0.5B-Chat-4bit", BPEStreamingDetokenizer),


llms/mlx_lm/__init__.py
```diff
@@ -1,4 +1,10 @@
 # Copyright © 2023-2024 Apple Inc.
 
 from ._version import __version__
+
+import os
+os.environ["TRANSFORMERS_NO_ADVISORY_WARNINGS"] = "1"
+
```
awni, line 7: Get's rid of the warning:
```text
None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
```

llms/mlx_lm/__init__.py
```diff
@@ -1,4 +1,9 @@
 # Copyright © 2023-2024 Apple Inc.
 
+import os
+
 from ._version import __version__
+
+os.environ["TRANSFORMERS_NO_ADVISORY_WARNINGS"] = "1"
```
hschaeufler, line 7: Maybe a comment would be nice, why you set the flag.

hschaeufler (approved):
Blanks are now there: import ‘package:flutter/material.dart’; Thank you for fixing

angeloskath (approved):
👍

PR #1075, opened 2024-10-29
Quantized KV Cache

barronalex:
Add `QuantizedKVCache` for long contexts.

## Performance
All tests with Qwen2.5 7B Instruct 4 bit on M2 Ultra

- We're about 50% faster with an 8 bit quantized KV cache at the maximum 128K sequence length.
- We start seeing speed improvements after about 6K sequence length over the fp16 KV cache.
- Memory is also much lower: `11.5GB` vs `8GB` for 128K.

long
long
long

## Accuracy
For 8 bit I'm seeing identical generations to `fp16` in most cases. 4 bit isn't any faster than 8 bit currently so I think that's the sweet spot. Perplexity experiments to follow.

## Usage

Specify `--kv-bits=8` to enable 8 bit KV cache quantization.
```
mlx_lm.generate --model mlx-community/Qwen2.5-7B-Instruct-4bit --prompt "Write a long story about Einstein." --kv-bits 8
```

Optionally, you can specify `--quantized-kv-start` to set the minimum sequence length before we start quantizing the KV cache.  We don't quantize the entire time because it's faster to run in `fp16` for the first few thousand steps due to the additional overhead.

`mlx_lm.cache_prompt` also supports quantizing the KV cache now.

### DIFF
--- a/llms/mlx_lm/generate.py
+++ b/llms/mlx_lm/generate.py
@@ -258,9 +258,9 @@ def main():
         top_p=args.top_p,
         max_kv_size=args.max_kv_size,
         prompt_cache=prompt_cache if using_cache else None,
-        quantized_kv_start=args.quantized_kv_start,
-        kv_group_size=args.kv_group_size,
         kv_bits=args.kv_bits,
+        kv_group_size=args.kv_group_size,
+        quantized_kv_start=args.quantized_kv_start,
     )
     if not args.verbose:
         print(response)
--- a/llms/mlx_lm/generate.py
+++ b/llms/mlx_lm/generate.py
@@ -7,7 +7,7 @@ import sys
 import mlx.core as mx
 
 from .models.cache import QuantizedKVCache, load_prompt_cache
-from .utils import check_quantized_kv_args, generate, load
+from .utils import generate, load
 
 DEFAULT_PROMPT = "hello"
 DEFAULT_MAX_TOKENS = 100
@@ -15,6 +15,7 @@ DEFAULT_TEMP = 0.0
 DEFAULT_TOP_P = 1.0
 DEFAULT_SEED = 0
 DEFAULT_MODEL = "mlx-community/Llama-3.2-3B-Instruct-4bit"
+DEFAULT_QUANTIZED_KV_START = 5000
 
 
 def str2bool(string):
@@ -108,24 +109,24 @@ def setup_arg_parser():
         help="A file containing saved KV caches to avoid recomputing them",
     )
     parser.add_argument(
-        "--quantized-kv-start",
-        help="Use a quantized KV cache from this step onwards.",
+        "--kv-bits",
         type=int,
+        help="Number of bits for KV cache quantization. "
+        "Defaults to no quantization.",
         default=None,
     )
     parser.add_argument(
         "--kv-group-size",
         type=int,
-        help="Group size for kv cache quantization. "
-        "--quantized-kv-start must be provided to have an effect.",
+        help="Group size for KV cache quantization.",
         default=64,
     )
     parser.add_argument(
-        "--kv-bits",
+        "--quantized-kv-start",
+        help="When --kv-bits is set, start quantizing the KV cache "
+        "from this step onwards.",
         type=int,
-        help="Number of bits for kv cache quantization. "
-        "--quantized-kv-start must be provided to have an effect.",
-        default=8,
+        default=DEFAULT_QUANTIZED_KV_START,
     )
     return parser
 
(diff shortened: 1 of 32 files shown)


llms/mlx_lm/models/cache.py
```diff
@@ -126,6 +129,124 @@ def is_trimmable(self):
         return False
 
 
+def quantized_scaled_dot_product_attention(
```
angeloskath, line 132: I would probably move this to `base.py` and replace it with a function called `scaled_dot_product_attention` that takes the same arguments as `mx.fast.scaled_dot_product_attention` plus the `cache`, then it routes to this or the `mx.fast` one.

This will also simplify the implementations a lot as the if goes in here and the models don't really care about the type of cache they are using.

llms/mlx_lm/utils.py
```diff
@@ -220,11 +247,16 @@ def logit_bias_processor(_, logits):
 
     # Create the KV cache for generation
     if prompt_cache is None:
-        prompt_cache = cache.make_prompt_cache(model, max_kv_size)
+        prompt_cache = cache.make_prompt_cache(
```
angeloskath, line 250: Nitpick: There are a few things that have been split into multiple lines and I wonder if some config is different on your machine.

llms/mlx_lm/utils.py
```diff
@@ -124,6 +127,27 @@ def apply_repetition_penalty(logits: mx.array, tokens: mx.array, penalty: float)
...
+
+def maybe_quantize_kv_cache(prompt_cache, quantized_kv_start, kv_group_size, kv_bits):
+    if (
+        quantized_kv_start
+        and prompt_cache[0].offset > quantized_kv_start
+        and not isinstance(prompt_cache[0], cache.QuantizedKVCache)
+    ):
+        return [
+            c.to_quantized(group_size=kv_group_size, bits=kv_bits) for c in prompt_cache
+        ]
```
angeloskath, line 147: Why not change the cache in place? Is there a reason we may want to return it? Simplifies step as it doesn't need the nonlocal keyword however it does an in-place operation that may be less noticeable. Have no idea which is better really.

llms/mlx_lm/utils.py
```diff
@@ -235,6 +267,10 @@ def _step(y):
             for processor in logits_processor:
                 logits = processor(tokens, logits)
 
+        prompt_cache = maybe_quantize_kv_cache(
+            prompt_cache, quantized_kv_start, kv_group_size, kv_bits
+        )
```
angeloskath, line 272: Going from the comment above this could simply become:

```python
maybe_quantize_kv_cache(
    prompt_cache, quantized_kv_start, kv_group_size, kv_bits
)
```

angeloskath (changes requested):
This is fantastic!

I only have one comment regarding the placement of `quantized_sdpa` and the corresponding if statements. Can't wait to process 100k context on an 8GB device

llms/mlx_lm/generate.py
```diff
@@ -107,6 +107,26 @@ def setup_arg_parser():
...
+        "--quantized-kv-start",
+        help="Use a quantized KV cache from this step onwards.",
+        type=int,
+        default=None,
+    )
+    parser.add_argument(
+        "--kv-group-size",
+        type=int,
+        help="Group size for kv cache quantization. "
+        "--quantized-kv-start must be provided to have an effect.",
```
awni, line 120: Why require that param? Is `0` not a reasonable default?

llms/mlx_lm/models/cache.py
```diff
@@ -320,6 +451,16 @@ def trim(self, n):
...
 
+    def to_quantized(self, group_size: int = 64, bits: int = 4) -> QuantizedKVCache:
+        quant_cache = QuantizedKVCache(group_size=group_size, bits=bits)
+        quant_cache.offset = self.offset
+        if self.keys is not None:
+            quant_cache.keys = mx.quantize(self.keys, group_size=group_size, bits=bits)
+            quant_cache.values = mx.quantize(
+                self.values, group_size=group_size, bits=bits
+            )
+        return quant_cache
```
awni, line 462: I'm wondering if it makes sense to allow a rotating kv cache to be quantized? The quantized kv cache doesn't rotate as far as I can tell and if this cache is in a rotated state (e.g. the write head is on the same as the offset), then the behavior might is unclear / probably wrong?

awni:
Really nicely done, can't wait to land this! I think the suggestion by @angeloskath to move to a common sdpa which takes a cache is the main thing to change. Then we can sed replace `mx.fast.scaled...` in all the model files and have this work everywhere!

llms/mlx_lm/models/cache.py
barronalex: Great call! I've updated it everywhere in `mlx_lm`

llms/mlx_lm/utils.py
barronalex: Changed it to in place. I think it's marginally cleaner.

llms/mlx_lm/generate.py
barronalex: I changed this around so that `--kv-bits` determines whether we do KV cache quantization or not since I think it's a bit more intuitive that way.

I have the default `quantized-kv-start=5000` since that's the best performance trade off at the moment.

llms/mlx_lm/models/cache.py
barronalex: I think you're right, for now I've put in a NYI error.

angeloskath (approved):
🚀

llms/mlx_lm/models/cache.py
```diff
@@ -126,6 +129,88 @@ def is_trimmable(self):
...
+        self.bits = bits
+
+    def update_and_fetch(self, keys, values):
+        B, n_kv_heads, num_steps, k_head_dim = keys.shape
+        prev = self.offset
+
+        if self.keys is None or (prev + num_steps) > self.keys[0].shape[-2]:
+            el_per_int = 8 * mx.uint32.size // self.bits
+            new_steps = (self.step + num_steps - 1) // self.step * self.step
+            shape = (B, n_kv_heads, new_steps, k_head_dim // el_per_int)
```
awni, line 148: One of the models (deep seek) has a different `v_head_dim` than `k_head_dim` (which is why we use two different shapes in the KV cache.  Can you do the same here so it works with that model as well?

barronalex: Good catch! Will do

awni (approved):
🚀

PR #1080, opened 2024-11-01
Whisper improvements

awni:
- Allow safetensors as input for faster loading
- Use fewer and more carefully chosen eval
- Precise softmax instead of casting
- Async eval in decode loop

Benchmarks on M2 Ultra

Model | Pre MLX 0.18.0 | Pre MLX Main | Post MLX Main |
--- | ---- | --- | ----
mlx-community/whisper-large-v3-turbo | 16.093 s | 15.642 s | 12.246 s

```python
import time
import mlx_whisper

tic = time.time()
text = mlx_whisper.transcribe("/Users/awnihannun/.cache/whisper/alice.mp3", path_or_hf_repo="mlx_models")["text"]
toc = time.time()

print(f"Time {toc-tic:.3f}")
```

### DIFF
--- a/whisper/convert.py
+++ b/whisper/convert.py
@@ -181,7 +181,7 @@ def load_torch_weights_and_config(
         )
 
     if name_or_path.endswith(".pt"):
-        checkpoint = torch.load(name_or_path, map_location="cpu")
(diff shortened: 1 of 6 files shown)


barronalex (approved):
Really nice!!

PR #1089, opened 2024-11-03
chore(mlx-lm): add max token arg for mlx_lm.chat

mzbac:
adding max token argument so user can use mlx_lm.chat to generate long reponse


llms/mlx_lm/chat.py
```diff
@@ -11,6 +11,7 @@
 DEFAULT_TEMP = 0.0
 DEFAULT_TOP_P = 1.0
 DEFAULT_SEED = 0
+DEFAULT_MAX_TOKENS = 100
```
awni, line 14: Maybe this should be a bit higher.. like 256 or 512?

awni (approved):
Thanks for the addition!

PR #1094, opened 2024-11-06
[MLX LM] Sampler refactor + a few improvements

### LINKED ISSUE #1032: [Feature Request] Hook up the Min-P sampler to CLI so it can be used in all the same places Top-P can
Min-P is a favourite amongst the chatbot community e.g. r/LocalLlama as a sampling method over Top-P, since by eliminating any tokens that are below a likeliness threshold instead of just choosing the top 40 tokens (and thus potentially still including some that have low confidence), you can turn up the temperature of a model and still retain generally coherent text. This results in much more diverse outputs. I myself use min-P a lot when I'm using chatbots such as in LMStudio, often opting to turn off Top-P completely.

It'd be great if MLX_LM.chat had an option to set min-P, the same as you can currently set top-P

(I'd be happy to turn this into a pull request if needed! Though I can't vouch for my coding skills lol)

awni:
This is part 1 of a potential second part diff to refactor samplers and logits processors outside of the `generate_step` function. The idea being that function is getting too fat.. and it will also be much more flexible to experiment with alternative samplers / logits processors without hacking `generate_step`.

To start this diff:
- Refactors out the sampler and logits processor creation into sampler_utils.py
- Merges the server generation into a single function for stream and completion
- `stream_generate` is more flexible: returns logits and token as well as text and accepts pre-tokenized inputs
- Use wired limit in `stream_generate`
- Use `stream_generate` in the server instead of `generate_step`
- Put generation in a separate stream to avoid breaking async eval when not using the CPU 
- Closes #1032

### DIFF
--- a/llms/mlx_lm/cache_prompt.py
+++ b/llms/mlx_lm/cache_prompt.py
@@ -152,6 +152,7 @@ def main():
 
         model(y[:step_size][None], cache=cache)
         mx.eval([c.state for c in cache])
+        mx.metal.clear_cache()
         processed += min(y.size, step_size)
         y = y[step_size:]
         current = time.time()
@@ -165,14 +166,13 @@ def main():
         )
 
     print()
-    print(f"Peak memory: {mx.metal.get_peak_memory() / 2**30:.3f} GB")
+    print(f"Peak memory: {mx.metal.get_peak_memory() / 1e9:.3f} GB")
 
     print("Saving...")
     metadata = {}
     metadata["model"] = args.model
     metadata["chat_template"] = tokenizer.chat_template
     metadata["tokenizer_config"] = json.dumps(tokenizer_config)
-    print(f"Peak memory: {mx.metal.get_peak_memory() / 2**30:.3f} GB")
     save_prompt_cache(args.prompt_cache_file, cache, metadata)
 
 
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -300,10 +300,9 @@ def stream_generate(
             range(max_tokens),
             generate_step(prompt_tokens, model, **kwargs),
         ):
-            if token == tokenizer.eos_token_id:
-                break
             detokenizer.add_token(token)
-
+            if n == (max_tokens - 1) or token == tokenizer.eos_token_id:
+                break
             # Yield the last segment if streaming
             yield detokenizer.last_segment, token, logits
 
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -204,8 +204,7 @@ def generate_step(
             when ``kv_bits`` is non-None. Default: ``0``.
 
     Yields:
-        Generator[Tuple[mx.array, mx.array], None, None]: A generator producing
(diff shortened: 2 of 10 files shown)


llms/mlx_lm/cache_prompt.py
```diff
@@ -152,6 +152,7 @@ def main():
 
         model(y[:step_size][None], cache=cache)
         mx.eval([c.state for c in cache])
+        mx.metal.clear_cache()
```
awni, line 155: Clear cache here like in `generate_step` to avoid hording RAM.

llms/mlx_lm/cache_prompt.py
```diff
@@ -165,14 +166,13 @@ def main():
         )
 
     print()
-    print(f"Peak memory: {mx.metal.get_peak_memory() / 2**30:.3f} GB")
+    print(f"Peak memory: {mx.metal.get_peak_memory() / 1e9:.3f} GB")
```
awni, line 169: Changed all of these to `1e9` as technically that's GB and we were slightly under counting.

llms/tests/test_prompt_cache.py
```diff
@@ -299,7 +299,7 @@ def test_cache_to_quantized(self):
         ):
             i += 1
             self.assertEqual(tok, toks[i])
-            self.assertTrue(mx.allclose(logits, all_logits[i], rtol=1e-2))
+            self.assertTrue(mx.allclose(logits, all_logits[i], rtol=2e-2))
```
awni, line 302: Test was flaky on M3 at 1e-2 🤷‍♂️

llms/mlx_lm/utils.py
```diff
@@ -345,23 +289,26 @@ def stream_generate(
...
 
-        # Yield the last segment if streaming
-        yield detokenizer.last_segment
+            # Yield the last segment if streaming
+            yield detokenizer.last_segment, token, logits
 
-    detokenizer.finalize()
-    yield detokenizer.last_segment
+        detokenizer.finalize()
+        yield detokenizer.last_segment, token, logits
```
awni, line 311: TODO fix this.. it repeats if you hit the max token count

angeloskath (approved):
The refactoring is awesome!

There is a bit of a large overlap between `stream_generate` and `generate` (in terms of implementation but I think it is fine.

awni:
I have some questions that you may or may not have opinions on. Mostly for a follow up PR:

- Thinking about moving `make_sampler` and `make_logits_processor` outside of `generate_step` so that one can pass in any sampler / logit processor to that function. Thoughts?
- Merging `stream_generate` and `generate`: yes I want to. `generate` could easily be a convenience wrapper around `stream_generate` which does the optional formatting and verbose logging.
- Given the above, I'm considering moving timing/memory reporting into `stream_generate` to simplify needing to do this in multiple places (it's also easy to do incorrectly) which returns a richer output (maybe a data class with all the fields). I don't want to overcomplicate it.. but that might make it easier to use.

At any rate.. I'll probably put together a PR for some of that soon and we can discuss more concretely. But that's kind of where this is heading in case you have early thoughts on it.

PR #1099, opened 2024-11-08
Generation refactor: part 2

awni:
- Add a generation response data class for `stream_generate`
- `generate` now uses `stream_generate`
- Server also uses stats recorded by `stream_generate`
- Maybe controversial: I removed the color formatting option. This is broken and there is not a simple fix given that tokens to unicode characters is many-to-many. So we can add it back as a hack which works for most cases or just remove it as I don't think it's used much anymore..

### DIFF
--- a/llms/mlx_lm/generate.py
+++ b/llms/mlx_lm/generate.py
@@ -98,11 +98,6 @@ def setup_arg_parser():
         help="Log verbose output when 'True' or 'T' or only print the response when 'False' or 'F'",
     )
     parser.add_argument(
-        "--colorize",
-        action="store_true",
-        help="Colorize output based on T[0] probability",
-    )
-    parser.add_argument(
         "--max-kv-size",
         type=int,
         help="Set the maximum key-value cache size",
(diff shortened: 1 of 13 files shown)


llms/mlx_lm/generate.py
```diff
@@ -137,33 +132,6 @@ def setup_arg_parser():
     return parser
 
 
-def colorprint(color, s):
```
chimezie, line 140: Do you mind elaborating on what you mean by the token to Unicode mapping is many-to-mant and how this functionality it is broken? I actually make use of this functionality and find it very useful for visually incorporating token probabilities into how the generation can be 'interpreted' or even open up the door for more sophisticated things  [this](https://www.reddit.com/r/LocalLLaMA/comments/1girzia/exploring_ais_inner_alternative_thoughts_when/)

awni: Sometimes a single token correspond to multiple unicode characters in the output. Sometimes multiple tokens correspond to a single unicode character in the output.

awni: For the later case you'd have to combine the per-step probabilities to get a single color.

awni: You can also get the wrong result when decoding with color printing because it's not always possible to decode a chunk of correctly at each time step.

So the implementation for this needs to be rethought a bit..

llms/mlx_lm/utils.py
```diff
@@ -334,64 +381,37 @@ def generate(
...
-       kwargs: The remaining options get passed to :func:`generate_step`.
-          See :func:`generate_step` for more details.
+       kwargs: The remaining options get passed to :func:`stream_generate`.
+          See :func:`stream_generate` for more details.
     """
-    if not isinstance(tokenizer, TokenizerWrapper):
-        tokenizer = TokenizerWrapper(tokenizer)
-
+    if formatter is not None:
+        print("Text formatting is deprecated and will be removed in the next version.")
```
angeloskath, line 388: Not sure this makes it clear that it is not used anymore 🤔

angeloskath (approved):
I like it. This is a backwards incompatible change in more ways than 1 but I think it is a clear improvement. If we do want to add features like colorize these should be implemented in the `generate` script rather in the `generate` function.

Issue #1095, opened 2024-11-06
LoRA trains on tokens with duplicated BOS for some models


chimezie:
During LoRa training, iterate_batches [calls](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/tuner/trainer.py#L104) `tokenizer.encode()` (with default arguments) on the dataset item, which in the case of ChatDataset or CompletionsDataset, returns the raw text with the model's chat template applied:

```python
tokenizer.apply_chat_template(.. conversation .., tokenize=False, add_generation_prompt=True)
```

Invoking `tokenizer.encode` with default arguments on the result of this can produced a token sequence with duplicated BOS tokens:

```python
In [..]: tokenizer = AutoTokenizer.from_pretrained("mistralai/Mistral-Nemo-Instruct-2407")

In [..]: message = [{"role": "user", "content": "Who is on first?"}, {"role": "assistant", "content": "Is that a trick question?"}]

In [..]: print(tokenizer.encode(tokenizer.apply_chat_template(message, tokenize=False, add_generation_prompt=True), add_special_tokens=False))
[1, 3, 31500, 1395, 1408, 2158, 1063, 4, 5356, 1455, 1261, 31544, 4098, 1063, 2]

In [..]: print(tokenizer.apply_chat_template(message, tokenize=True, add_generation_prompt=True))
[1, 3, 31500, 1395, 1408, 2158, 1063, 4, 5356, 1455, 1261, 31544, 4098, 1063, 2]

In [..]: print(tokenizer.encode(tokenizer.apply_chat_template(message, tokenize=False, add_generation_prompt=True)))
[1, 1, 3, 31500, 1395, 1408, 2158, 1063, 4, 5356, 1455, 1261, 31544, 4098, 1063, 2]

In [..]: tokenizer = AutoTokenizer.from_pretrained("google/gemma-2-2b-it")

In [..]: print(tokenizer.apply_chat_template(message, tokenize=True, add_generation_prompt=True))
[2, 106, 1645, 108, 6571, 603, 611, 1370, 235336, 107, 108, 106, 2516, 108, 2437, 674, 476, 17760, 2872, 235336, 107, 108, 106, 2516, 108]

In [..]: print(tokenizer.encode(tokenizer.apply_chat_template(message, tokenize=False, add_generation_prompt=True)))
[...]

PR #1090, opened 2024-11-04
Generalize HF datasets to a collection of HF datasets via `hf_datasets`

chimezie:
Also adds support for custom chat HF datasets (#1088), and fixes (#1087)

PR #1103, opened 2024-11-10
Completion only fine-tuning of instruction models with collections of HF datasets

chimezie:
PR Merge of #825, #1090 and fix for #1095 due to combined benefit for use in fine-tuning instruction models with HF completion datasets

### DIFF
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -166,7 +166,7 @@ def iterate_completion_batches(
             for j in batch_idx[i]:
                 prompt, completion = dataset.get_prompt_and_completion(j)
                 prompt_lengths.append(input_length(prompt, completion, tokenizer))
-                full_sequence = tokenizer.encode(dataset[j], add_special_tokens=False)
+                full_sequence = tokenizer.encode(dataset[j])
                 if full_sequence[-1] != tokenizer.eos_token_id:
                     full_sequence.append(tokenizer.eos_token_id)
                 batch.append(full_sequence)
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -166,7 +166,7 @@ def iterate_completion_batches(
             for j in batch_idx[i]:
                 prompt, completion = dataset.get_prompt_and_completion(j)
                 prompt_lengths.append(input_length(prompt, completion, tokenizer))
-                full_sequence = tokenizer.encode(dataset[j])
+                full_sequence = tokenizer.encode(dataset[j], add_special_tokens=False)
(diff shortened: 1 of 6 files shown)


awni:
Should we close the other three PRs in favor of this one?

chimezie:
> Should we close the other three PRs in favor of this one?

Yes. I'll do that.  Just wasn't sure if they made sense being done all at once versus piecemeal

ivanfioravanti:
Yesterday I was having many issues during fine-tuning with error "apply_chat_template raise ValueError( ValueError: No chat template is set for this processor. Please either set the chat_template attribute, or provide a chat template as an argument."
Is this solving them? I will try this now.

ivanfioravanti:
chat_template was missing in tokenizer_config.json (old model), solved!

ivanfioravanti:
Is there an ETA for this PR? It's really useful to simplify training on existing HF datasets.

chimezie:
The last thing I needed to do (which is now complete) was to update how the completion mask is calculated.  Looking at DataCollatorForCompletionOnlyLM and axolotl for guidance and the latter had the most straightforward solution (see:  [#28950](https://github.com/huggingface/transformers/issues/28950)).  

I was hoping to rely on a more standard approach via the [**return_assistant_tokens_mask**](https://huggingface.co/docs/transformers/main/en/internal/tokenization_utils#transformers.PreTrainedTokenizerBase.apply_chat_template.return_assistant_tokens_mask) keyword argument of _apply_chat_template_  but this only seems to be available for chat templates that support it via the {% generation %} keyword, and it doesn't seem widely supported. 

In any case, it is ready for a review from @awni , etc.

ivanfioravanti:
Amazing job 🤩

llms/mlx_lm/tuner/trainer.py
```diff
@@ -190,7 +195,14 @@ def iterate_completion_batches(
...
 
-def iterate_batches(dataset, tokenizer, batch_size, max_seq_length, train=False):
+def iterate_batches(
+    dataset,
+    tokenizer,
+    batch_size,
+    max_seq_length,
+    train=False,
+    response_generation_tokens=None,
+):
```
chimezie, line 205: Because the custom iterate_batches method does not have the training arguments in its method signature (perhaps it should), there was no way to pass down the response token IDs to it without changing its signature in a way that doesn't break previous use.

chimezie:
> Amazing job 🤩

Thank you.

awni (approved):
Thanks a lot for the addition. It will be really nice to test completion-only fine tuning!

chimezie:
> Thanks a lot for the addition. It will be really nice to test completion-only fine tuning!

My pleasure!

PR #1117, opened 2024-11-23
Allow loading from diffusers ckpt

### LINKED ISSUE #1116: Use other flux-based model files?
Thanks for the great work!
Recently I see new models like [shuttleai/shuttle-3-diffusion · Hugging Face](https://huggingface.co/shuttleai/shuttle-3-diffusion), which is finetuned on flux-schnell. 
Is there any way to use mflux on these models?
I tried with replacing the schnell safetensor's file with shuttle 3's in ~/.cache/huggingface/hub, and ran txt2image.py. This gave me the following errors:
```text
ValueError: Received parameters not in model: model.diffusion_model.single_blocks.22.modulation.lin.bias model.diffusion_model.single_blocks.21.linear2.bias model.diffusion_model.single_blocks.5.modulation.lin.bias model.diffusion_model.double_blocks.14.img_attn.qkv.bias……
```

angeloskath:
Fixes #1116 .

The original checkpoint saved just the transformer model but in the diffusers checkpoint the transformer is nested in the `model` and `diffusion_model`. It seems that the transformer folder would be loadable as is but this small addition to `sanitize()` will work just as well.

### DIFF
--- a/flux/flux/model.py
+++ b/flux/flux/model.py
@@ -85,6 +85,8 @@ class Flux(nn.Module):
     def sanitize(self, weights):
         new_weights = {}
         for k, w in weights.items():
+            if k.startswith("model.diffusion_model."):
+                k = k[22:]
             if k.endswith(".scale"):
                 k = k[:-6] + ".weight"
             for seq in ["img_mlp", "txt_mlp", "adaLN_modulation"]:


awni (approved):
👍

PR #1118, opened 2024-11-24
Fix: Allow converting whisper models from local paths

remixer-dec:
For some reason current implementation logic in the conversion script only allows to use models from predefined list or from huggingface, but when an existing path to a model on a local drive is specified in `--torch-name-or-path` argument, it throws an error.  This error is misleading since it is thrown when a local path actually exists.


awni (approved):
Thanks for the fix!

PR #1119, opened 2024-11-24
Fix object property value in mlx_lm.server chat completions response not matching OpenAI spec

kconner:
In `mlx_lm.server`'s POST /chat/completions, the "object" property did not match OpenAI's spec.

The value produced was "chat.completions" or "chat.completions.chunk", but should be "chat.completion" or "chat.completion.chunk" for compatibility with clients expecting an OpenAI API.

Refer to:

[The chat completion object](https://platform.openai.com/docs/api-reference/chat/object)

> object string
> The object type, which is always chat.completion.

[The chat completion chunk object](https://platform.openai.com/docs/api-reference/chat/streaming)

> object string
> The object type, which is always chat.completion.chunk.

In particular, this solves a problem for me in which aider 0.64.1 reports hitting a token limit on any chat completion request to mlx_lm.server, no matter how small, despite apparently correct counts in the usage property.

### DIFF
--- a/llms/mlx_lm/SERVER.md
+++ b/llms/mlx_lm/SERVER.md
@@ -92,7 +92,7 @@ curl localhost:8080/v1/chat/completions \
 
 - `system_fingerprint`: A unique identifier for the system.
 
-- `object`: Any of "chat.completions", "chat.completions.chunk" (for
+- `object`: Any of "chat.completion", "chat.completion.chunk" (for
   streaming), or "text.completion".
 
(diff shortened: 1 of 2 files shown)


kconner:
I read the contributing guidelines a step late and will be happy to go learn how to add tests if requested!

awni (approved):
Thanks!

awni:
No need to add a test for this but please run the formatting hooks so the linter doesn't fail.

PR #1122, opened 2024-11-25
Put prompt processing in same stream

awni:
This is really an issue in MLX core with setting streams with compiled functions. But it is not trivial to fix and will require a new release, so patching it here as well:

```python
mlx_lm.generate --model mlx-community/Meta-Llama-3.1-8B-Instruct-4bit --prompt - -m 256 < prompt20.txt
```

Pre:

```
Prompt: 1929 tokens, 767.674 tokens-per-sec
Generation: 256 tokens, 41.312 tokens-per-sec
Peak memory: 5.082 GB
```

Post:

```python
Prompt: 1929 tokens, 775.695 tokens-per-sec
Generation: 256 tokens, 72.733 tokens-per-sec
Peak memory: 5.197 GB
```

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -274,13 +274,14 @@ def generate_step(
             y = sampler(logprobs)
             return y, logprobs.squeeze(0)
 
-    while y.size > prefill_step_size:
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
🙏

PR #1125, opened 2024-11-26
Accept mx.array type for prompt argument for stream_generate

neilmehta24:
This change allows mlx_lm.utils.stream_generate to accept an mx.array for its prompt argument


awni (approved):
Thanks!!

PR #1128, opened 2024-11-27
Add olmo2

### LINKED ISSUE #1127: Support for OLMo2
AllenAI has released two nice small open models OLMo2 7B and 13B with quite reasonable performances: https://allenai.org/blog/olmo2

Would it be possible to add support for them? I saw that OLMo is already in, maybe that simplifies things

awni:
closes #1127


awni:
Basically Llama architecture with:
- query and key layer norm
- post-layernorm decoder block

barronalex (approved):
Looks great!

PR #1131, opened 2024-12-01
Fix bug in FluxSampler.timesteps method

hehua2008:
According to https://github.com/black-forest-labs/flux/blob/main/demo_gr.py
```python
        timesteps = get_schedule(
            opts.num_steps,
            x.shape[-1] * x.shape[-2] // 4,
            shift=(not self.is_schnell),
        )
```
and https://github.com/black-forest-labs/flux/blob/main/src/flux/sampling.py
```python
def get_schedule(
    num_steps: int,
    image_seq_len: int,
    base_shift: float = 0.5,
    max_shift: float = 1.15,
    shift: bool = True,
) -> list[float]:
    # extra step for zero
    timesteps = torch.linspace(1, 0, num_steps + 1)

    # shifting the schedule to favor high timesteps for higher signal images
    if shift:
        # estimate mu based on linear estimation between two points
        mu = get_lin_function(y1=base_shift, y2=max_shift)(image_seq_len)
        timesteps = time_shift(mu, 1.0, timesteps)

    return timesteps.tolist()
```
and https://github.com/ml-explore/mlx-examples/blob/main/flux/flux/sampler.py
```python
    def random_timesteps(self, B, L, dtype=mx.float32, key=None):
        if self._schnell:
            # TODO: Should we upweigh 1 and 0.75?
            t = mx.random.randint(1, 5, shape=(B,), key=key)
            t = t.astype(dtype) / 4
        else:
            t = mx.random.uniform(shape=(B,), dtype=dtype, key=key)
            t = self._time_shift(L, t)
```

time_shift should be used for FLUX-Dev, not for FLUX-Schnell

### DIFF
--- a/flux/flux/sampler.py
+++ b/flux/flux/sampler.py
@@ -25,7 +25,7 @@ class FluxSampler:
     ):
         t = mx.linspace(start, stop, num_steps + 1)
 
-        if self._schnell:
+        if not self._schnell:
             t = self._time_shift(image_sequence_length, t)
 
         return t.tolist()


angeloskath (approved):
Thanks for catching that 🙏! Indeed it is a mistake!

PR #1132, opened 2024-12-02
Mixed Quantizations

barronalex:
Support passing a `quant_predicate` callable to `convert` that chooses a custom quantization for each layer in the model.

Example usage in [this gist](https://gist.github.com/barronalex/84addb8078be21969f1690c1454855f3).

## Benchmarks

Below is a comparison of the average accuracy across 8 `lm-eval` 0-shot reasoning tasks for a variety of our quants.
All experiments were conducted with Qwen2.5-7B-Instruct.

`arc_challenge`, `arc_easy`, `boolq`, `hellaswag`, `openbookqa`, `piqa`, `social_iqa`, `winogrande`

quant

The full benchmarking script is [here](https://gist.github.com/barronalex/0f816636443ca7b6821f31ef488da2a6).

A few thoughts:
- For these tasks, anything from ~5bpw up seems to be around the full fp16 performance
- Mixed quants help, particularly for 3 bit, but you're much better off using n+1 bits with a larger group size if it fits
- There's a big drop off from 4->3 bit, trained quants are on the way which will address this :)
- Llama.cpp's Q4_K_M and our mixed 4/6 bit quant end up with almost identical results (the dots are overlapping). Benchmarking the llama.cpp k quant required some custom code to dequantize the quantized scales/biases which is in this [branch](https://github.com/ml-explore/mlx-examples/tree/load-gguf) and support for 16 group size at inference in this [branch](https://github.com/ml-explore/mlx/tree/gs16). I'm not planning to merge these since it's really inefficient but it's interesting to note that the double quantization doesn't seem to help much in this setting.

### DIFF
--- a/llms/mlx_lm/tuner/utils.py
+++ b/llms/mlx_lm/tuner/utils.py
@@ -250,8 +250,9 @@ def remove_lora_layers(model: nn.Module) -> nn.Module:
 
 
 def nparams(module):
-    if isinstance(module, (nn.QuantizedLinear, nn.QuantizedEmbedding)):
-        return module.weight.size * 32 // module.bits
+    if hasattr(module, "bits"):
+        n = 0 if not hasattr(module, "bias") else module.bias.size
+        return n + module.weight.size * 32 // module.bits
     return sum(v.size for _, v in tree_flatten(module.parameters()))
 
 
--- a/llms/mlx_lm/tuner/utils.py
+++ b/llms/mlx_lm/tuner/utils.py
@@ -249,12 +249,13 @@ def remove_lora_layers(model: nn.Module) -> nn.Module:
     return model
 
 
-def print_trainable_parameters(model):
-    def nparams(m):
-        if isinstance(m, (nn.QuantizedLinear, nn.QuantizedEmbedding)):
-            return m.weight.size * (32 // m.bits)
-        return sum(v.size for _, v in tree_flatten(m.parameters()))
+def nparams(module):
+    if isinstance(module, (nn.QuantizedLinear, nn.QuantizedEmbedding)):
+        return module.weight.size * 32 // module.bits
+    return sum(v.size for _, v in tree_flatten(module.parameters()))
 
+
+def print_trainable_parameters(model):
     leaf_modules = tree_flatten(
         model.leaf_modules(), is_leaf=lambda m: isinstance(m, nn.Module)
     )
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -404,6 +404,30 @@ def load_config(model_path: Path) -> dict:
     return config
 
 
+def compute_bits_per_weight(config, weights):
+    weight_count = 0
+    bit_count = 0
+    for name, param in weights.items():
+        bit_count += param.size * param.dtype.size * 8
+        if ".scales" in name or ".biases" in name:
+            continue
+        if param.dtype == mx.uint32:
+            base_name = name.replace(".weight", "")
+            quant_config = config["quantization"]
+            quant_layer_config = quant_config.get(base_name)
+            bits = (
+                quant_config["bits"]
+                if not quant_layer_config
+                else quant_layer_config["bits"]
+            )
+            weight_count += param.size * 32 / bits
+        else:
+            weight_count += param.size
+
+    bits_per_weight = bit_count / weight_count
+    return bits_per_weight
+
+
 def load_model(
     model_path: Path,
     lazy: bool = False,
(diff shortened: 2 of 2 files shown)


llms/mlx_lm/utils.py
```diff
@@ -679,13 +690,31 @@ def quantize_model(
...
     quantized_config = copy.deepcopy(config)
-    nn.quantize(model, q_group_size, q_bits)
     quantized_config["quantization"] = {"group_size": q_group_size, "bits": q_bits}
+
+    # Add any custom quantization parameters to the config as we go
+    def _class_predicate(p, m):
+        bool_or_params = quant_predicate(p, m, config)
+        if isinstance(bool_or_params, dict):
+            quantized_config["quantization"][p] = bool_or_params
+        return bool_or_params
```
awni, line 710: Should we be storing the `True` / `False` values here as well? If we don't store them, then we can't  reconstruct quants that use `True` / `False` but the default quantization parameters from the config right?

barronalex: Great point! Just fixed it.

awni: Sorry, I may have mislead you here.  This check:

> return f"{p}.scales" in weights

should take care of the weights which are not quantized when we load. Up to you if you want to keep the explicit config for those. I can see how it can be useful.

llms/mlx_lm/utils.py
```diff
@@ -404,6 +404,30 @@ def load_config(model_path: Path) -> dict:
     return config
 
 
+def compute_bits_per_weight(config, weights):
+    weight_count = 0
+    bit_count = 0
+    for name, param in weights.items():
+        bit_count += param.size * param.dtype.size * 8
```
awni, line 411: Nit:
```suggestion
        bit_count += param.nbytes * 8
```

awni (approved):
Very nice! I left a couple comments you can pick and choose from. Feel free to merge when ready!

llms/mlx_lm/utils.py
barronalex: Nice idea! It came out much cleaner this way.
Maybe I'm misunderstanding but can we check for `to_quantized` here? That signifies that the module is _not_ quantized, but not all unquantized layers implement that method.

llms/mlx_lm/tuner/utils.py
```diff
@@ -249,12 +249,13 @@ def remove_lora_layers(model: nn.Module) -> nn.Module:
...
 
-def print_trainable_parameters(model):
-    def nparams(m):
-        if isinstance(m, (nn.QuantizedLinear, nn.QuantizedEmbedding)):
-            return m.weight.size * (32 // m.bits)
-        return sum(v.size for _, v in tree_flatten(m.parameters()))
+def nparams(module):
+    if isinstance(module, (nn.QuantizedLinear, nn.QuantizedEmbedding)):
+        return module.weight.size * 32 // module.bits
+    return sum(v.size for _, v in tree_flatten(module.parameters()))
```
awni, line 255: Do you mind making two small modifications to this:

1. Include the [quantized switch layer](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/switch_layers.py#L9) in this list.
2. Include the bias if present (I think we need that right o/w under counting params)

llms/mlx_lm/utils.py
awni: Nice, it looks great. And yea you're right.. that was a bad suggestion.

awni:
Your plot makes 4/32 look like a very attractive setting from a quality / bpw trade-off. Though I don't know how 4/32 compares to 4/64 in terms of run-time. If it follows the bpw line closely that would be very nice.

llms/mlx_lm/tuner/utils.py
barronalex: I switch it to check for the "bits" attribute. Hopefully that's not too brittle. Good call on the bias.

barronalex:
At least on my M1 Max there's a relatively small difference between b4g64 and b4g32:

4 bit 64 group size
```
Prompt: 33 tokens, 146.652 tokens-per-sec
Generation: 200 tokens, 59.179 tokens-per-sec
Peak memory: 4.326 GB
```
4 bit 32 group size
```
Prompt: 33 tokens, 143.811 tokens-per-sec
Generation: 200 tokens, 55.531 tokens-per-sec
Peak memory: 4.798 GB
```

llms/mlx_lm/tuner/utils.py
awni: SGTM! Thanks!

PR #1133, opened 2024-12-02
Allow prompt callback to `generate_step`

awni:
- Add prompt callback to generate step
- Use the callback in `cache_prompt` (simplifies some code)
- Refactor `generate_step` to accept `max_tokens`.

CC @neilmehta24, the prompt callback takes two arguments (the prompt tokens processed and the total prompt tokens) as it's more flexible. Let me know if that works for you.

### DIFF
--- a/llms/mlx_lm/cache_prompt.py
+++ b/llms/mlx_lm/cache_prompt.py
@@ -8,7 +8,7 @@ import time
 import mlx.core as mx
 
 from .models.cache import make_prompt_cache, save_prompt_cache
-from .utils import load, maybe_quantize_kv_cache
+from .utils import generate_step, load
 
 DEFAULT_QUANTIZED_KV_START = 5000
 
@@ -49,12 +49,6 @@ def setup_arg_parser():
         "--use-default-chat-template",
         action="store_true",
         help="Use the default chat template",
-    )
-    parser.add_argument(
-        "--cache-limit-gb",
-        type=int,
-        default=None,
-        help="Set the MLX cache limit in GB",
     )
     parser.add_argument(
         "--max-kv-size",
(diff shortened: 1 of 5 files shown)


neilmehta24:
That will work great. Thanks for this feature!

llms/mlx_lm/cache_prompt.py
```diff
@@ -144,26 +135,28 @@ def main():
...
         speed = processed / (current - start)
         msg = f"\rProcessed {processed:6d} tokens ({speed:6.2f} tok/s)"
+        nonlocal max_msg_len
         max_msg_len = max(max_msg_len, len(msg))
         print(msg + " " * (max_msg_len - len(msg)), end="", flush=True)
 
-        maybe_quantize_kv_cache(
-            cache, args.quantized_kv_start, args.kv_group_size, args.kv_bits
-        )
+    for _ in generate_step(
```
angeloskath, line 149: This does an unnecessary sampling but I like very much that it removes so much duplicated logic.

angeloskath (approved):
Very nice!

I would change the `n < max_tokens` to `n != max_tokens` and document the negative `max_tokens` and I think it is golden :-)

llms/mlx_lm/cache_prompt.py
awni: Right, it does the async eval on the first token / log probs. It should be negligible compared to the rest of the computation..

It's nice it simplifies the logic a bit.. mostly I changed this to use the callback and verify it was working correctly.

Another option I considered is to split out a `prefill_prompt` API which `generate_step` can use and which `cache_prompt.py` can use. Maybe it's better, but so far we don't need to use it anywhere except here. If we end up needing to do the prompt computation in more places I will play around with that.

llms/mlx_lm/utils.py
```diff
@@ -286,20 +296,26 @@ def _step(y):
     mx.async_eval(y, logprobs)
     n = 0
     while True:
-        next_y, next_logprobs = _step(y)
-        mx.async_eval(next_y, next_logprobs)
+        if n < max_tokens:
```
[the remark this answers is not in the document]
awni: Nice suggestion, thanks!

awni:
Also closed https://github.com/ml-explore/mlx-examples/issues/1134

PR #1137, opened 2024-12-05
Chang Flux's default max_shift to 1.15 to match the official one

hehua2008:
IMPORTANT!!!
This is a very important and magic value!!!
With the right value 1.15, we can train LoRA very much faster than ever before!!!

See in https://github.com/black-forest-labs/flux/blob/main/src/flux/sampling.py

```python
def get_lin_function(
    x1: float = 256, y1: float = 0.5, x2: float = 4096, y2: float = 1.15
) -> Callable[[float], float]:
    m = (y2 - y1) / (x2 - x1)
    b = y1 - m * x1
    return lambda x: m * x + b


def get_schedule(
    num_steps: int,
    image_seq_len: int,
    base_shift: float = 0.5,
    max_shift: float = 1.15,
    shift: bool = True,
) -> list[float]:
    # extra step for zero
    timesteps = torch.linspace(1, 0, num_steps + 1)

    # shifting the schedule to favor high timesteps for higher signal images
    if shift:
        # estimate mu based on linear estimation between two points
        mu = get_lin_function(y1=base_shift, y2=max_shift)(image_seq_len)
        timesteps = time_shift(mu, 1.0, timesteps)

    return timesteps.tolist()
```

### DIFF
--- a/flux/flux/sampler.py
+++ b/flux/flux/sampler.py
@@ -7,7 +7,7 @@ import mlx.core as mx
 
 
 class FluxSampler:
-    def __init__(self, name: str, base_shift: float = 0.5, max_shift: float = 1.5):
+    def __init__(self, name: str, base_shift: float = 0.5, max_shift: float = 1.15):
         self._base_shift = base_shift
         self._max_shift = max_shift
         self._schnell = "schnell" in name


hehua2008:
@angeloskath Be sure to merge this commit and then you'll find the magic!

angeloskath:
Thanks for finding the bug! I will test it and merge during the weekend!

angeloskath (approved):
Thanks for catching that!

It doesn't make a big difference for me but since 1.15 is obviously the correct value we should switch.

PR #1143, opened 2024-12-07
fix missing --final-adapter-file crash in dreambooth.py

petersibley:
This fixes a crash I spotted at the very end of the `dreambooth.py` example in the `flux` directory.

The code assumed a command line argument was passed in but then always saved to `final_adapters.safetensors`  I think this addresses the issue.


angeloskath (approved):
Thanks for the fix! I opted to just remove the message as `save_adapters()` already informs us where the adapters were saved.

PR #1144, opened 2024-12-09
[mlx-lm] Use top p in server

awni:
This was accidentally dropped during a previous refactor I did.


chimezie (approved):
Looks good

awni:
Also added a fix to make sure we quantize the cache during the prefill stage.

angeloskath (approved):
👍

PR #1146, opened 2024-12-09
Replace unicode errors instead of raising exception

### LINKED ISSUE #1142: BUG: mlx-lm encounters utf decode error
When I used mlx-lm stream_generate, a utf decode error may happen, It's not 100% happened, when I ran it for about 10 times, and the error showed.

```
In [10]: from mlx_lm import load, stream_generate
    ...: 
    ...: repo = "/Users/xuyeqin/.xinference/cache/qwen2_5-instruct-mlx-7b-4bit"
    ...: model, tokenizer = load(repo)
    ...: 
    ...: prompt = "请出十道关于唐朝的选择题，并给出正确答案"
    ...: 
    ...: messages = [{"role": "user", "content": prompt}]
    ...: prompt = tokenizer.apply_chat_template(
    ...:     messages, tokenize=False, add_generation_prompt=True
    ...: )
    ...: 
    ...: for response in stream_generate(model, tokenizer, prompt, max_tokens=51
    ...: 2,temp=1.0, repetition_context_size=20, top_p=1.0):
    ...:     print(response.text, end="", flush=True)
    ...: print()
[Warning] Specifying sampling arguments to ``generate_step`` is deprecated. Pass in a ``sampler`` instead.
当然可以，以下是为唐朝设计的10道选择题，包含了唐朝的历史、文化、科技等方面的知识。题目和答案仅供参考。

1. 唐朝是在哪一年开始的？
   - A) 618年 
   - B) 581年 
   - C) 681年 
   - D) 711年 
   正确答案：A

2. 下列哪位诗人不是唐朝的著名诗人？
   - A) 杜甫 
   - B) 苏轼 
   - C) 李白 
   - D) 白居易 
   正确答案：B

3. 唐朝时期发明了什么？
   - A) 造纸术 
   - B) 火药 
   - C) 指南针 
   - D) 活字印刷术 
   正确答案：B

4. 哪位唐朝帝王被称为“唐太宗”，综治之才？
   - A) 李世民 
   - B) 李隆基 
   - C) 李渊 
   - D) 李治 
   正确答案：A

5. 下列哪一事件不属于唐朝初期的重要事件？
   - A) 长安的建设（即大兴城） 
   - B) 开元盛世 
   - C) 设置安西都护府 
   - D) 平定内乱，首次统一中国 
   正确答案：B （开元盛世是唐朝的繁荣时期，但并不是初期的事件）

6. 在唐朝的宗教信仰方面，佛教的地位如何？
   - A) 继续是唯一信仰 
   - B) 比较受欢迎 
   - C) 被王权和世俗势力所打压 
   - D) 只在朝中官员中流行 
   正确答案：B

7. 唐朝时期在哪个城市设立的第一个市舶司？
   - A) 长安 
   - B) 扬州 
   - C) 江南市 
   - D) 广州 
   正确答案：D

8. 下列哪位皇帝被认为开创了唐朝著名的“贞观之治”？
   - A) 唐太宗李世民 
   ----------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
Cell In[10], line 13
      8 messages = [{"role": "user", "content": prompt}]
      9 prompt = tokenizer.apply_chat_template(
     10     messages, tokenize=False, add_generation_prompt=True
     11 )
---> 13 for response in stream_generate(model, tokenizer, prompt, max_tokens=512,temp=1.0, repetition_context_size=20, top_p=1.0):
     14     print(response.text, end="", flush=True)
     15 print()

File ~/miniconda3/lib/python3.11/site-packages/mlx_lm/utils.py:356, in stream_generate(model, tokenizer, prompt, max_tokens, **kwargs)
    343         break
    345     yield GenerationResponse(
    346         text=detokenizer.last_segment,
    347         token=token,
   (...)
    353         peak_memory=mx.metal.get_peak_memory() / 1e9,
    354     )
--> 356 detokenizer.finalize()
    357 yield GenerationResponse(
    358     text=detokenizer.last_segment,
    359     token=token,
   (...)
    365     peak_memory=mx.metal.get_peak_memory() / 1e9,
    366 )

File ~/miniconda3/lib/python3.11/site-packages/mlx_lm/tokenizer_utils.py:216, in BPEStreamingDetokenizer.finalize(self)
    215 def finalize(self):
--> 216     current_text = bytearray(self._byte_decoder[c] for c in self._unflushed).decode(
    217         "utf-8"
    218     )
    219     self.text += self._maybe_trim_space(current_text)
    220     self._unflushed = ""

UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 1-2: unexpected end of data
```

angeloskath:
Fixes #1142 . I chose `replace` but `ignore` may be an even better option as this is only likely to happen at the end when we have predicted half a multibyte unicode character.

### DIFF
--- a/llms/mlx_lm/tokenizer_utils.py
+++ b/llms/mlx_lm/tokenizer_utils.py
@@ -132,7 +132,7 @@ class SPMStreamingDetokenizer(StreamingDetokenizer):
         self.tokens = []
 
     def _flush(self):
-        text = self._unflushed.replace(self._sep, b" ").decode("utf-8")
+        text = self._unflushed.replace(self._sep, b" ").decode("utf-8", "replace")
         if not self.text and self.trim_space and text and text[0] == " ":
             text = text[1:]
         self.text += text
@@ -202,7 +202,7 @@ class BPEStreamingDetokenizer(StreamingDetokenizer):
         if is_added or self._byte_decoder[v[0]] == 32:
             current_text = bytearray(
                 self._byte_decoder[c] for c in self._unflushed
-            ).decode("utf-8")
+            ).decode("utf-8", "replace")
             self.text += self._maybe_trim_space(current_text)
             if is_added:
                 self.text += v
@@ -214,7 +214,8 @@ class BPEStreamingDetokenizer(StreamingDetokenizer):
 
     def finalize(self):
         current_text = bytearray(self._byte_decoder[c] for c in self._unflushed).decode(
-            "utf-8"
+            "utf-8",
+            "replace",
         )
         self.text += self._maybe_trim_space(current_text)
         self._unflushed = ""


awni:
Oh that's a nice solution! I didn't know that was configurable.

 In this case, replace will replace with `U+FFFD` and then we [ignore the entire last segment](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/tokenizer_utils.py#L54)? Maybe we should use `ignore` or alternatively fix that `last_segment` logic to just pop the replacement chracter?

angeloskath:
Ah good point. Why are we skipping segments that end in replacement char 🤔?

awni:
> Why are we skipping segments that end in replacement char 🤔

Yes that logic seems pretty suspect. I think it was inherited from when we used to decode at every new token 🤷‍♂️ ..

qinxuye:
`replace` seems OK, we encounter `U+FFFD` at the end when streaming from qwen series model, and we just chose to strip it in our framework, ignoring the last segment is not a good choice IMHO.

qinxuye:
> `replace` seems OK, we encounter `U+FFFD` at the end when streaming from qwen series model(this is common not only for MLX), and we just chose to strip it in our framework, ignoring the last segment is not a good choice IMHO.

awni (approved):
Thank you!

awni:
Huh not sure what broke the tests 🤔

angeloskath:
Yeah me neither, let me fix it real quick and then I can merge.

PR #1154, opened 2024-12-12
Bpe stream without space

### LINKED ISSUE #1136: The problem of stream_generate() method returning Chinese
When I call the big model and use stream_generate() to return the Chinese result, the effect is the same as the result returned by generate(). It only returns greetings generated from all texts, and the waiting time is relatively long; If the return result is in English, stream_generate() returns the streaming result.

### Here are my code calls and result processing:

```
streamer = stream_generate(self.model,self.tokenizer,prompt=chat_template,max_tokens=generation_config.max_new_tokens,sampler=sampler,prompt_cache=self.prompt_cache)

for response in streamer:
    print(response.text, end="", flush=True)
```
### Why is that?

awni:
Streaming BPE detokenization for character sequences which don't have spaces

Closes #1136

### DIFF
--- a/llms/mlx_lm/_version.py
+++ b/llms/mlx_lm/_version.py
@@ -1,3 +1,3 @@
 # Copyright © 2023-2024 Apple Inc.
 
-__version__ = "0.20.2"
+__version__ = "0.20.4"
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Makes sense. Unless we are breaking a character add it.

PR #1157, opened 2024-12-14
Add support for cohere2

Blaizzy:
Adds support for Cohere2 with sliding attention.

Thanks a lot to @N8python for the inspiration!

Bf16
Screenshot 2024-12-14 at 4 53 12 PM

4bit
Screenshot 2024-12-14 at 5 10 12 PM

### DIFF
--- a/llms/mlx_lm/models/cohere2.py
+++ b/llms/mlx_lm/models/cohere2.py
@@ -1,0 +1,207 @@
+# Copyright © 2023-2024 Apple Inc.
+
+from dataclasses import dataclass
+from typing import Optional, Tuple
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .base import BaseModelArgs, create_attention_mask
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    model_type: str
+    hidden_size: int = 4096
+    head_dim: int = 128
+    num_hidden_layers: int = 32
+    intermediate_size: int = 14336
+    num_attention_heads: int = 32
+    num_key_value_heads: int = 8
+    rope_theta: float = 50000.0
+    vocab_size: int = 256000
+    layer_norm_eps: float = 1e-05
+    logit_scale: float = 0.0625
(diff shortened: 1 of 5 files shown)


Blaizzy:
Thanks @N8python!

Verified this change saves ~2GB for 4bit:

Screenshot 2024-12-14 at 11 35 09 PM

and ~1.2-1.5GB for bf16:
Screenshot 2024-12-14 at 10 20 56 PM

llms/mlx_lm/models/cohere2.py
```diff
@@ -0,0 +1,215 @@
...
+                queries = self.rope(queries, offset=cache.offset)
+                keys = self.rope(keys, offset=cache.offset)
+
+        if cache is not None:
+            keys, values = cache.update_and_fetch(keys, values)
+
+        if self.use_sliding_window and mask is not None:
+            key_len = keys.shape[-2]
+            if mask.shape[-1] != key_len:
+                mask = mask[..., -key_len:]
```
awni, line 89: This changed slightly. You would be over trimming the keys/values during the prefill stage otherwise.

Blaizzy:
My pleasure!

llms/mlx_lm/models/cohere2.py
Blaizzy: Oh, I see.

I thought of this but it was giving me a shape error when I tried exactly this. Because I knew the make_cache was already handling the kv slicing when I checked the shapes.

...
===
window_size 4096
keys shape (1, 8, 4096, 128)
values shape (1, 8, 4096, 128)
mask shape after (512, 4096)
===
window_size 4608
keys shape (1, 8, 4608, 128)
values shape (1, 8, 4608, 128)
mask shape after (512, 4608)
===
window_size 4608
keys shape (1, 8, 4608, 128)
values shape (1, 8, 4608, 128)
mask shape after (512, 4608)
===
...
It seems like I should added the changes in mask (L158-163) that you added.

PR #1174, opened 2024-12-19
Distributed `mlx_lm.evaluate`

barronalex:
Add a distributed version of `mlx_lm.evaluate` that runs on multiple nodes and produces identical outputs.

Also fix a few bugs:
- Add masking so that changing the `batch_size` no longer affects the output 
- Fixed a bug in `loglikelihood_rolling` tasks, e.g. wiki text

```
mlx_lm.evaluate --model mlx-community/Qwen2.5-7B-Instruct-bf16 --tasks winogrande
```

On 1 M2 Ultra:
```
Acc:   0.6985003946329913
Time: 16 sec 
```
On 4 M2 Ultra:

### DIFF
--- a/llms/mlx_lm/evaluate.py
+++ b/llms/mlx_lm/evaluate.py
@@ -16,10 +16,9 @@ from lm_eval.api.model import LM
 from lm_eval.api.registry import register_model
 from tqdm import tqdm
 
+from .models.base import create_causal_mask
 from .models.cache import make_prompt_cache
 from .utils import load, stream_generate
-
-PAD = 0
 
 
 def _len_longest_common_prefix(a, b):
@@ -39,31 +38,14 @@ def _rstrip_until(s, untils):
     return s[: min(f)]
 
 
-def _pad_inputs(
-    inputs,
-    maxlen,
-    genlen=0,
-    pad_left=False,
-    pad_multiple=32,
-    truncate=False,
-):
-    # pad the prompts to the left with at least genlen tokens.
-    actual_maxlen = max(len(p) for p in inputs) + genlen
-    if actual_maxlen > maxlen:
-        if not truncate:
-            raise ValueError("Inputs are too long.")
-        else:  # drop begining
-            actual_maxlen = maxlen
-            inputs = [p[max(0, len(p) - maxlen) :] for p in inputs]
-    if pad_multiple > 0:
-        maxlen = (actual_maxlen + pad_multiple - 1) // pad_multiple
-        maxlen *= pad_multiple
-    assert PAD == 0
-    lr = np.array((1, 0) if pad_left else (0, 1))
-    return np.stack(
-        [np.pad(np.array(x, np.int32), lr * (maxlen - len(x))) for x in inputs],
+def _pad_inputs(inputs):
+    lengths = mx.array([len(x) for x in inputs])
+    maxlen = lengths.max()
+    padded = mx.stack(
+        [mx.pad(mx.array(x), (0, maxlen - len(x))) for x in inputs],
         axis=0,
     )
+    return padded, lengths
 
 
 @register_model("mlxlm")
@@ -79,28 +61,31 @@ class MLXLM(LM):
         self._model, self._tokenizer = load(path_or_hf_repo)
         self._max_tokens = max_tokens or self._tokenizer.model_max_length
 
-    def _score_fn(self, inputs, tokenize=True, step_size=32):
-        if tokenize:
-            inputs = self._tokenizer.encode(inputs)
-        inputs = _pad_inputs(inputs, self._max_tokens, truncate=False)
-        inputs = mx.array(inputs)
+    def _score_fn(self, inputs, step_size=64):
+        inputs, lengths = _pad_inputs(inputs)
         inputs, targets = inputs[..., :-1], inputs[..., 1:]
 
         cache = make_prompt_cache(self._model)
 
-        mask = targets != PAD
+        # TODO: come up with a better way to get the dtype
+        dtype = self._model.model.embed_tokens(inputs).dtype
 
         scores, is_greedy = [], []
         for i in range(0, inputs.shape[1], step_size):
-            logits = self._model(inputs[:, i : i + step_size], cache=cache)
+            inp = inputs[:, i : i + step_size]
+            T = inp.shape[1]
 
+            offset = cache[0].offset
+            mask = create_causal_mask(T, offset, lengths=lengths).astype(dtype)
+
+            logits = self._model(inp, cache=cache, mask=mask)
             log_probs = nn.log_softmax(logits.astype(mx.float32))
+
             score = mx.take_along_axis(
                 log_probs, targets[:, i : i + step_size, mx.newaxis], axis=-1
             )[..., 0]
-            ig = mask[:, i : i + step_size] * (
-                targets[:, i : i + step_size] == mx.argmax(logits, axis=-1)
-            )
+            ig = targets[:, i : i + step_size] == mx.argmax(logits, axis=-1)
+            ig = mx.where(mx.arange(T) + offset < lengths[:, None], ig, False)
 
             mx.eval(score, ig)
             mx.metal.clear_cache()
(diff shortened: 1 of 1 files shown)


ivanfioravanti:
This is great! I'm testing it with M2 Ultra + 2 M4 Max. WOW! Great job @barronalex 
When will this be reviewed and merged?

llms/mlx_lm/evaluate.py
```diff
@@ -39,31 +38,14 @@ def _rstrip_until(s, untils):
...
-    if pad_multiple > 0:
-        maxlen = (actual_maxlen + pad_multiple - 1) // pad_multiple
-        maxlen *= pad_multiple
-    assert PAD == 0
-    lr = np.array((1, 0) if pad_left else (0, 1))
-    return np.stack(
-        [np.pad(np.array(x, np.int32), lr * (maxlen - len(x))) for x in inputs],
+def _pad_inputs(inputs):
+    lengths = mx.array([len(x) for x in inputs])
+    maxlen = lengths.max()
```
awni, line 43: I would avoid doing small computations like this which require just a couple kernel launches + a graph eval in MLX. It will be like 100x (or more) faster to do:

```python
maxlen = max(len(x) for x in inputs)
```

llms/mlx_lm/evaluate.py
```diff
@@ -79,28 +61,31 @@ def __init__(
...
+        dtype = self._model.model.embed_tokens(inputs).dtype
 
         scores, is_greedy = [], []
         for i in range(0, inputs.shape[1], step_size):
-            logits = self._model(inputs[:, i : i + step_size], cache=cache)
+            inp = inputs[:, i : i + step_size]
+            T = inp.shape[1]
+
+            offset = cache[0].offset
+            mask = create_causal_mask(T, offset, lengths=lengths).astype(dtype)
```
awni, line 79: I think a nicer solution might be to allow boolean masks in our SDPA. It's a really easy mistake to make the mask the wrong type and get inadvertent up-casting. If you send a bool instead it will always cast to the right type. (Just making a comment for later). It will also use a lot less memory for large contexts.

llms/mlx_lm/evaluate.py
```diff
@@ -39,31 +38,14 @@ def _rstrip_until(s, untils):
...
-        maxlen *= pad_multiple
-    assert PAD == 0
-    lr = np.array((1, 0) if pad_left else (0, 1))
-    return np.stack(
-        [np.pad(np.array(x, np.int32), lr * (maxlen - len(x))) for x in inputs],
+def _pad_inputs(inputs):
+    lengths = mx.array([len(x) for x in inputs])
+    maxlen = lengths.max()
+    padded = mx.stack(
+        [mx.pad(mx.array(x), (0, maxlen - len(x))) for x in inputs],
```
awni, line 45: I'd probably also do this in Python.. but it's minor since `padded` goes into the main graph.

```suggestion
    padded = mx.array([x + [0] * (maxlen - len(x)) for x in inputs])
```

llms/mlx_lm/evaluate.py
```diff
@@ -111,37 +96,26 @@ def _score_fn(self, inputs, tokenize=True, step_size=32):
...
+            scores, length, is_greedy = self._score_fn(batch)
             for j in range(len(batch)):
-                if sorted_spans is None:  # full sequence score
-                    mask = mx.arange(scores[j].shape[-1]) < length
-                    score = (scores[j].astype(mx.float32) * mask).sum(axis=-1)
-                    ig = (is_greedy[j].astype(mx.int32) * mask).sum(axis=-1)
+                if score_spans is None:  # full sequence score
+                    l = length[j].item()
+                    score = scores[j][:l].astype(mx.float32).sum()
+                    ig = is_greedy[j][:l].astype(mx.int32).sum()
```
awni, line 110: It's not really necessary to cast this here since `mx.array([False, True]).sum()` has type `mx.int32`

llms/mlx_lm/evaluate.py
```diff
@@ -111,37 +96,26 @@ def _score_fn(self, inputs, tokenize=True, step_size=32):
...
                 else:  # subsequence score
-                    start, end = sorted_spans[i + j]
+                    start, end = score_spans[i + j]
                     score = scores[j][start:end].astype(mx.float32).sum()
                     ig = is_greedy[j][start:end].astype(mx.int32).sum()
```
awni, line 114: Same for the cast to `int32` here.

llms/mlx_lm/evaluate.py
```diff
@@ -111,37 +96,26 @@ def _score_fn(self, inputs, tokenize=True, step_size=32):
...
 
+    def _loglikelihood(self, texts, score_spans=None):
         results = []
-        for i in tqdm(range(0, len(sorted_inputs), self._batch_size)):
-            batch = sorted_inputs[i : i + self._batch_size]
-            scores, length, is_greedy = self._score_fn(batch, tokenize=tokenize)
+        for i in tqdm(range(0, len(texts), self._batch_size)):
+            batch = texts[i : i + self._batch_size]
+            scores, length, is_greedy = self._score_fn(batch)
             for j in range(len(batch)):
```
awni, line 106: It's probably over optimization.. but having a bunch of evals in a loop like this is not such a good pattern. It hits latency really hard. Especially things like:

```
for i in range(len(lengths)):
  l = lengths[i].item()
```

Which does a kernel launch for the scatter + full GPU synch at each iteration. 

I know it probably will make little to no difference in runtime since this isn't the bottleneck, but perhaps still good to change it to set a good example.

awni: One way to do it would be:
- have a mask making function which takes `start_offset = mx.array, end_offset = mx.array` and makes the mask.
-  Then multiply `scores` and `is_greedy` by the mask. 
- Then sum along the time axis.
-  Then eval everything in one shot `mx.eval(scores, is_greedy)`
- Then convert them all to lists, zip and return

awni: Although it looks like keeping them as `mx.array` could be useful rather than converting back and forth here and at L205

llms/mlx_lm/evaluate.py
```diff
@@ -209,13 +183,45 @@ def loglikelihood(self, requests) -> list[tuple[float, bool]]:
...
+        # sort by length to get batches with little padding.
+        sorted_indices = sorted(range(len(shortened)), key=lambda i: -len(shortened[i]))
+        shortened = [shortened[i] for i in sorted_indices]
+        completion_spans = [completion_spans[i] for i in sorted_indices]
+
+        group = mx.distributed.init() if mx.distributed.is_available() else None
+        if group is not None:
+            # split strided so we have approximately the same lengths on each node
+            shortened = shortened[group.rank() :: group.size()]
+            completion_spans = completion_spans[group.rank() :: group.size()]
```
awni, line 197: You don't necessarily need to check `is_available()`. If it's not available then `rank = 0` and `size = 1` and everything should work fine. So you can avoid needing to condition code on `group is not None`.

ivanfioravanti:
Any news on this PR? It would be great to speed up some distributed evals on DeepSeek R1 😜

barronalex:
That’s awesome! I’ll get it in later today.

barronalex:
@awni thanks for the comments! I think this is good to merge now.

llms/mlx_lm/evaluate.py
```diff
@@ -346,11 +361,8 @@ def main():
     )
     parser.add_argument(
         "--apply-chat-template",
```
barronalex, line 363: It was impossible to disable this before, so I've changed it to be off by default (which mirrors the `lm_eval` behavior)

awni: I'm not sure about defaulting it to off for instruct models.. it seems like you would always want this on for most models that are used regularly? Does it make sense to change this to `--ignore-chat-template` instead to be able to shut it off if needed?

awni (approved):
Looks great! Just one comment. Let me know what you think. Otherwise LGTM!

PR #1176, opened 2024-12-20
Change the eos-token argument for `mlx_lm.generate`

angeloskath:
Currently `--eos-token` is only used if the `eos_token_ids` key is not set in the model config. As a result, most times it is ignored and it can't be used to neither add nor override any eos token.

This PR removes that argument and adds `--extra-eos-token` which allows us to set any number of tokens and token ids in the `eos_token_ids` set. The only behavior that is changed is for the generate script.


awni (approved):
Looks great, thanks!

Issue #1177, opened 2024-12-21
mlx_lm.manage wrong default


ivanfioravanti:
Default for deletion is Y, but in case of empty string it is N and no deletion occurs.

awni:
This was closed in #1178

PR #1178, opened 2024-12-21
Improvements to mlx_lm.manage

ivanfioravanti:
Default is now N and size added to deletion confirmation screen.
Fix for #1177

### DIFF
--- a/llms/mlx_lm/manage.py
+++ b/llms/mlx_lm/manage.py
@@ -6,10 +6,12 @@ from transformers.commands.user import tabulate
 
 
 def ask_for_confirmation(message: str) -> bool:
+    """Ask user for confirmation with Y/N prompt.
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/manage.py
```diff
@@ -18,7 +20,7 @@ def ask_for_confirmation(message: str) -> bool:
             return True
         if answer in n:
             return False
-        print(f"Invalid input. Must be one of {all_values}")
+        print(f"Invalid input. Must be one of: yes/no/y/N or empty for no")
```
awni, line 23: Is it intentional that the `y` lower case but the `N` upper case? Could we make them the same?

ivanfioravanti: You are right. Fixed.

awni (approved):
Thanks!

PR #1180, opened 2024-12-23
Add support for fewshot and apply chat template lm_eval functionality

chimezie:
Adds `--fewshot-as-multiturn` and `--apply-chat-template` to script, which correspond to lm_eval options of similar names and very often used to ensure apples-to-apples comparisons of lm_evaluation results


awni:
CC @barronalex this PR adds the chat template functions we were discussing.

PR #1189, opened 2025-01-03
Fix encoding with special tokens + chat template

### LINKED ISSUE #1188: mlx_lm.generate adds duplicated `<|begin_of_text|>` token
Hi, one `begin_of_text` token is added here

https://github.com/ml-explore/mlx-examples/blob/3a58c361096e5be7a927e7719c5ef66bace9a8ab/llms/mlx_lm/generate.py#L202

and a second one here

https://github.com/ml-explore/mlx-examples/blob/3a58c361096e5be7a927e7719c5ef66bace9a8ab/llms/mlx_lm/utils.py#L357

This can produce small differences in the generated text

awni:
There were a few cases where the way we use chat templates was broken:

For some (but not all) models with a chat template if you apply the chat template without tokenization and then encode to tokens you get duplicate bos tokens. After applying a chat template one should always use `add_special_tokens=False` when encoding to token ids. 

Since we sometimes needs `add_special_tokens=True` (e.g. for base models w/out a chat template) and it's difficult to tell in the general case when we need to add them for a given string prompt, the cleanest solution I found is to just encode prompts to tokens before passing to generate / other places.

Also added a few lines in `stream_generate` to attempt to autodetect the setting of `add_special_tokens` when given a string prompt, but there are edge cases that can sneak through, so it's preferable to encode prior to calling generate.

Also fixed an issue in LoRA where we add the generation prompt to certain datasets which is not correct.

Closes #1188

### DIFF
--- a/llms/README.md
+++ b/llms/README.md
@@ -58,7 +58,7 @@ prompt = "Write a story about Einstein"
 
 messages = [{"role": "user", "content": prompt}]
 prompt = tokenizer.apply_chat_template(
-    messages, tokenize=False, add_generation_prompt=True
+    messages, add_generation_prompt=True
 )
 
 text = generate(model, tokenizer, prompt=prompt, verbose=True)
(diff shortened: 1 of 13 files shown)


barronalex (approved):
Looks great!

PR #1193, opened 2025-01-07
Use upload_large_folder

pcuenca:
There are in my opinion two reasons to use `upload_large_folder` instead of `upload_folder`:

- The `multi_commits` args are no longer supported by `upload_folder`. Attempting to convert and upload a model with a recent version of `huggingface_hub` will fail. This includes a new install of `mlx-lm` in an empty environment.
- `upload_large_folder` is more efficient and resilient than `upload_folder`. There are also [some limitations](https://huggingface.co/docs/huggingface_hub/v0.27.1/en/package_reference/hf_api#huggingface_hub.HfApi.upload_large_folder), but I think they don't apply here.

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -673,12 +673,10 @@ def upload_to_hub(path: str, upload_repo: str, hf_path: str):
 
     api = HfApi()
     api.create_repo(repo_id=upload_repo, exist_ok=True)
-    api.upload_folder(
+    api.upload_large_folder(
         folder_path=path,
         repo_id=upload_repo,
         repo_type="model",
-        multi_commits=True,
-        multi_commits_verbose=True,
     )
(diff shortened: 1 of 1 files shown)


awni:
Indeed, I had already changed it in https://github.com/ml-explore/mlx-examples/pull/1191 but since that won't land for a few days, let's land this one. I assume we should pin a higher version of transformers [in ](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/requirements.txt#L3) ?

pcuenca:
Ah, I missed your change in #1191, sorry. I don't think it's necessary to pin a higher version of transformers, the removal of the args was done in huggingface-hub 0.27.0, transformers [requires 0.24 at least](https://github.com/huggingface/transformers/blob/7f7677307cfc8722ccc4680b993811098379844c/setup.py#L120), so an install in an empty environment would pull the latest version of huggingface-hub.

PR #1194, opened 2025-01-07
Support snapshot_download for ModelScope 

wangxingjun778:
1. Refer to www.modelscope.cn
2. Specify the `model_id` to download from the Modelscope Hub
3. Speed up the download process especially for CN users

### DIFF
--- a/llms/mlx_lm/requirements.txt
+++ b/llms/mlx_lm/requirements.txt
@@ -1,4 +1,4 @@
-mlx>=0.19.2
+mlx>=0.22.0
 numpy
 transformers[sentencepiece]>=4.39.3
 protobuf
--- a/llms/mlx_lm/requirements.txt
+++ b/llms/mlx_lm/requirements.txt
@@ -4,4 +4,3 @@ transformers[sentencepiece]>=4.39.3
 protobuf
 pyyaml
 jinja2
-modelscope
--- a/llms/mlx_lm/requirements.txt
+++ b/llms/mlx_lm/requirements.txt
@@ -4,3 +4,4 @@ transformers[sentencepiece]>=4.39.3
 protobuf
 pyyaml
 jinja2
+modelscope
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -353,9 +353,13 @@ def stream_generate(
         tokenizer = TokenizerWrapper(tokenizer)
 
     if not isinstance(prompt, mx.array):
-        prompt = mx.array(
-            prompt if isinstance(prompt, list) else tokenizer.encode(prompt)
-        )
+        if isinstance(prompt, str):
+            # Try to infer if special tokens are needed
+            add_special_tokens = tokenizer.bos_token is None or not prompt.startswith(
+                tokenizer.bos_token
+            )
+            prompt = tokenizer.encode(prompt, add_special_tokens=add_special_tokens)
+        prompt = mx.array(prompt)
 
     detokenizer = tokenizer.detokenizer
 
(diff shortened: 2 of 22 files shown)


llms/mlx_lm/utils.py
```diff
@@ -153,22 +160,39 @@ def get_model_path(path_or_hf_repo: str, revision: Optional[str] = None) -> Path
...
+                        allow_patterns=[
+                            "*.json",
+                            "*.safetensors",
+                            "*.py",
+                            "tokenizer.model",
+                            "*.tiktoken",
+                            "*.txt",
+                        ],
+                    )
                 )
```
awni, line 195:
```suggestion
            if use_modelscope:
                revision = revision or "master"
            model_path = Path(
                snapshot_download(
                    repo_id=path_or_hf_repo,
                    revision=revision,
                    allow_patterns=[
                        "*.json",
                        "*.safetensors",
                        "*.py",
                        "tokenizer.model",
                        "*.tiktoken",
                        "*.txt",
                    ],
                )
            )
```

llms/mlx_lm/utils.py
```diff
@@ -15,7 +16,13 @@
 
 import mlx.core as mx
 import mlx.nn as nn
-from huggingface_hub import snapshot_download
+
+use_modelscope = os.getenv('MLX_USE_MODELSCOPE', 'False').lower() == 'true'
```
awni, line 20: I'm debating about this as an environment variable 🤔 .. an alternative is to put the provider in the path ilke:

```
modelscope:reponame
```

That would also let us distinguish local vs hf in cases where it's ambiguous.

yingdachen: we are open to discussions. here is our take:

1. currently local vs hf-repo is not distinguished with a provider prefix, local is always preferred, we will only look in hf when the no local path exist.
2. requiring adding a prefix for hf, such as 'hf:reponame', may not seem a good idea, since it is a breaking change and will require updating all sample codes on hf model cards.
3. using environment variable allows the identical code to be used for both hf and modelscope, user only needs to set up an environment variable once, across multiple model usage. this is more aligned with why users may opt for modelscope: for some reason, they cannot access hf in their environment. 

again, we can take either approach preferred by mlx project, but we think it is worthwhile to share our thoughts on why we chose the environment variable approach in the first place :)

btw, there is also a mlx-community org on modelscope: https://www.modelscope.com/organization/mlx-community, which provides gateway for mlx model usage where hf is not accessible.  all community members  are welcome to join.

awni: Thanks, I would definitely not want to break back-compatability. So I was thinking the search process could be:

1. Check if there is a prefix (e.g. `modelscope:`, `local:`, `hf:`) and prefer that.
2. Otherwise fallback to the default search behavior.

But I think what you have now is also fine for the time being and simpler from an implementation standpoint so let's leave it.

awni:
Would you mind addressing the other comments, then we can merge this?

wangxingjun778:
Has been tested in the following envs:
  * MBP M4
  * macOS  15.1.1
 
Example:
```python

model, tokenizer = load("Qwen/Qwen2.5-0.5B-Instruct")

prompt = "Write a story about Einstein"

messages = [{"role": "user", "content": prompt}]
prompt = tokenizer.apply_chat_template(
    messages, add_generation_prompt=True
)

text = generate(model, tokenizer, prompt=prompt, verbose=True)
```

Run and get results:
image

wangxingjun778:
If there are no further comments, we can merge this.  :)

llms/mlx_lm/requirements.txt
```diff
@@ -4,3 +4,4 @@ transformers[sentencepiece]>=4.39.3
 protobuf
 pyyaml
 jinja2
+modelscope
```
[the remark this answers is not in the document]
wangxingjun778: remove modelscope

awni (approved):
Thanks for the addition!

PR #1197, opened 2025-01-09
Export and run LLMs in C++

awni:
Comments:

- So far only tested and working with `model_type="llama"`
- Requires MLX PR https://github.com/ml-explore/mlx/pull/1722
- C++ does tokenization and streaming detokenization but no chat templates 
- C++ is same speed as Python.
- Compilation helps both C++ and Python. On M1 Max it's about 0.7 toks/sec for 4-bit LLama 3 8B. Toks/sec goes from 61.5 to 62.2.


llms/export/tokenizer.cpp
```diff
@@ -0,0 +1,189 @@
...
+  ids.push_back(bos_id_);
+
+  // Initialize merges to integer list
+  auto merge_segment = [&ids, &one_step_merge, this](const std::string& segment) {
+
+    std::vector<int> splits;
+    for (int i = 0; i < segment.size(); ++i) {
+      splits.push_back(i);
+      if (static_cast<unsigned char>(segment[i]) > 128) {
+        i++;
```
angeloskath, line 140: If my understanding is correct this skips the 2nd byte if it is a unicode character. But what if it is a unicode character with 3 or 4 bytes?

llms/export/tokenizer.cpp
```diff
@@ -0,0 +1,189 @@
...
+  auto segments = unicode_regex_split(text, pre_tokenizer_regex_);
+
+  auto one_step_merge = [this](std::string segment, std::vector<int>& splits) {
+    int merge_idx;
+    int rank = INT32_MAX;
+    for (int i = 0; i < splits.size() - 2; ++i) {
+      auto start = splits[i];
+      auto mid = splits[i + 1];
+      auto end = splits[i + 2];
+      std::string candidate = segment.substr(start, mid - start);
```
angeloskath, line 100: Absolute nitpick but if you move the candidate outside the loop and change this to 

```
candidate.clear()
candidate.insert(segment.begin() + start, segment.begin() + mid);
candidate += " ";
candidate.insert(segment.begin() + mid, segment.begin() + end);
```

then you 'll have just one heap allocation instead of `splits.size()` ones.

llms/export/tokenizer.cpp
angeloskath: If we do have the possibility for the full utf-8 range then this can be changed to

```cpp
auto c = static_cast<unsigned char>(segment[i]);
if (c > 128) {
  if ((c >> 3) == 30) {
    i += 3;
  } else if ((c >> 4) == 14) {
    i += 2;
  } else {
    i += 1;
  }
}
```

awni: I think it's correct as is. The BPE merges are specified with BPE encoded bytes.  The byte values are < 256:
- Values < 128 are ASCII and take a single char in in a std::string
- Values 128 <= x < 256 will occupy two bytes since they correspond to utf8 code points which occupy two bytes.

angeloskath: Got it so basically we don't have the full utf-8 range.

llms/export/mlxlm.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+#include "mlxlm.h"
+
+namespace mx = mlx::core;
+
+#define seconds(x)                                                             \
+  (std::chrono::duration_cast<std::chrono::nanoseconds>(x).count() / 1e9)
+#define time_now() std::chrono::high_resolution_clock::now()
+
+// Maybe compile
```
angeloskath, line 15: You are already doing that :-)

llms/export/tokenizer.cpp
```diff
@@ -0,0 +1,201 @@
...
+    for (int i = 0; i < segment.size(); ++i) {
+      splits.push_back(i);
+      if (static_cast<unsigned char>(segment[i]) >= 128) {
+        i++;
+      }
+    }
+    splits.push_back(segment.size());
+
+    while (one_step_merge(segment, splits)) {
+    };
```
angeloskath, line 156: Just curious, any particular reason that you didn't write this in one line and omitted the `;`?

awni:
Moved https://github.com/ml-explore/mlx-lm/pull/12

PR #1199, opened 2025-01-10
Utilize a specific model version from HuggingFace during LoRA fine-tuning

sindhu-satish:
Some models on HuggingFace do not have safetensor files. There are unmerged PRs on HuggingFace with these files. This commit enables users to further customize and specify the commit hash of a particular model.

### DIFF
--- a/lora/lora.py
+++ b/lora/lora.py
@@ -22,6 +22,13 @@ def build_parser():
         default="mlx_model",
         help="The path to the local model directory or Hugging Face repo.",
     )
+
+    parser.add_argument(
+        "--revision",
+        default="main",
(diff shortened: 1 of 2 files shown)


awni:
The lora in mlx-examples/lora is really meant more as a simple example rather than something we are going to keep adding features to.

If you want to update this or send a new PR for the [LoRA in mlx-lm](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/lora.py) that would make more sense.

awni:
@sindhu-satish do you plan to return to this?

sindhu-satish:
@awni yep - will update the PR soon :)

Issue #1185, opened 2024-12-30
Distributed LORA training fails with METAL error


ivanfioravanti:
In the past I tried mlx_lm.lora distributed with M2 Ultra and M3 Max and everything was working well.
Today I tried again with 1 M2 Ultra and 2 M4 Max and I get following error.

Single run works on each host, I tried them 1 by 1 using: mpirun --hostfile ~/hosts.txt -n 1 /Users/ivan/.pyenv/shims/mlx_lm.lora --model mistralai/Mistral-7B-v0.3 --train --data ~/Desktop/data --max-seq-length 8192 --batch-size 1

As soon as I start a run on 2+ hosts the error appears. I used this for 3 hosts mpirun --hostfile ~/hosts.txt -n 3 --wdir /Users/ivan/.pyenv/shims  mlx_lm.lora --model mistralai/Mistral-7B-v0.3 --train --data ~/Desktop/data --max-seq-length 8192 --batch-size 3

libc++abi: terminating due to uncaught exception of type std::runtime_error: [METAL] Command buffer execution failed: Caused GPU Timeout Error (00000002:kIOGPUCommandBufferCallbackErrorTimeout)
[Mac:46703] *** Process received signal ***
[Mac:46703] Signal: Abort trap: 6 (6)
[Mac:46703] Signal code:  (0)
[Mac:46703] [ 0] 0   libsystem_platform.dylib            0x000000018ff66de4 _sigtramp + 56
[Mac:46703] [ 1] 0   libsystem_pthread.dylib             0x000000018ff2ff70 pthread_kill + 288
[Mac:46703] [ 2] 0   libsystem_c.dylib                   0x000000018fe3c908 abort + 128
[Mac:46703] [ 3] 0   libc++abi.dylib                     0x000000018fee644c _ZN10__cxxabiv130__aligned_malloc_with_fallbackEm + 0
[Mac:46703] [ 4] 0   libc++abi.dylib                     0x000000018fed4a24 _ZL28demangling_terminate_handlerv + 320
[Mac:46703] [ 5] 0   libobjc.A.dylib                     0x000000018fb7d3f4 _ZL15_objc_terminatev + 172
[Mac:46703] [ 6] 0   libc++abi.dylib                     0x000000018fee5710 _ZSt11__terminatePFvvE + 16
[Mac:46703] [ 7] 0   libc++abi.dylib                     0x000000018fee56b4 _ZSt9terminatev + 108
[...]

PR #1200, opened 2025-01-10
reduction moved to CPU in case of distributed training

ivanfioravanti:
Workaround for #1185 in case of distributed training, as suggested by @awni

### DIFF
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -159,9 +159,8 @@ def evaluate(
         ntokens += toks
         mx.eval(all_losses, ntokens)
 
-    all_losses = mx.distributed.all_sum(all_losses)
-    stream = mx.cpu if mx.distributed.init().size() > 1 else None
-    ntokens = mx.distributed.all_sum(ntokens, stream=stream)
+    all_losses = mx.distributed.all_sum(all_losses, stream=mx.cpu)
+    ntokens = mx.distributed.all_sum(ntokens, stream=mx.cpu)
 
     return (all_losses / ntokens).item()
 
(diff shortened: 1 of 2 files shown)


llms/mlx_lm/tuner/trainer.py
```diff
@@ -160,7 +160,8 @@ def evaluate(
         mx.eval(all_losses, ntokens)
 
     all_losses = mx.distributed.all_sum(all_losses)
-    ntokens = mx.distributed.all_sum(ntokens)
+    stream = mx.cpu if mx.distributed.init().size() > 1 else None
```
angeloskath, line 163: Nit: You don't really need this line as it defaults to noop if we are not in a distributed setting.

llms/mlx_lm/tuner/trainer.py
```diff
@@ -160,7 +160,8 @@ def evaluate(
         mx.eval(all_losses, ntokens)
 
     all_losses = mx.distributed.all_sum(all_losses)
```
angeloskath, line 162: Should also be done on the cpu.

angeloskath (changes requested):
Left some small comments then we can merge. Thanks for finding and fixing this!

llms/mlx_lm/tuner/trainer.py
ivanfioravanti: But here the strange thing is that I never got timeout on this part so far, how is it possible? I will move it to cpu anyway for consistency. Usually timeout is just at the very first val loss eval

llms/mlx_lm/tuner/trainer.py
ivanfioravanti: The goal was to run the next line in CPU or GPU based on distributed setting. Are you suggesting to always run in CPU?
ntokens = mx.distributed.all_sum(ntokens, stream=stream)

llms/mlx_lm/tuner/trainer.py
ivanfioravanti: Shouldn't we move to cpu just latest operation in a distributed flow? Because first one completes anyway and pass to the next so there is no timeout waiting for slowest node 🤔

angeloskath: No I think it can timeout on the first one as well. Basically any time a GPU operation may take longer than a number of seconds (I think it is 5) then it can timeout. Besides it will be cleaner to just run this on the CPU in general. No need to think about whether we are in a distributed setting or not.

If we aren't then all ops in `mx.distributed` are simply noops.

ivanfioravanti: You are the boss here 😎 I moved everything to the CPU.

angeloskath (approved):
Thanks!

PR #1196, opened 2025-01-08
fix(lora): config yaml & arg default merge bug

jjaareet:
currently, cli arg defaults overwrite lora config yaml

```bash
cat config.yaml | grep fine_tune_type
fine_tune_type: full

mlx_lm.lora --config config.yaml
...

cat adapters/adapter_config.json | grep fine_tune_type
"fine_tune_type": "lora",
```

setting the cli arg overwrites the config yaml as expected but should not be necessary

```bash
cat config.yaml | grep fine_tune_type
fine_tune_type: full

mlx_lm.lora --config config.yaml --fine-tune-type full
...

cat adapters/adapter_config.json | grep fine_tune_type
"fine_tune_type": "full",
```

after these changes the config yaml can be used as expected

```bash
cat config.yaml | grep fine_tune_type
fine_tune_type: full

mlx_lm.lora --config config.yaml
...

cat adapters/adapter_config.json | grep fine_tune_type
"fine_tune_type": "full",
```

cli arg overwriting config yaml should continue to work as well

```bash
cat config.yaml | grep fine_tune_type
fine_tune_type: full

mlx_lm.lora --config config.yaml --fine-tune-type lora
...

cat adapters/adapter_config.json | grep fine_tune_type
"fine_tune_type": "lora",
```

changes:
- remove all cli arg defaults
  - `default=None` works as expected but removing for consistency and clarity
- add missing `CONFIG_DEFAULTS`
  - `"config": None` & `"grad_checkpoint": False`
- add missing `type=str` to args
  - not actually necessary but maintains consistency and the pattern is used elsewhere

PR #1205, opened 2025-01-14
fix(lora): add back store_true default args

jjaareet:
sorry! i did not test https://github.com/ml-explore/mlx-examples/pull/1196 well enough and caused an issue when trying to clean up default args

currently cli args with `action="store_true"` are not overwritten by config yaml because they are set to `False` instead of `None`
https://github.com/ml-explore/mlx-examples/blob/c117af83b8cbec15523bd0d69e7a57f01237ca89/llms/mlx_lm/lora.py#L289-L291

this results in needing to set `--train`, `--test`, and `--grad-checkpoint` via cli as config yaml is not respected

setting `default=None` on these args fixes the issue

### DIFF
--- a/llms/mlx_lm/lora.py
+++ b/llms/mlx_lm/lora.py
@@ -78,6 +78,7 @@ def build_parser():
         "--train",
         action="store_true",
         help="Do training",
+        default=None,
     )
     parser.add_argument(
         "--data",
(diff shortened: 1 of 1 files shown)


awni (approved):
Thanks!

PR #1212, opened 2025-01-20
Fix dataset variable name, in `datasets.py`

felladrin:
Fixes the error `NameError: name 'train_ds' is not defined`, which is currently thrown when using a configuration like this for the LORA tuner:

```yml
hf_dataset:
  name: "OpenAssistant/oasst_top1_2023-08-25"
  train_split: "train"
  valid_split: "test"
  text_feature: "text"
```


angeloskath (approved):
Thanks!

PR #1215, opened 2025-01-20
whisper: add directory transcription support

cavit99:
Add support for transcribing all files in a directory recursively. The implementation simply lets ffmpeg handle file validation instead of filtering by any extension. Update README with minimal documentation for directory support.

### DIFF
--- a/whisper/README.md
+++ b/whisper/README.md
@@ -31,6 +31,11 @@ mlx_whisper audio_file.mp3
 
 This will make a text file `audio_file.txt` with the results.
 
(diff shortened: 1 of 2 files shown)


awni:
Thanks for the contribution. For now let's hold off on merging this as it adds some code complexity for a task which should be fairly simple to do in the terminal:

```bash
find . -name '*.flac' -print | xargs mlx_whisper
```

You could alias a command like that to another name if it's tedious to type it out every time. We can revisit this if it makes sense in the future.

PR #1216, opened 2025-01-20
Some fixes for pipeline parallel deep seek r1

awni:
- Use `stream=mx.cpu` during prompt processing otherwise the gpu sometimes times out
- Allow model to be an argument in the `pipeline_generate.py` script.


angeloskath (approved):
Pity that we have to do this ...

PR #1220, opened 2025-01-23
Add the possibility to cache model instead of loading from disk each time.

haixuanTao:
This PR adds the possibility to cache the model in memory in order to transcribe audio in real time.


awni:
The model is already cached by the `ModelHolder` since it's put in a class member variable. It won't be reloaded from disk unless you explicitly change the model.

Does that work for you? If so, let's close this since it is not necessary.

haixuanTao:
Ok indeed. 

Thanks. 

Sorry then.

Issue #1224, opened 2025-01-26
Improve support for custom LoRa train method


chimezie:
It is generally useful to allow third-party, upstream software to customize the LoRa **train** function (in mlx_lm.tuner.trainer). This was the motivation for allowing user-provided loss,  iterate_batches, and training_callback functions. Another useful extension mechanism is to enable the user to specify a custom TrainingArgs data class and instance to the train function.

Other configuration-driven training capabilities could be built on the existing architecture with this change.

However, the TrainingArgs instance is not passed on to the **training_callback** or **iterate_batches** functions. Doing so would allow upstream software to pass additional information without changing the signatures of those functions.

For an example of a specific situation where making this signature change could be useful, see where the iterate_batches function [needed a breaking change to its signature](https://github.com/ml-explore/mlx-examples/pull/1103/commits/1981b131e3893a8bc8e35c5932492ce30b76b552#r1874993511) to support user-specified 'response generation tokens' that could be used to identify the boundaries between input tokens and the rest of the sequence for supporting input masking (or completion only) instruction tuning: #484 & #1103  

Specifying these tokens on a custom TrainingArgs data instance passed down to the iterate_batches functions would facilitate this extension and similar ones without breaking changes to the signatures of those methods.    

The same principle could apply to the training_callback function for other scenarios. To do so, all that would be required is to pass the training argument instance as an additional keyword argument that defaults to None. This would not break the current signatures of the methods but would also reduce the need to break them for subsequent extensions.

awni:
It seems reasonable to me! If you are interested to submit a PR that would be great.

PR #1228, opened 2025-01-28
Pass down TrainingArgs instance to iterate_batches function and TrainingCallback methods

chimezie:
Addresses #1224

### DIFF
--- a/llms/mlx_lm/tuner/trainer.py
+++ b/llms/mlx_lm/tuner/trainer.py
@@ -5,7 +5,7 @@ import shutil
 import time
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Union
+from typing import Optional, Union
 
 import mlx.core as mx
 import mlx.nn as nn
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/tuner/trainer.py
```diff
@@ -227,6 +231,7 @@ def step(batch):
             batch_size=args.batch_size,
             max_seq_length=args.max_seq_length,
             train=True,
+            args=args,
```
awni, line 234: I thought about this a bit more and it should be easy enough to enclose the training args in any custom iterate batches function (that gets passed to train). So I don't think we really need to pass it through this way.

llms/mlx_lm/tuner/trainer.py
```diff
@@ -258,7 +263,7 @@ def step(batch):
                     "val_loss": val_loss,
                     "val_time": val_time,
                 }
-                training_callback.on_val_loss_report(val_info)
+                training_callback.on_val_loss_report(val_info, args=args)
```
awni, line 266: Similarly it should be easy enough to enclose the `args` in the user provided loss reporting functions.

PR #1229, opened 2025-01-28
Better overflow correction for DeepSeek v3

awni:
Arrived at clipped values by trial-end-error on a couple queries. Results seem to be identical with/without them (up to the point where overflow would happen).

Improves generation speed about 6% as compared to bf16.


barronalex (approved):
Nice 🚀

PR #1230, opened 2025-01-29
Include revision version for HF models while loading

sindhu-satish:
Raising this PR here: https://github.com/ml-explore/mlx-examples/pull/1199

-------

Some models on HuggingFace do not have safetensor files. There are unmerged PRs on HuggingFace with these files. This commit enables users to further customize and specify the commit hash of a particular model.

### DIFF
--- a/llms/mlx_lm/lora.py
+++ b/llms/mlx_lm/lora.py
@@ -73,6 +73,13 @@ def build_parser():
         help="The path to the local model directory or Hugging Face repo.",
     )
 
+    parser.add_argument(
+        "--revision",
+        default="main",
+        type=str,
+        help="Hash value of the commit to checkout from the Hugging Face repo.",
+    )
+
     # Training args
     parser.add_argument(
         "--train",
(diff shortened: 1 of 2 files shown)


awni:
Could you give an example of the model file that is in that state.. maybe a simpler solution is to just merge those PRs?

sindhu-satish:
@awni yeah, there are actually a ton of models with this issue. You can find some of these PRs here: https://huggingface.co/SFconvertbot/activity/all. These are just the PRs the bot raised. There are a lot of safetensor PRs that humans raise too.

Also the code defaults to `revision='main'` in case someone does not want/need this feature

awni:
Is there a specific model you've actually used this feature for? I'd like to test it?

sindhu-satish:
Yeah, you can try this: https://huggingface.co/deepseek-ai/deepseek-math-7b-base/tree/main . use this PR: https://huggingface.co/deepseek-ai/deepseek-math-7b-base/commit/6779f1b605fed54c5115ab537563ac0943c265fe

PR #1231, opened 2025-01-29
Bug fix - enabling Qwen2 support

sindhu-satish:
Just added `qwen2` in `MODEL_MAPPING` in `utils.py`, so `mlx_lm.models.qwen2` can be successfully imported. Tested this out with `mlx-community/Unsloth-DeepSeek-R1-Distill-Qwen-32B-4bit` on HF.

### DIFF
--- a/llms/mlx_lm/lora.py
+++ b/llms/mlx_lm/lora.py
@@ -73,13 +73,6 @@ def build_parser():
         help="The path to the local model directory or Hugging Face repo.",
     )
 
-    parser.add_argument(
-        "--revision",
-        default="main",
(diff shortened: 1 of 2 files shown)


awni:
@sindhu-satish could you split out the qwen2 diff from the revision support? That way we can land the qwen2 fix right away.

sindhu-satish:
Done!

awni:
Actually I'm not entirely sure what is the purpose of this PR? There is no need for a remapping of qwen2 to qwen2. Was there a speicifc issue you are facing?

PR #1240, opened 2025-02-02
[mlx-lm] For pipeline parallel only download local shard

awni:
- Only download weights needed for the local shard for the pipeline parallel model
- Also added option to make dsv2 pipeline parallel for faster testing on smaller models

### DIFF
--- a/llms/mlx_lm/examples/pipeline_generate.py
+++ b/llms/mlx_lm/examples/pipeline_generate.py
@@ -4,10 +4,11 @@
 Run with:
 
 ```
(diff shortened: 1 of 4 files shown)


barronalex (approved):
Nice!
I should have messaged I also wrote a version of this earlier today but I like your way of doing it with `model.pipeline` followed by checking the index.

awni:
Oops! Sorry should have mentioned it also!

barronalex:
All good! Excited to use this later.

PR #1242, opened 2025-02-04
Fix deepseek sharding

awni:
Broke a couple things in https://github.com/ml-explore/mlx-examples/pull/1240, this fixes that.


angeloskath (approved):
👍

PR #1257, opened 2025-02-07
Faster DSv2/3 expert score computation

awni:
Uses put along axis directly (simpler) and compiles for v3. For R1 it's between 0.5-1 tok/sec.


angeloskath (approved):
👍

PR #1259, opened 2025-02-07
Some improvements to speedup alignment computation in MLX Whisper

awni:
- Don't upcast qk scores until needed
- Don't store alignment head info as mx.array since we only use it as if it were a list (all the `.item` are causing lots of graph evals)
- Use precise softmax

Speedup is pretty small but also good to demonstrate correct usage.


angeloskath (approved):
￼

PR #1260, opened 2025-02-08
Add logits processor to spec gen

### LINKED ISSUE #1255: [mlx_lm] Use logits_processors in speculative_generate_step
Right now, `logits_processors` is unused in `speculative_generate_step` https://github.com/ml-explore/mlx-examples/blob/9c2ef38d4d7a6c6afc1f9c90cb26cadb95f7343e/llms/mlx_lm/utils.py#L335 so things like repetition penalty and whatever else one may want to do in a logits processor is not being applied when using speculative decoding.

awni:
Closes #1255

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -368,8 +368,8 @@ def speculative_generate_step(
         Tuple[mx.array, mx.array]: One token and a vector of log probabilities.
     """
 
-    y = prompt
-    tokens = None
+    y = prompt.astype(mx.uint32)
+    prev_tokens = None
 
     # Create the KV cache for generation
     if prompt_cache is None:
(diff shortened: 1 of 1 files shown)


llms/mlx_lm/utils.py
```diff
@@ -390,17 +390,41 @@ def speculative_generate_step(
...
+                    y = y[: -(n_predict - 1)]
+                for i in range(n_predict):
+                    prev_tokens = (
+                        mx.concat([prev_tokens, y]) if prev_tokens is not None else y
+                    )
+                    y, logprobs = _process_and_sample(
+                        prev_tokens, logits[:, i : i + 1, :]
+                    )
+                    out_y.append(y)
+                    out_logprobs.append(logprobs)
```
awni, line 422: The unrolling here is a bit unfortunate.. but the logits processors can be autoregressive themselves (e.g. the `repetition_penalty`.

angeloskath (approved):
👍

Honestly I don't mind the unrolling, it is quite natural. The only annoying thing is the manhandling of the `prev_tokens` but I don't see anything that is clearly better.

PR #1267, opened 2025-02-09
Remove temp argument

applecool:
Currently, the `temp` argument [here](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/examples/chat.py#L26) causes the `generate()` to fail with the following error:
`TypeError: generate_step() got an unexpected keyword argument 'temp'`

Removal of this deprecated argument fixes it. 

cc @awni


awni (approved):
Ty!

PR #1271, opened 2025-02-10
Update requirements.txt - adding torch

mike-allaway:


awni:
torch shouldn't be a dependency for this example.

PR #1276, opened 2025-02-11
Fix sharding for more even number of layers

awni:
It was possible for the last rank to get 0 / a pretty uneven number of layers before.

For example for 60 layers with 31 machines.


barronalex (approved):
Nice!

PR #1278, opened 2025-02-12
Fix lora timings after validation

awni:
Accidentally starting the clock in the wrong place after validation would give artificially high times for the next report. Timings make sense now.


angeloskath (approved):
👍

PR #1283, opened 2025-02-13
Add plamo-2-1b model

mitmul:
This PR adds the latest SLM from Preferred Networks, PLaMo-2-1B.

Currently it can successfully generate correct responses when I use the following code:
```python
import mlx.core as mx
from mlx_lm.utils import load
from mlx_lm.models import cache

model, tokenizer = load(
    "../plamo-2-1b-mlx",
    tokenizer_config={
        "trust_remote_code": True,
    },
    model_config={
        "trust_remote_code": True,
    },
)

user_input = "美味しいカレーの作り方のレシピを紹介します。"

input_ids = tokenizer.encode(user_input)
input_ids = mx.array(input_ids, dtype=mx.int64)

y = input_ids
prompt_cache = cache.make_prompt_cache(model)

for _ in range(100):
    logits = model(inputs=y[None], cache=prompt_cache)
    logits = logits[:, -1, :]
    logprobs = logits - mx.logsumexp(logits, keepdims=True)
    next_y = mx.argmax(logprobs, axis=-1)

    response = tokenizer.decode(next_y.item())
    print(response, end="")
    y = mx.concat([y, mx.array(next_y)])
```

However, when I run the following command from my terminal:
```bash
python -m mlx_lm.generate \
--model plamo-2-1b-mlx \
--prompt '美味しいカレーの作り方のレシピを紹介します。' \
--verbose true \
--max-tokens 128 \
--ignore-chat-template
```
the outputs are totally corrupted.
Now I'm investigating the cause of why `mlx_lm.generate` doesn't work for this model.
I guess it relates to the cache implementation in this code.


mitmul:
I found a bug in my code that performs causal-conv1d update, which was related to the difference between the channel-first conv1d in torch and channel-last conv1d in MLX. I fixed it and confirmed that it runs properly as expected, so I changed this PR's state to review-ready 🙏

awni:
I tried running:

````
mlx_lm.generate --model pfnet/plamo-2-1b --prompt "The talles mountain in"
```

But it crashes with missing layers. Did you encounter that?

```
ValueError: Received parameters not in model: model.layers.layers.13.post_mixer_norm.weight model.layers.layers.10.mixer.A_log model.layers.layers.6.mixer.dt_bias model.layers.layers.14.mlp.gate_up_proj.weight model.layers.layers.4.mixer.dt_norm_weight model.layers.layers.3.post_mlp_norm.weight model.layers.layers.8.mixer.A_log model.layers.layers.15.mixer.qkv_proj.weight model.layers.layers.3.pre_mlp_norm.weight ...```

**Comment by mitmul** [comment]
Thank you for checking. That's because the `config.json` in `pfnet/plamo-2-1b` has `"model_type": "plamo"` and the old architecture for PLaMo-13B has existed in mlx_lm as `plamo`, so the older one is selected in your case instead of the right one, `plamo2` (Samba architecture).

Now I'm discussing with the members of the model development in Preferred Networks if we can change the value of `model_type` in config.json on the huggingface repo: `pfnet/plamo-2-1b`.

**Comment by mitmul** [comment]
@awni Sorry for taking time to update the original `config.json`, but now it's updated from `"model_type": "plamo"` to `"model_type": "plamo2"`: https://huggingface.co/pfnet/plamo-2-1b/discussions/2/files and I confirmed that the following command successfully works:
```
$ python -m mlx_lm.generate --model pfnet/plamo-2-1b --prompt "The talles mountain in"
config.json: 100%|███████████████████████████████████████████████████████████████████████| 1.38k/1.38k [00:00<00:00, 23.0MB/s]
modeling_plamo.py: 100%|█████████████████████████████████████████████████████████████████| 66.2k/66.2k [00:00<00:00, 14.7MB/s]
Fetching 15 files: 100%|██████████████████████████████████████████████████████████████████████| 15/15 [00:00<00:00, 24.59it/s]
==========
 southeastern France, called the Montagnes Noires (Black Mountains) in French, is the highest mountain range in Europe, with peaks reaching heights of over 10,000 feet (3,000 m). The range is bordered by the Mediterranean Sea to the west, the Massif Central to the east, and the Pyrenees to the south. The highest peak in the range is Mont Blanc (Mont Blanc), which rises to 15,771 feet (4,809 m) above sea level.

The Montagne
==========
Prompt: 6 tokens, 11.293 tokens-per-sec
Generation: 100 tokens, 45.815 tokens-per-sec
Peak memory: 5.294 GB
```

Could you take a look into this PR again? Thanks.

**Comment by awni** [comment]
This is the error I'm getting now:
```
  File "/Users/awnihannun/miniconda3/bin/mlx_lm.generate", line 33, in <module>
    sys.exit(load_entry_point('mlx-lm', 'console_scripts', 'mlx_lm.generate')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/awnihannun/mlx-examples/llms/mlx_lm/generate.py", line 195, in main
    model, tokenizer = load(
                       ^^^^^
  File "/Users/awnihannun/mlx-examples/llms/mlx_lm/utils.py", line 791, in load
    model, config = load_model(model_path, lazy)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/awnihannun/mlx-examples/llms/mlx_lm/utils.py", line 729, in load_model
    model = model_class(model_args)
            ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/awnihannun/mlx-examples/llms/mlx_lm/models/plamo2.py", line 1602, in __init__
    self.model = PlamoModel(config)
                 ^^^^^^^^^^^^^^^^^^
  File "/Users/awnihannun/mlx-examples/llms/mlx_lm/models/plamo2.py", line 1420, in __init__
    self.embed_tokens = nn.Embedding(config.vocab_size, config.hidden_size)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/awnihannun/mlx/python/mlx/nn/layers/embedding.py", line 25, in __init__
    self.weight = mx.random.normal(shape=(num_embeddings, dims), scale=scale)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: [reshape] Cannot reshape array of size 0 into shape (2).
```

**Comment by awni** [comment]
Could you add a test for the Plamo2 model [here](https://github.com/ml-explore/mlx-examples/blob/main/llms/tests/test_models.py)?

**Comment by mitmul** [comment]
@awni I added a test case for plamo2. Well, I could not reproduce the error you showed in https://github.com/ml-explore/mlx-examples/pull/1283#issuecomment-2676231153 . Could you please tell me how to reproduce it? I tried the following commands and I couldn't see the same errors:
```sh
$ python3 -V
Python 3.12.9
$ python3 -m venv venv --copies
$ source venv/bin/activate
$ pip install "git+https://github.com/mitmul/mlx-examples.git@mitmul/add-plamo2-1b-support#egg=mlx-lm&subdirectory=llms"
$ pip install numba
$ python -m mlx_lm.generate --model pfnet/plamo-2-1b --prompt "The talles mountain in"
Fetching 15 files: 100%|███████████████████████████████████████████| 15/15 [00:00<00:00, 245760.00it/s]
==========
 southeastern France, called the Montagnes Noires (Black Mountains) in French, is the highest mountain range in Europe, with peaks reaching heights of over 10,000 feet (3,000 m). The range is bordered by the Mediterranean Sea to the west, the Massif Central to the east, and the Pyrenees to the south. The highest peak in the range is Mont Blanc (Mont Blanc), which rises to 15,771 feet (4,809 m) above sea level.

The Montagne
==========
Prompt: 6 tokens, 81.134 tokens-per-sec
Generation: 100 tokens, 45.491 tokens-per-sec
Peak memory: 5.289 GB
```

I'm wondering if you have installed `numba` before running the generate command? The plamo2 requires numba package for its original tokenizer.

awni:
It's working for me now! Thanks for the test!

awni (approved):

llms/mlx_lm/models/plamo2.py
```diff
@@ -1054,113 +374,53 @@ def __init__(self, config: ModelArgs, layer_idx: int) -> None:
...
-                scale=self.scale,
-                mask=attention_mask,
-            )
+        q = q.reshape(B, T, self.q_num_heads, self.qk_dim).transpose(0, 2, 1, 3)
+        k = k.reshape(B, T, self.k_num_heads, self.qk_dim).transpose(0, 2, 1, 3)
+        v = v.reshape(B, T, self.v_num_heads, self.v_dim).transpose(0, 2, 1, 3)
 
-        attn_output = attn_output.transpose(0, 2, 1, 3)
+        q = mx.fast.layer_norm(q, None, None, 1e-6) * self.q_weight[:, None]
+        k = mx.fast.layer_norm(k, None, None, 1e-6) * self.k_weight[:, None]
```
mitmul, line 396: @awni Thanks for cleaning up the code, but I wonder if this part should be `mx.fast.rms_norm` because it seems to be come from these lines:

https://github.com/mitmul/mlx-examples/blob/fb1559e1f315ee69e85de9b060365f1a1cd7cd03/llms/mlx_lm/models/plamo2.py#L1083-L1084

awni: Layer norm without a weight or bias is the same as rms norm without a weight of bias. This is just a hack because our layer norm supports not setting both whereas the RMS norm requires a weight.

llms/mlx_lm/models/plamo2.py
```diff
@@ -1054,113 +374,53 @@ def __init__(self, config: ModelArgs, layer_idx: int) -> None:
...
-                # `generate` function creates attention mask that does not consider sliding window
-                m_swa = m_swa[None, None]
-                attention_mask = attention_mask[:, :, -query_states.shape[2] :, -key_states.shape[2] :]
-                attention_mask = mx.where(m_swa, attention_mask, float("-inf"))
-
-            # like AttentionMaskConverter._unmask_unattended in huggingface.transfoermers,
-            # we need to attend to all tokens in masked rows for `scaled_dot_product_attention`
-            bool_mask = mx.logical_not(mx.isneginf(attention_mask))
-            valid_tokens = mx.sum(bool_mask, axis=-1).astype(mx.bool_)  # type: ignore  # (..., q_len)
-            attention_mask = mx.where(valid_tokens[..., None], attention_mask, float(0.0))
```
mitmul: @awni I think we need to revert this part for Sliding Window Attention adopted in PLaMo2?

awni: No we should just use the `RotatingKVCache` instead in the `make_cache` method. I left a little TODO there.

mitmul: Ah, that makes sense. Thank you for letting me know that class.

llms/mlx_lm/models/plamo2.py
awni: You're right.. that's my mistake, they are not the same.. I introduced a bug with that change. I will get it fixed with your PR.. thanks for flagging.

PR #1287, opened 2025-02-14
[transformer_lm] Use more standard window strategy

### LINKED ISSUE #1286: transformer_lm freezing at training 1000 iters flagged for gpu
I've repeated this test a few times and with --num_iters at the default 100000 I always stop getting feedback at 1000, but the GPU History still shows that they're slammed.  Is that normal?  Has anyone else seen this?  Are there limits on the mlx AdamW optimizer?  I saw similar issues trying to translate Karpathy's nanogpt to mlx but it always died at 586 iterations, for no apparent reason.

Studio M1 Ultra 48gpu 128gb unified 4Tb ssd

awni:
- Use a more standard windowing strategy so we don't blow up the dataset size by duplicating a bunch of tokens per example
- Remove numpy dependency

Closes #1286

### DIFF
--- a/transformer_lm/main.py
+++ b/transformer_lm/main.py
@@ -8,7 +8,6 @@ import datasets
 import mlx.core as mx
 import mlx.nn as nn
 import mlx.optimizers as optim
-import numpy as np
 from mlx.utils import tree_flatten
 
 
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
👍

PR #1291, opened 2025-02-19
Fix logits processor bugs with spec dec

awni:
Fixes ~3 bugs with spec-dec:

- Pass 2D logits to logits processor
- Fix previous token book-keeping
- Incorrect normalization of log probs for base model


awni:
CC @mattjcly

angeloskath (approved):
🙏

PR #1300, opened 2025-02-25
Mixed quant recipes

pcuenca:
Adding a couple of mixed 2/6, 3/6 quantization recipes for discussion, based on @barronalex's [gist](https://gist.github.com/barronalex/84addb8078be21969f1690c1454855f3). I think that mixed 4/6 is the same approach used now when q_bits is 4.

We could add some additional interesting recipes if there's interest.

I'll also run some tests on a couple of models.


awni (approved):
Very nice, thank you!

PR #1308, opened 2025-02-27
Fix plamo2 model to use rms_norm

mitmul:
Thank you for fixing my messy code in #1283. However, it seems that some parts of the model code have changed, leading to slightly different behavior from the original. This PR aims at fixing it.

### DIFF
--- a/llms/mlx_lm/models/plamo2.py
+++ b/llms/mlx_lm/models/plamo2.py
@@ -354,15 +354,6 @@ class Mamba(nn.Module):
         return y
 
 
-def swa_mask(q_len: int, kv_len: int, window_size: int) -> mx.array:
-    max_len = max(q_len, kv_len)
-    mask = mx.tril(
-        mx.triu(mx.ones((max_len, max_len), dtype=mx.bool_), k=-window_size),  # type: ignore
-        k=window_size,
-    )
-    return mask[-q_len:, -kv_len:]
-
-
 class Attention(nn.Module):
     def __init__(self, config: ModelArgs) -> None:
         super().__init__()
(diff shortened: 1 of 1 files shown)


mitmul:
I removed the redundant parts of the code because, as mentioned in the comment [here](https://github.com/ml-explore/mlx-examples/pull/1283#discussion_r1973891880), sliding window attention should be implemented using RotatingKVCache rather than being directly incorporated into the model code.

But I still see some differences between MLX's layer norm and the `_rms_norm` func in the plamo2 modeling code as shown in this comment: https://github.com/ml-explore/mlx-examples/pull/1283#discussion_r1974165042

llms/mlx_lm/models/plamo2.py
```diff
@@ -392,8 +400,8 @@ def __call__(
         k = k.reshape(B, T, self.k_num_heads, self.qk_dim).transpose(0, 2, 1, 3)
         v = v.reshape(B, T, self.v_num_heads, self.v_dim).transpose(0, 2, 1, 3)
 
-        q = mx.fast.layer_norm(q, None, None, 1e-6) * self.q_weight[:, None]
-        k = mx.fast.layer_norm(k, None, None, 1e-6) * self.k_weight[:, None]
+        q = _rms_norm(q, 1e-6) * self.q_weight[:, None]
+        k = _rms_norm(k, 1e-6) * self.k_weight[:, None]
```
awni, line 404: Thanks for fixing this. Once we have the new MLX release we can use `mx.fast.rms_norm` directly (with a `None` weight) and it should be faster.

awni (approved):

PR #1311, opened 2025-02-28
chore(lora): Add real-time log buffering fix for nohup execution

PierreLouisLetoquart:
Fixes issue where nohup.out only shows logs after script completion, making real-time monitoring impossible. Detects Python version and applies appropriate buffering fix.

### DIFF
--- a/lora/lora.py
+++ b/lora/lora.py
@@ -1,9 +1,9 @@
 # Copyright © 2023-2024 Apple Inc.
 
-import sys
 import argparse
 import json
 import math
+import sys
 import time
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


lora/lora.py
```diff
@@ -15,6 +17,12 @@
 from models import LoRALinear
 
 
+# Disable output buffering to see print statements in real-time
+if sys.version_info >= (3, 7):
+    sys.stdout.reconfigure(line_buffering=True)
+else:
+    sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 1)
```
awni, line 24: We dont' need to check for 3, 7 here because the package only works on 3.9 or higher.

PierreLouisLetoquart: I fixed it

awni:
Can you run the format hooks?

awni (approved):

PR #1312, opened 2025-02-28
support kimi + more options in chat mode

awni:
Support the new Kimi MOE model (it's model type is `deepseek_v3` but it uses regular RoPE).

Some QOL updates to `mlx_lm.chat` to allow reset and print help.


angeloskath (approved):
🙏

PR #1314, opened 2025-02-28
Allow mask prompt in config

### LINKED ISSUE #1313: --mask-prompt
Hi - in the Lora.md a it mentions that one can pass --mask-prompt to calculate loss only on the completion. 

Is this actually implemented as I can't get mlx_lm.lora  to accept --mask-prompt and when I put it in the yaml I am not sure there's any effect ?

awni:
Closes #1313


angeloskath (approved):
👍

PR #1315, opened 2025-03-01
Adding multiple optimizers to mlx lm

Goekdeniz-Guelmez:
Added ability to select between Adam, AdamW, and Muon optimizers with customizable parameters via command-line or YAML config. This gives users more flexibility to optimize their fine-tuning process for different models and datasets.

### DIFF
--- a/llms/mlx_lm/lora.py
+++ b/llms/mlx_lm/lora.py
@@ -47,12 +47,6 @@ CONFIG_DEFAULTS = {
     "optimizer_config": {
         "adam": {"bias_correction": False},
         "adamw": {"weight_decay": 0.01, "bias_correction": False},
-        "muon": {
-            "momentum": 0.95,
-            "weight_decay": 0.01,
-            "nesterov": True,
-            "ns_steps": 5,
-        },
     },
     "data": "data/",
     "seed": 0,
@@ -109,9 +103,9 @@ def build_parser():
     parser.add_argument(
         "--optimizer",
         type=str,
-        choices=["adam", "adamw", "muon"],
+        choices=["adam", "adamw"],
         default="adam",
-        help="Optimizer to use for training: adam, adamw, or muon",
+        help="Optimizer to use for training: adam or adamw",
     )
     parser.add_argument(
         "--mask-prompt",
@@ -256,8 +250,6 @@ def train_model(
         opt_class = optim.Adam
     elif optimizer_name == "adamw":
         opt_class = optim.AdamW
-    elif optimizer_name == "muon":
-        opt_class = optim.Muon
     else:
         raise ValueError(f"Unsupported optimizer: {optimizer_name}")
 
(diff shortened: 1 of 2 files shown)


llms/mlx_lm/lora.py
```diff
@@ -229,11 +254,38 @@ def train_model(
...
-        learning_rate=(
-            build_schedule(args.lr_schedule) if args.lr_schedule else args.learning_rate
+
+    # Initialize the selected optimizer
+    lr = build_schedule(args.lr_schedule) if args.lr_schedule else args.learning_rate
+    
+    optimizer_name = args.optimizer.lower()
+    optimizer_config = args.optimizer_config.get(optimizer_name, {})
+    
+    if optimizer_name == "adam":
```
awni, line 264: Change the switch to set `opt_class = optim.Adam` (e.g. just the class).

Then make the `opt` from the config and class like so:

```
opt = opt_class(**optimizer_config)
```

Also in the config only set parameters which don't have a default for the optimizer. Otherwise let's just inherit the defaults rather than duplicating them here.

llms/mlx_lm/lora.py
```diff
@@ -43,6 +43,26 @@
...
+            "eps": 1e-8,
+            "weight_decay": 0.01,
+            "bias_correction": False
+        },
+        "muon": {
+            "momentum": 0.95,
+            "weight_decay": 0.01,
+            "nesterov": True,
+            "ns_steps": 5
+        }
```
awni, line 64: Get rid of `muon` since it's not in the core package. We can add it back in if that changes.

llms/mlx_lm/lora.py
```diff
@@ -43,6 +43,26 @@
     "model": "mlx_model",
     "train": False,
     "fine_tune_type": "lora",
+    "optimizer": "adam",
+    "optimizer_config": {
+        "adam": {
+            "betas": [0.9, 0.999],
+            "eps": 1e-8,
```
awni, line 50: See comment below about only setting parameters here which don't have defaults for a given optimizer. Most of them should have defaults.

awni:
Thanks for adding this! Left a few comments. Please take a look, then we can merge!

Goekdeniz-Guelmez:
Great, thanks for the review! I ran the format hooks as well and it should be mergable now!

llms/mlx_lm/lora.py
```diff
@@ -95,14 +100,19 @@ def build_parser():
         choices=["lora", "dora", "full"],
         help="Type of fine-tuning to perform: lora, dora, or full.",
     )
-
+    parser.add_argument(
+        "--optimizer",
+        type=str,
+        choices=["adam", "adamw"],
+        default="adam",
```
awni, line 107: One more thing, I think this needs to default to `None` to get properly read from the config file.

That logic goes:

if arg is None:
  arg = from_config_file

if arg is still None:
  arg = from CONFIG DEFAULTS

Goekdeniz-Guelmez:
Default should be None now.

awni:
Flaky test failure is unrelated to the diff, fix is in #1322

awni (approved):
Thanks for the addition!

PR #1323, opened 2025-03-05
Change DEFAULT_SEED to None for stochastic generation by default

cavit99:
## Change DEFAULT_SEED to None for stochastic generation by default

This PR implements the suggestion from issue #1318 to change the default seed behavior for text generation.

### Changes
- Changed `DEFAULT_SEED` from `0` to `None` in both `generate.py` and `chat.py`
- Modified the main functions to only set the seed when explicitly provided
- Updated help text to clarify the new default behavior

### Rationale
As described in issue #1318:

> The mlx_lm scripts (chat.py and generate.py) currently enforce a fixed default seed of 0 for text generation, resulting in identical outputs across runs even when using stochastic sampling (e.g., --temp 1). This behavior stems from an explicit call to mx.random.seed(0) in both scripts, overriding the natural time-based RNG provided by MLX's core library in mlx.core.random.
> 
> While this ensures reproducibility, it contrasts with the typical expectation in LLM interfaces—random outputs by default unless a seed is explicitly set—and requires manual workarounds for varied outputs.

This implementation follows the suggestion by maintainer @awni:

> Why don't we just default the seed to None instead of 0? If it's not set then it takes MLXs time-based default instead of 0, which I think is pretty reasonable. One can always explicitly set the seed for reproducibility.

### Benefits
- Aligns with LLM industry standards (random outputs by default)
- Maintains backward compatibility (users can still specify `--seed 0` for old behavior)
- Uses MLX's built-in time-based RNG when no seed is provided
- Requires minimal code changes with no API changes

This change improves usability for creative and exploration use cases while maintaining the ability to get reproducible outputs when needed.

### DIFF
--- a/llms/mlx_lm/chat.py
+++ b/llms/mlx_lm/chat.py
@@ -11,7 +11,7 @@ from .utils import load, stream_generate
 
 DEFAULT_TEMP = 0.0
 DEFAULT_TOP_P = 1.0
-DEFAULT_SEED = 0
+DEFAULT_SEED = None
 DEFAULT_MAX_TOKENS = 256
 DEFAULT_MODEL = "mlx-community/Llama-3.2-3B-Instruct-4bit"
 
@@ -36,7 +36,12 @@ def setup_arg_parser():
     parser.add_argument(
         "--top-p", type=float, default=DEFAULT_TOP_P, help="Sampling top-p"
     )
-    parser.add_argument("--seed", type=int, default=DEFAULT_SEED, help="PRNG seed")
+    parser.add_argument(
+        "--seed",
+        type=int,
+        default=DEFAULT_SEED,
+        help="PRNG seed (default: use MLX's time-based RNG)",
+    )
     parser.add_argument(
         "--max-kv-size",
         type=int,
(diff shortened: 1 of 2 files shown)


awni:
Can you install the pre-commit format hooks (and run them for this diff)?

llms/mlx_lm/generate.py
```diff
@@ -87,7 +87,12 @@ def setup_arg_parser():
         default=DEFAULT_MIN_TOKENS_TO_KEEP,
         help="Minimum tokens to keep for min-p sampling.",
     )
-    parser.add_argument("--seed", type=int, default=DEFAULT_SEED, help="PRNG seed")
+    parser.add_argument(
+        "--seed",
+        type=int,
+        default=DEFAULT_SEED,
+        help="PRNG seed (default: use MLX's time-based RNG)",
```
awni, line 94:
```suggestion
        help="PRNG seed",
```

llms/mlx_lm/chat.py
```diff
@@ -36,7 +36,12 @@ def setup_arg_parser():
     parser.add_argument(
         "--top-p", type=float, default=DEFAULT_TOP_P, help="Sampling top-p"
     )
-    parser.add_argument("--seed", type=int, default=DEFAULT_SEED, help="PRNG seed")
+    parser.add_argument(
+        "--seed",
+        type=int,
+        default=DEFAULT_SEED,
+        help="PRNG seed (default: use MLX's time-based RNG)",
```
awni, line 43:
```suggestion
        help="PRNG seed",
```

awni (approved):
Thanks for the update!!

PR #1324, opened 2025-03-06
Remove lm head if unused

awni:
Remove the LM head in llama models if unused so weight validation doesn't fail.

Closes https://huggingface.co/spaces/mlx-community/mlx-my-repo/discussions/36


angeloskath (approved):
👍

PR #1326, opened 2025-03-07
Fix mixed quant option

awni:
The argument was broken because of the string value, but not using the string value causes a name clash. Using a simpler approach here.


angeloskath (approved):
👍

PR #1342, opened 2025-03-13
Fix prompt cache issue in server.py. 

sigjhl:
Addresses the non-functioning prompt_cache in server.py. 

Honestly I don't understand why this fixes it, but prompt_cache seems to become functional with this change.

### DIFF
--- a/llms/mlx_lm/server.py
+++ b/llms/mlx_lm/server.py
@@ -452,17 +452,24 @@ class APIHandler(BaseHTTPRequestHandler):
 
     def get_prompt_cache(self, prompt):
         cache_len = len(self.prompt_cache.tokens)
+        # Check if the cache is valid for the current prompt
         if (
             self.prompt_cache.model_key != self.model_provider.model_key
             or cache_len >= len(prompt)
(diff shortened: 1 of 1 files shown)


sigjhl:
> What's the issue you are seeing?

I made a simple HTML frontend for mlx_lm.server, but the prompt cache seemed to not work with a long initial prompt. 
Specifically, the very short (1-2 tokens) follow-up prompt would take as long as processing the first prompt (about 3000-5000 tokens long). 

<schematic>
1st prompt (5k tokens)
1st response (300 tokens):~30s TTFT
2nd prompt ("And?")
2nd response (300 tokens): ~30s TTFT

I thought it was something wrong with the HTML but eventually fixing the server.py like that worked for me. Sorry I got a little carried away; should have raised an issue first but fixing it after struggling for hours I got too excited. 

Sorry if the bug is not reproducible!

awni:
I added a fix here: https://github.com/ml-explore/mlx-lm/pull/29 and we can close this in favor of that.

This case was previously unsupported but it should be much faster now for multiple consecutive queries against the same prompt.

PR #1348, opened 2025-03-16
Update mixed quant predicates to include high_bits parameter

### LINKED ISSUE #1347: [Bug] Default `quant_predicate` constructors for `mixed_2_6` and `mixed_3_6` are not initialized correctly
The user should rightfully assume that variables named `mixed_2_6` and `mixed_3_6`, especially when given as model values in the utility itself, would be initialized such that the mixed quantization state initialized would be bounded by a **low of 2** and a **high of 6**.

That is unfortunately not how this is currently working.

https://github.com/ml-explore/mlx-examples/blob/2fce02acd87193487eafeeec639eb8903cc96083/llms/mlx_lm/utils.py#L1055-L1056

The `low_bits` is passed in correctly, but `high_bits` is left to the default value:

https://github.com/ml-explore/mlx-examples/blob/2fce02acd87193487eafeeec639eb8903cc96083/llms/mlx_lm/utils.py#L1019-L1021

To address this, I will create a PR that explicitly sets `high_bits` for these two defaults.

sealad886:
Explicitly pass the high_bits parameter in the mixed_2_6 and mixed_3_6 quant predicates instead of using default value.

Fixes ml-explore/mlx-lm#17

### DIFF
--- a/llms/mlx_lm/utils.py
+++ b/llms/mlx_lm/utils.py
@@ -1052,8 +1052,8 @@ def mixed_quant_predicate_builder(
     return mixed_quant_predicate
 
 
-mixed_3_6 = mixed_quant_predicate_builder(low_bits=3)
-mixed_2_6 = mixed_quant_predicate_builder(low_bits=2)
+mixed_3_6 = mixed_quant_predicate_builder(low_bits=3, high_bits=6)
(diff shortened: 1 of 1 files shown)


awni:
Thanks for the fix. Can you please move this to the new mlx-LM repo?

sealad886:
See migrated PR [here](https://github.com/ml-explore/mlx-lm/pull/20).

PR #1350, opened 2025-03-16
LoRa/README.md should be --hf-path instead of --hf-repo

tdu-naifen:
--hf-repo is not part of `fuse.py` arguments


awni (approved):
Thanks!

PR #1358, opened 2025-03-31
doc: 1、中文说明

huashuiwujiale:


angeloskath:
Thank you for writing that! Unfortunately I am inclined to reject the PR as none of us speaks and writes Chinese and I feel automatic tools are good enough to generate this from the English version when needed. It also would put the burden on us to maintain it and provide it in more languages, Spanish, Hindi, French, German where do we stop?

Once again thanks for taking the time and for your interest in MLX.

PR #1359, opened 2025-04-04
Made llama and mistral files mypy compatible

ParamThakkar123:
solves a part of #829

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,5 @@ workflows:
           type: approval
       - apple/authenticate:
           context: pr-approval
-      - mlx_lm_build_and_test:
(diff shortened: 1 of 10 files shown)


ParamThakkar123:
lora files added

ParamThakkar123:
@awni I get this error on the CI. How can we fix this ?? 
Error calling workflow: 'prb'
Cannot find a definition for job named mlx_lm_build_and_test

awni (approved):

awni:
Thanks for the updates! I didn't keep all of them because they weren't the same behavior / suboptimal. But hopefully the mypy still runs cleaner after this.

ParamThakkar123:
Thanks @awni !

PR #1390, opened 2025-10-04
Update cli.py command help text lines mentioning --word-timestamps

acsr:
There were several mentions of `--word-timestamps` option as `--word_timestamps`and another one using underscores not matching the actual option.


angeloskath (approved):
Thanks!

PR #1403, opened 2026-01-10
llama inference with tensor parallelism

stefpi:
related to [#2973](https://github.com/ml-explore/mlx/pull/2973) on mlx repo.

- made llama.py use tensor parallelism if run with `mlx.launch` on 2 or more devices
- also added class_predicate to quantize because it wasn’t working properly with quantized mlx models from HF


llms/llama/llama.py
```diff
@@ -48,6 +52,16 @@ def __init__(self, args: ModelArgs):
             args.head_dim, traditional=args.rope_traditional, base=args.rope_theta
         )
 
+    def shard(self, group: mx.distributed.Group):
+        self.n_heads = self.n_heads // group.size()
+        self.n_kv_heads = self.n_kv_heads // group.size()
+        self.repeats = self.n_heads // self.n_kv_heads
```
awni: Not necessary to scale repeats (it should be unchanged)
```suggestion
```

llms/llama/llama.py
```diff
@@ -339,9 +369,20 @@ def load_model(model_path):
         quantization = config.pop("quantization", None)
     model = Llama(ModelArgs(**config))
     if quantization is not None:
-        nn.quantize(model, **quantization)
+        class_predicate = (
+            lambda p, m: isinstance(m, (nn.Linear, nn.Embedding))
+            and f"{p}.scales" in weights
+        )
```
awni, line 374: I'm curious about that change? Was it needed for a specific model?

stefpi: I was having trouble with running quantized models, specifically [Llama-2-4bit](https://huggingface.co/mlx-community/Llama-2-7b-chat-4-bit/tree/main), and I saw this being used in the gguf llm example for quantized models.

PR #1409, opened 2026-03-11
Add wan 2.1 model

belkakari:
This PR adds a WAN 2.1 text2video and image2video model support with optimizations like TeaCache and step distilled models support. Based on [original WAN 2.1 implementation](https://github.com/Wan-Video/Wan2.1) and [LightX2V](https://github.com/ModelTC/LightX2V)

Basic commands:

1.3B text-to-video
```
python txt2video.py 'A cat playing piano' --output out.mp4
```

14B text-to-video
```
python txt2video.py 'A cat playing piano' --model t2v-14B --quantize --output out_14B.mp4
```

14B image-to-video
```
python img2video.py 'Astronaut riding a horse' \
   --image ./inputs/astronaut-on-a-horse.png --quantize --output out_i2v.mp4
```

Step distilled models:
T2V
```
wget https://huggingface.co/lightx2v/Wan2.1-Distill-Models/blob/main/wan2.1_t2v_14b_lightx2v_4step.safetensors
python txt2video.py 'A cat playing piano' \
    --model t2v-14B --checkpoint ./wan2.1_t2v_14b_lightx2v_4step.safetensors \
    --sampler euler --steps 4 --guidance 1.0 \
    --quantize --output out_t2v_distilled.mp4
```

I2V
```
wget https://huggingface.co/lightx2v/Wan2.1-Distill-Models/resolve/main/wan2.1_i2v_480p_lightx2v_4step.safetensors
python img2video.py 'Astronaut riding a horse' \
    --image ./inputs/astronaut-on-a-horse.png --checkpoint ./wan2.1_i2v_480p_lightx2v_4step.safetensors \
    --sampler euler --steps 4 --guidance 1.0 --shift 5.0 \
    --quantize --output out_i2v_distilled.mp4
```


video/wan2.1/wan/model.py
```diff
@@ -0,0 +1,433 @@
...
+    x = nn.silu(x)
+    x = mx.matmul(x, w2.T) + b2
+    return x
+
+
+@partial(mx.compile, shapeless=True)
+def _project_time_fn(e, w, b):
+    x = nn.silu(e)
+    x = mx.matmul(x, w.T) + b
+    return x
```
angeloskath: Almost certainly no need to compile the above. Using `nn.Linear` layers would be more understandable and the same speed. It would also use `mx.addmm` which fuses the `x @ w.T + b` into one op. For sinusoidal embedding you can use `nn.SinusoidalPositionalEncoding`. You can wrap them in lists or `nn.Sequential` like their PyTorch counterparts. It would also help with quantization as they would be automatically quantized which won't quite happen now (unless you implement a custom `to_quantized` function).

video/wan2.1/wan/model.py
```diff
@@ -0,0 +1,433 @@
...
+        self.out_dim = out_dim
+        self.num_heads = num_heads
+        self.num_layers = num_layers
+        self.head_dim = dim // num_heads
+
+        # Patch embedding: raw Conv3d weight/bias
+        self.patch_embedding_weight = mx.random.normal((dim, *patch_size, in_dim)) * (
+            1.0 / (in_dim * math.prod(patch_size)) ** 0.5
+        )
+        self.patch_embedding_bias = mx.zeros((dim,))
```
angeloskath: Same goes for here, why not `nn.Conv3d` ?

video/wan2.1/wan/model.py
```diff
@@ -0,0 +1,433 @@
...
+        for i in range(num_layers):
+            block = WanAttentionBlock(
+                dim,
+                ffn_dim,
+                num_heads,
+                cross_attn_norm,
+                eps,
+                cross_attn_type=model_type,
+            )
+            setattr(self, f"block_{i}", block)
```
angeloskath: Why not a list ? This just makes your life hard when you want to iterate over them which you always will want to.

video/wan2.1/wan/model.py
```diff
@@ -0,0 +1,433 @@
...
+            # Transpose Conv3d weights for patch_embedding
+            if (
+                "patch_embedding" in new_key
+                and "weight" in new_key
+                and len(value.shape) == 5
+            ):
+                value = mx.transpose(value, (0, 2, 3, 4, 1))
+
+            # blocks.N -> block_N
+            new_key = re.sub(r"blocks\.(\d+)\.", r"block_\1.", new_key)
```
angeloskath: Most of these are not needed when the blocks are put in a list and the ffn in a sequential (or a list) etc.

video/wan2.1/wan/model.py
```diff
@@ -0,0 +1,433 @@
...
+            new_key = re.sub(r"img_emb\.proj\.0\.(\w+)", r"img_emb_norm1.\1", new_key)
+            new_key = re.sub(r"img_emb\.proj\.1\.(\w+)", r"img_emb_linear1.\1", new_key)
+            new_key = re.sub(r"img_emb\.proj\.3\.(\w+)", r"img_emb_linear2.\1", new_key)
+            new_key = re.sub(r"img_emb\.proj\.4\.(\w+)", r"img_emb_norm2.\1", new_key)
+
+            remapped[new_key] = value
+
+        # Merge separate Q/K/V into QKV for self-attention,
+        # and K/V into KV for cross-attention
+        remapped = WanModel._merge_qkv_weights(remapped)
```
angeloskath, line 243: That is good especially for distributed inference later but generally it doesn't provide much of a speedup. The q, k, v projections will happen in parallel in the GPU anyway. 

Just to be clear, a good optimization but probably not the lowest hanging fruit.

angeloskath: The lowest hanging fruit is probably updating the `modulation` parameter in the layernorm to contain the `1 + ` so that the layernorm can just use `self.modulation + e` directly see comments in layers.py .

video/wan2.1/wan/layers.py
```diff
@@ -0,0 +1,315 @@
...
+def _modulate(x, scale, shift):
+    return x * (1 + scale) + shift
+
+
+@partial(mx.compile, shapeless=True)
+def _residual_gate(x, y, gate):
+    return x + y * gate
+
+
+_gelu = mx.compile(nn.gelu_approx)
```
angeloskath: `nn.gelu_approx` is already compiled and also shapelessly. No need for that.

video/wan2.1/wan/layers.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+@partial(mx.compile, shapeless=True)
+def _layer_norm(x, eps):
+    mean = x.mean(axis=-1, keepdims=True)
+    var = x.var(axis=-1, keepdims=True)
+    return (x - mean) / mx.sqrt(var + eps)
```
angeloskath: `mx.fast.layer_norm` can be called without scale and bias and will be quite faster than the above due to the fused mean and variance.

video/wan2.1/wan/layers.py
```diff
@@ -0,0 +1,315 @@
...
+        self.elementwise_affine = elementwise_affine
+        if elementwise_affine:
+            self.weight = mx.ones((dim,))
+            self.bias = mx.zeros((dim,))
+
+    def __call__(self, x: mx.array) -> mx.array:
+        if self.elementwise_affine:
+            return mx.fast.layer_norm(x, self.weight, self.bias, self.eps)
+        else:
+            return _layer_norm(x, self.eps)
```
angeloskath: This can be simplified as follows

```suggestion
        self.weight = mx.ones((dim,)) if elementwise_affine else None
        self.bias = mx.zeros((dim,)) if elementwise_affine else None

    def __call__(self, x: mx.array) -> mx.array:
        return mx.fast.layer_norm(x, self.weight, self.bias, self.eps)
```

or simply replaced with `nn.LayerNorm(..., affine=False)` or `nn.LayerNorm(..., affine=True)`.

video/wan2.1/wan/layers.py
```diff
@@ -0,0 +1,315 @@
...
+        num_heads: int,
+        cross_attn_norm: bool = False,
+        eps: float = 1e-6,
+        cross_attn_type: str = "t2v",
+    ):
+        super().__init__()
+        self.dim = dim
+
+        self.norm1 = WanLayerNorm(dim, eps)
+        self.norm2 = WanLayerNorm(dim, eps)
```
angeloskath: Since they have no weights, they can probably be removed.

video/wan2.1/wan/layers.py
```diff
@@ -0,0 +1,315 @@
...
+        e = self.modulation + e
+
+        # Self-attention with modulation
+        x_norm = self.norm1(x)
+        y = self.self_attn(
+            _modulate(x_norm, e[:, 1], e[:, 0]),
+            grid_sizes,
+            freqs,
+        )
+        x = _residual_gate(x, y, e[:, 2])
```
angeloskath: Assuming that sanitize adds 1 to  `self.modulation` on load at positions 1 and 4 then `e` can be used directly `self.norm1` can be removed and this will look like the following iff `e` always has a batch size of 1.

```suggestion
        y = self.self_attn(
            mx.fast.layer_norm(x, e[0, 1], e[0, 0]),
            grid_sizes,
            freqs,
        )
        x = _residual_gate(x, y, e[:, 2])
```

if `e` is batched then it's fine (but the layernorm should still use `mx.fast`).

video/wan2.1/wan/layers.py
```diff
@@ -0,0 +1,315 @@
...
+    ):
+        super().__init__()
+        self.dim = dim
+        out_features = math.prod(patch_size) * out_dim
+        self.norm = WanLayerNorm(dim, eps)
+        scale = 1.0 / dim**0.5
+        self.head_weight = mx.random.uniform(
+            low=-scale, high=scale, shape=(out_features, dim)
+        )
+        self.head_bias = mx.zeros((out_features,))
```
angeloskath: Why not a linear layer?

video/wan2.1/wan/rope.py
```diff
@@ -0,0 +1,152 @@
...
+            "sin": width_sin,
+            "dim": dim_width,
+            "full_dim": width_dim,
+        },
+        "theta": theta,
+        "head_dim": head_dim,
+    }
+
+
+@partial(mx.compile)
```
angeloskath: Very little need to compile here. I would also remove the `partial`.

video/wan2.1/wan/sampler.py
```diff
@@ -0,0 +1,433 @@
...
+from typing import List, Optional
+
+import mlx.core as mx
+
+_cpu = mx.cpu
+
+
+def _lambda64(alpha: mx.array, sigma: mx.array) -> mx.array:
+    # float64 is not supported on metal GPU, running on CPU for numerical stability
+    with mx.stream(_cpu):
```
angeloskath:
```suggestion
    with mx.stream(mx.cpu):
```

video/wan2.1/wan/sampler.py
```diff
@@ -0,0 +1,433 @@
...
+FlowUniPCMultistepScheduler for Wan2.1 denoising.
+
+UniPC multistep solver adapted for flow matching prediction.
+"""
+
+from typing import List, Optional
+
+import mlx.core as mx
+
+_cpu = mx.cpu
```
angeloskath:
```suggestion
```

video/wan2.1/wan/sampler.py
```diff
@@ -0,0 +1,433 @@
...
+        sigmas = shift * sigmas / (1 + (shift - 1) * sigmas)
+        timesteps = sigmas * self.num_train_timesteps
+
+        indices = mx.array([self.num_train_timesteps - x for x in denoising_step_list])
+        self.sigmas = sigmas[indices].astype(mx.float32)
+        self.timesteps = timesteps[indices].astype(mx.float32)
+        self.num_inference_steps = len(denoising_step_list)
+
+    def step(self, model_output, timestep, sample):
+        """Euler flow-matching update: x_new = x - sigma * f + f * sigma_next."""
```
angeloskath: This can definitely be shapelessly compiled and will have gains given that a single kernel can be produced instead of the copying around that is happening otherwise.

video/wan2.1/txt2video.py
```diff
@@ -0,0 +1,164 @@
...
+        nargs="?",
+        const=8,
+        default=0,
+        choices=[0, 4, 8],
+        metavar="{4,8}",
+        help="Quantize DiT weights (default: 8-bit when flag used without value)",
+    )
+    parser.add_argument(
+        "--n-prompt",
+        default="镜头晃动，色调艳丽，过曝，静态，细节模糊不清，字幕，风格，作品，画作，画面，静止，整体发灰，最差质量，低质量，JPEG压缩残留，丑陋的，残缺的，多余的手指，画得不好的手部，画得不好的脸部，畸形的，毁容的，形态畸形的肢体，手指融合，静止不动的画面，杂乱的背景，三条腿，背景人很多，倒着走",
```
angeloskath: I think we should have this in English at the very least .

video/wan2.1/img2video.py
```diff
@@ -0,0 +1,168 @@
...
+        nargs="?",
+        const=8,
+        default=0,
+        choices=[0, 4, 8],
+        metavar="{4,8}",
+        help="Quantize DiT weights (default: 8-bit when flag used without value)",
+    )
+    parser.add_argument(
+        "--n-prompt",
+        default="镜头晃动，色调艳丽，过曝，静态，细节模糊不清，字幕，风格，作品，画作，画面，静止，整体发灰，最差质量，低质量，JPEG压缩残留，丑陋的，残缺的，多余的手指，画得不好的手部，画得不好的脸部，畸形的，毁容的，形态畸形的肢体，手指融合，静止不动的画面，杂乱的背景，三条腿，背景人很多，倒着走",
```
angeloskath: Same as in txt2img.py .

video/wan2.1/wan/vae.py
```diff
@@ -0,0 +1,601 @@
...
+            if i != len(dim_mult) - 1:
+                mode = "upsample3d" if temporal_upsample[i] else "upsample2d"
+                stage_layers.append(("resample", Resample(out_dim, mode=mode)))
+                scale *= 2.0
+            self.upsamples.append(stage_layers)
+
+        # Register all layers via setattr
+        for stage_idx, stage in enumerate(self.upsamples):
+            for layer_idx, (layer_type, layer) in enumerate(stage):
+                setattr(self, f"upsample_s{stage_idx}_l{layer_idx}_{layer_type}", layer)
```
angeloskath: Same comment as for model.py , no need for these to be set as attributes they can be nested lists.

video/wan2.1/wan/vae.py
```diff
@@ -0,0 +1,601 @@
...
+
+            if compile and i == 1 and self._compiled_decode is None:
+                self._compiled_decode = mx.compile(self.decoder._forward_functional)
+
+            if self._compiled_decode is not None:
+                out_frame, feat_cache = self._compiled_decode(frame, feat_cache)
+            else:
+                out_frame, feat_cache = self.decoder._forward_functional(
+                    frame, feat_cache
+                )
```
angeloskath: Why not always compile?

video/wan2.1/wan/vae.py
```diff
@@ -0,0 +1,601 @@
...
+        # Pre-decoder conv
+        x = self.conv2(z)
+
+        # Frame-by-frame decode with cache
+        num_frames = x.shape[1]
+        feat_cache = [None] * 32
+        outputs = []
+
+        for i in range(num_frames):
+            frame = x[:, i : i + 1, :, :, :]
```
angeloskath, line 357: Why not batch it? For memory saving?

video/wan2.1/wan/vae_layers.py
```diff
@@ -0,0 +1,342 @@
...
+        super().__init__()
+        self.dim = dim
+        self.norm = RMSNorm(dim)
+        scale = 1.0 / dim**0.5
+        self.to_qkv_weight = mx.random.uniform(
+            low=-scale, high=scale, shape=(dim * 3, 1, 1, dim)
+        )
+        self.to_qkv_bias = mx.zeros((dim * 3,))
+        self.proj_weight = mx.zeros((dim, 1, 1, dim))
+        self.proj_bias = mx.zeros((dim,))
```
angeloskath: Same as before, should just be linear layers.

angeloskath (commented):
Great work @belkakari !

I left quite a few comments, a lot of them stylistic and some of them performance related.

One more, general comment is that the weights are in float32. That seems unnecessary and will impact performance as well as memory. It also hides possible upcasting if someone wants to have the computation happen in bf16 or another dtype which will be much faster on M5s for instance.

video/wan2.1/wan/model.py
belkakari: Got it, added modulation as you've suggested

video/wan2.1/wan/vae.py
belkakari: Yes, if the video is long enough it won't fit into the memory. We can have configurable batch size as an additional parameter with default=1, wdyt?

video/wan2.1/wan/model.py
```diff
@@ -0,0 +1,335 @@
...
+                        [
+                            mx.zeros_like(v[:, :1]),
+                            mx.ones_like(v[:, :1]),
+                            mx.zeros_like(v[:, 2:4]),
+                            mx.ones_like(v[:, :1]),
+                            mx.zeros_like(v[:, 5:]),
+                        ],
+                        axis=1,
+                    )
+                    remapped[key] = v + update
```
angeloskath: Just use broadcasting.


```suggestion
                    remapped[key] = v + mx.array([0, 1, 0, 0, 1, 0])[:, None]
```

angeloskath (approved):

PR #1442, opened 2026-08-19
Fix GCN validation and test evaluation

tandede:
## Summary

- evaluate GCN validation and test metrics with the current model parameters in evaluation mode
- restore the model's prior training state after each evaluation
- stop returning full training logits once they are no longer reused
- add a regression test for evaluation mode and state restoration

## Problem

The training step returned logits produced before `optimizer.update` and while Dropout was active. The validation path reused those stale, stochastic logits for early stopping. The test path computed a fresh `test_y_hat`, but then accidentally calculated both metrics from the old training-loop `y_hat` instead.

The new `evaluate` helper temporarily disables training behavior, performs a fresh forward pass with the current parameters, materializes the metrics, and restores the caller's original train/eval state.

## Validation

- `python gcn/test.py`: 1 test passed
- `python gcn/main.py --epochs 5 --patience 5 --hidden_dim 8 --nb_layers 1 --dropout 0.5`: completed Cora training and final evaluation
- `pre-commit run --files gcn/main.py gcn/test.py`: Black and isort passed
- `python -m compileall -q gcn`: passed

In a seeded five-step reproduction, the old pre-update training-mode output reported test loss/accuracy of `1.955827 / 0.199`, while a fresh evaluation-mode pass with the final parameters reported `1.928995 / 0.236`; the maximum logit difference was `0.410258`.

### DIFF
--- a/gcn/main.py
+++ b/gcn/main.py
@@ -26,10 +26,23 @@ def eval_fn(x, y):
     return mx.mean(mx.argmax(x, axis=1) == y)
 
 
+def evaluate(model, x, adj, y, mask):
+    training = model.training
+    model.eval()
+    try:
+        y_hat = model(x, adj)
+        loss = loss_fn(y_hat[mask], y[mask])
+        accuracy = eval_fn(y_hat[mask], y[mask])
+        mx.eval(loss, accuracy)
+    finally:
+        model.train(training)
+    return loss, accuracy
+
+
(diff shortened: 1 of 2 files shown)


tandede:
Hi maintainers, I wanted to follow up on this PR in case it was overlooked. It corrects the evaluation mode used during GCN validation and testing and is ready for review. Please let me know if any changes or further validation would be helpful. Thank you!

